PR #19, opened 2015-09-29
MNT: Sort loops in jinja-templates

### LINKED ISSUE #18: Sort everything in jinja templating so that output is deterministic
How do you all feel about sorting all of the for loops in jinja templating?  It would be nice to have these loops output in deterministic ways so that the git diffs are clearer and not filled with loops being sorted in a different order (because python 3).

ericdill:
Closes #18

### DIFF
--- a/conda_smithy/templates/run_docker_build.tmpl
+++ b/conda_smithy/templates/run_docker_build.tmpl
@@ -13,7 +13,7 @@ docker info
 config=$(cat <<CONDARC
 
 channels:
-{% for channel in channels.get('sources', []) %}
+{% for channel in channels.get('sources', []) | sort %}
  - {{ channel }}
 {% endfor %}
  - defaults # As we need conda-build
@@ -70,4 +70,3 @@ conda build --test /recipe_root || exit 1
 {% endblock %}
 
 EOF
-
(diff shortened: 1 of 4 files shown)


ocefpaf:
I just learn a new thing.  Did not now you could sort a for loop in a jinja template!

:+1:

scopatz:
Thanks @ericdill!

conda_smithy/templates/run_docker_build.tmpl
```diff
@@ -13,7 +13,7 @@ docker info
 config=$(cat <<CONDARC
 
 channels:
-{% for channel in channels.get('sources', []) %}
+{% for channel in channels.get('sources', []) | sort %}
```
jakirkham: When this was added, channel priority was not a thing. So sorting the channels likely would not have matter much. However, now that channel priority is a thing. Sorting the channels instead of respecting the user specified order is a problem. As such am proposing we drop this channel sort in PR ( https://github.com/conda-forge/conda-smithy/pull/399 ).

jakirkham: When this was added, channel priority was not a thing. So sorting the channels likely would not have mattered much. However, now that channel priority is a thing. Sorting the channels instead of respecting the user specified order is a problem. As such am proposing we drop this channel sort in PR ( https://github.com/conda-forge/conda-smithy/pull/399 ).

PR #49, opened 2016-02-02
Fix: make it possible to change the recipe dir name

JanSchulz:
the recipe-dir name was hardcoded in most functions and even if passed in as an argument not used.

This now works for a `conda smithy init recipes`


JanSchulz:
Not sure if that's intentional, but currently the recipe dir is not checked if it is absolut or not. E.g. this breaks: `conda-smithy init c:\temp\feed\recipes``

JanSchulz:
Ok, scratch this PR, I made a mistake on how this works. Sorry...

PR #145, opened 2016-04-26
Workaround Python 3.4 64-bit failure on AppVeyor

jakirkham:
Add workaround for this issue ( https://github.com/conda/conda-build/issues/895 ). Related change in staged-recipes ( https://github.com/conda-forge/staged-recipes/pull/452 ). Related issue ( https://github.com/conda-forge/staged-recipes/issues/448 ).

Applies this workaround only to Python 3.4 64-bit where the issues were seen. This way Python 2.7 64-bit on Windows still benefits from the nice fixes from @patricksnape and @msarahan.

### DIFF
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -60,6 +60,10 @@ install:
     - cmd: conda config --add channels http://conda.binstar.org/conda-forge
     - cmd: conda info
     - cmd: conda install -n root --quiet --yes conda-build anaconda-client jinja2 setuptools
+    # Workaround for Python 3.4 and x64 bug in latest conda-build.
+    # FIXME: Remove once there is a release that fixes the upstream issue
+    # ( https://github.com/conda/conda-build/issues/895 ).
+    - cmd: if "%TARGET_ARCH%" == "x64" if "%CONDA_PY%" == "34" conda install conda-build=1.20.0 --yes
 
 # Skip .NET project specific build phase.
 build: off


ocefpaf:
This implies sending maintenance PRs to update the feedstocks and then remove the workaround once conda-build is fixed.  Maybe it is worth waiting a few days for the new conda-build. (Hopefully it is an easy fix.)

jakirkham:
> This implies sending maintenance PRs to update the feedstocks...

Don't think I agree with that assessment. I think we can safely point people who run into this issue to this change (in some sort of bugfix release) so they can use it. If people are not running into this issue (likely because of low activity), they likely will never need it because the rate of change on the feedstock is lower than the rate of change on conda-build.

> Maybe it is worth waiting a few days for the new conda-build. (Hopefully it is an easy fix.)

So, I'm not loving this mentality. I'm not sure why we are being timid about keeping the feedstocks working. Past experience has shown that `conda-build` normally includes a few notable bugs each release that cause challenges. For the longest time I was pinned to `conda-build` 1.16.0 with projects I worked on because of various issues I was running into. Even as some issues were fixed, new ones cropped up making updating impossible. Either we need to start accepting that reality and allow these sorts of maintenance changes in or we need to design some new workflow that handles these sorts of changes better. Though the notion that we should leave things broken over noise just doesn't sound responsible.

msarahan:
It is regrettable that conda and conda-build have such a reputation.  We are trying to do beta releases these days.  If we can set up some kind of test queue with conda-forge or with anything else (anaconda.org build perhaps), hopefully we can increase the quality and break this pattern of brokenness.

jakirkham:
Sorry if that comes across as harsh. Unfortunately it is just how things have been.

That sounds like a fantastic idea, @msarahan. Sounds like a fair bit of work to get right though. Could we come up with some sort of plan as to how this might be implemented?

jakirkham:
One other thought that comes to mind is having packages for relevant PRs. This way we can install them easily into different feedstock's PRs and see if they behave ok. Ideally this would help with the iteration cycle for the change and allow it to be tested in some non-trivial environments.

msarahan:
It's not unnecessarily or unrealistically harsh.  These release-then-rush-to-fix patterns are very hard on us, too.

I don't have a concrete plan in mind right now, but I'll note your interest and try to come up with ideas and plans.

Your packages-per-PR idea is a good one - sort of a more manual CI system of sorts?  I want to avoid the need for humans, but we will probably never be completely unnecessary.  In parallel to this, I'd like to create known troublesome feedstocks and keep those running as automated tests on conda release candidates.

jakirkham:
Sure. The rush to release is not helping. Completely agreed. Though having pre-release candidates (as you had mentioned) may help mitigate that strain.

Just something to think about. Maybe when you have something more concrete we can include it in a meeting to get more feedback.

Yeah, I'm not saying it is the perfect solution. Though I would hope that it would be an easy thing to get started. Also, it seems like it would be immediately useful for all sides of the table.

Having some tests with troublesome feedstocks sounds great. Though I almost wonder if we couldn't have them as unit tests. Maybe it requires a bit of tweaking with the CI? Perhaps having feedstocks for `conda` and `conda-build` here would help? Not sure, but certainly some more ideas to play with.

ocefpaf:
> Though the notion that we should leave things broken over noise just doesn't sound responsible.

@jakirkham by no means I want to leave things broken. I just think that the process may not involve:

- PR to conda-smithy;
- new conda-smithy release;
- update conda-smithy feedstock;
- teach people about conda-smithy, update conda-smithy, send maintenance PR where needed;
- re-do everything once the bug is fixed in conda-build.

I think we can just add the workaround as @melund did in the [xonsh-feedstock](https://github.com/conda-forge/xonsh-feedstock/blob/master/appveyor.yml#L62) and then remove it in the future with a normal cycle of maintenance PRs.

@msarahan bare in mind that, besides continuum internal use, I don't conda-build was probably never under such a stress use before. I see this breakage as a positive thing. I strongly believe conda-build will be more robust and stable after conda-forge.

jakirkham:
Before responding to all of these points @ocefpaf, I would say that the complaint seems to be that our release process is onerous and that is why we shouldn't fix things. However, my response is we should improve our release process so that it is not onerous. Holding back bug fixes because of this still sounds irresponsible to me. Of course, feel free to correct me if I am misreading something here.

> PR to conda-smithy

So, completely disagree that we don't do PRs here to fix these things. Longer explanation later.

> new conda-smithy release

Could go either way here. Having a release makes it easier to install. Development builds are straightforward.

> update conda-smithy feedstock

Maybe this should be automated (e.g. triggered by pushing a tag here and running CI successfully) so we can have a simpler, more rapid release cycle.

> teach people about conda-smithy

This is something we should do anyways as people need to know about it. Also, having docs for it might help avoid repeated explanations.

> send maintenance PR where needed

Already agreed that this is unnecessary. Would rather people use conda-smithy to do this.

> I think we can just add the workaround as @melund did in the xonsh-feedstock and then remove it in the future with a normal cycle of maintenance PRs.

Well, I respectfully disagree. The problem with doing that is actual relevant re-rendering with conda-smithy breaks this stuff. This means maintainers have to go back and fix this manually. In turn, this slows what we have deemed important changes from actually entering the ecosystem. So, no we should include the changes here for sure. Whether we go through the full release process or not is a different question.

> I see this breakage as a positive thing.

While I think I understand what you are trying to say here. Not sure that I agree. Breakage is rarely a good thing and it did start happening before conda-forge. So, won't go so far as to say pressure from here has caused it to be more broken. We definitely should figure out how to improve the situation.

> I strongly believe conda-build will be more robust and stable after conda-forge.

However, I do agree with this. It will involve our participation for that to work though.

PR #278, opened 2016-08-29
TST: Rename temporary recipe directory

jakirkham:
We want the underscore before not after as the first argument is the [suffix]( https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp ).


jakirkham:
Seems like this should be ok to merge too. Thoughts @pelson?

PR #290, opened 2016-09-01
add instructions on how to specify target channels to upload

shadowwalkersb:
I thought it would be quite useful to have these instructions. I had to figure them out by looking at the code and doing trial--and-error. Let me know if I need to update anything if you are interested in the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,11 +50,11 @@ instead of an organization by replacing the organization argument with
 This requires tokens for the CI services. You can give the name of a user instead
 of organization with `--user github_user_name`.
     - Optionally, you can choose a channel to upload to in `conda-forge.yml`.
-        ```
-        channels:
-         targets:
-           - [target_channel, target_label]
-        ```
+    ```
+      channels:
+       targets:
+         - [target_channel, target_label]
+    ```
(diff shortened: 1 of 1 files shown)


jakirkham:
I like the addition, but I don't like the location. Maybe we need it as a separate number or in a different/new section (feedstock management?) or similar.

shadowwalkersb:
I don't have much to offer regarding the location.

pelson:
I kind of think it fits OK with the line above (i.e. `--user github_user_name`). Not that I think either is the content I would want to see a new conda-forge feedstock maintainer having to go through - this is entirely for the non conda-forge usecase IMO.

@jakirkham - you have merge rights, and I'm 👍 . Please feel free to iterate/merge as you see fit.

shadowwalkersb:
Here is how it looks.

screen shot 2016-09-03 at 11 33 34 am

jakirkham:
Yep that looks good. Thanks @shadowwalkersb.

This might have introduced conflicts with PR ( https://github.com/conda-forge/conda-smithy/pull/286 ), but I guess there is one more thing we want to do there anyways.

Issue #289, opened 2016-09-01
Build failure on master


jakirkham:
We are seeing a build failure on `master`. Here is a snippet from one [build](https://travis-ci.org/conda-forge/conda-smithy/jobs/156757548) below.

```
$ conda install --file requirements.txt nose mock python=${PYTHON} ${CONDA_PKGS} --yes --quiet -c conda-forge
Fetching package metadata .........
Solving package specifications: ..........
Package plan for installation in environment /home/travis/conda:
The following packages will be downloaded:
    package                    |            build
    ---------------------------|-----------------
    ncurses-5.9                |                8         954 KB  conda-forge
    patchelf-0.9               |                1         349 KB  conda-forge
    python-3.5.2               |                2        31.4 MB  conda-forge
    certifi-2016.8.2           |           py35_0         219 KB  conda-forge
    filelock-2.0.6             |           py35_0           8 KB  conda-forge
    markupsafe-0.23            |           py35_0          33 KB  conda-forge
    pycrypto-2.6.1             |           py35_0         449 KB  conda-forge
    pygithub-1.27.1            |           py35_0         2.0 MB  conda-forge
    pytz-2016.6.1              |           py35_0         162 KB  conda-forge
    requests-2.11.0            |           py35_0         847 KB  conda-forge
    six-1.10.0                 |           py35_0          18 KB  conda-forge
    smmap-0.9.0                |           py35_0          30 KB  conda-forge
    conda-4.1.11               |           py35_1         390 KB  conda-forge
    gitdb-0.6.4                |           py35_0         113 KB  conda-forge
    python-dateutil-2.5.3      |           py35_0         237 KB  conda-forge
    setuptools-26.1.0          |           py35_0         353 KB  conda-forge
    clyent-1.2.1               |           py35_0          14 KB  conda-forge
    gitpython-2.0.8            |           py35_0         654 KB  conda-forge
    jinja2-2.8                 |           py35_1         278 KB  conda-forge
    nose-1.3.7                 |           py35_2         389 KB  conda-forge
    ruamel.yaml-0.11.11        |           py35_1         563 KB  conda-forge
    anaconda-client-1.5.1      |           py35_0         112 KB  conda-forge
    conda-build-2.0.0          |           py35_0         260 KB
    pbr-1.10.0                 |           py35_0          99 KB  conda-forge
    mock-2.0.0                 |           py35_0         101 KB  conda-forge
[...]

PR #294, opened 2016-09-06
Pin `setuptools` as CI failure workaround

jakirkham:
Workaround for issue ( https://github.com/conda-forge/conda-smithy/issues/289 ).

This temporarily pins `setuptools` to a known working version. The underlying problem is going to take a bit to get sorted out and resolved. In the meantime, this should at least ensure CI is working here again until it gets fixed.

cc @msarahan


jakirkham:
Going to toss this one in, @pelson, so that `master` and PRs stop failing. Feel free to let me know if there should be something more and we can do a follow-on.

PR #300, opened 2016-09-13
feedstock upload script supports conda build 2.0.0

jjhelmus:


jjhelmus:
Change to upload_or_check_non_existence.py script is the same as the change made in conda-forge/conda-forge-build-setup-feedstock#34

jakirkham:
No complaints from me.

That being said, @pelson and I discussed switching to calling the script from `conda-forge-build-setup`. It will take a bit of tweaking on both ends to get working, but that seems like the best path forward as that script is going to require more work soon.

In any event, would be happy to merge this even if we do follow it up with removal.

jjhelmus:
Happy to help move either solution forward.  Our python 3.5 package currently has an issue (conda-forge/python-feedstock#70) which needs conda build 2.0.1+ to fix.

jakirkham:
Alright, I don't see any harm in merging this so have done so. This way we have this change if we need to cut a release that works with `conda-build` version `2`.

That being said, we should start moving on getting this content out of the feedstock and using `conda-forge-build-setup`. We need to do that as this script has some other issues that need to be fixed. If you would be open to starting a PR on that @jjhelmus, I think we should be able to get it merged in quick order. FWIW the changes should be pretty trivial.

jakirkham:
Please see PR ( https://github.com/conda-forge/conda-smithy/pull/301 ) for moving to `conda-forge-build-setup`. Seem to be having some struggles with Windows.

jakirkham:
Should add there is still a bit of an issue using the included script. Please see the closed PR ( https://github.com/conda-forge/conda-smithy/pull/273 ). That got included in `conda-forge-build-setup`, but not here. We would need it included here if we really want to keep using the included script.

PR #302, opened 2016-09-13
Correct arg used to list feedstocks

dkillick:
Listing available feedstocks on a given GitHub organisation was not working because the specified argument did not exist. I've corrected the specified argument so that it gets the requested organization from the incoming `args`. I've also updated the help string to make reference to the fact that you can specifiy a GitHub organization using the `--organization` flag.


jakirkham:
Nice fix. Not entirely sure how this worked before. 😄 Tested it locally and found it worked great.

PR #326, opened 2016-10-06
fix whitespace README template

shadowwalkersb:


jakirkham:
Have you tried re-rendering with this? It may snatch up too much whitespace. Normally I try to pick one side for the `-` and then adjust other things to match.

conda_smithy/templates/README.md.tmpl
```diff
@@ -34,7 +34,7 @@ It is possible to list all of the versions of `{{ package.name() }}` available o
 conda search {{ package.name() }} --channel {{ channel_name }}
 ```
 
-{% if channel_name == 'conda-forge' %}
+{%- if channel_name == 'conda-forge' -%}
```
jakirkham: So this isn't quite the right way to do this. It is eliminating all the whitespace between this and the next title. Would pick one side to `-` and then move this line up or down in the document accordingly to avoid further whitespace errors.

jakirkham: So this isn't quite the right way to do this. It is eliminating all the whitespace between this and the next title. Would pick one side to `-` and then move this line up or down in the document accordingly to avoid further whitespace errors.

jakirkham:
Thanks @shadowwalkersb .

PR #331, opened 2016-10-17
Repo initialization adjustments

jakirkham:
Follow up on PR ( https://github.com/conda-forge/conda-smithy/pull/309 ). Makes some adjustments to repo initialization. In particular, makes sure that `ci_support/checkout_merge_commit.sh` is added to `git`'s index with executable permissions. Also uses `git add -A` to ensure that all files are added to index regardless of our location. Lastly went ahead and dropped the OS check as there are all sorts of weird situations that could occur were the permissions are not correctly added. So it just makes sense to add them always. Checked on my Mac to verify it still worked fine.


jakirkham:
Replaced with PR ( https://github.com/conda-forge/conda-smithy/pull/335 ).

Issue #330, opened 2016-10-17
Use bash calls explicitly instead of relying on executable permissions


nicoddemus:
As discussed in conda-forge/pytest-cov-feedstock#6, `conda smithy` should use explicit bash calls instead of relying on executable permissions because those can get lost on Windows.

In that specific case, I was running a vagrant box on Windows and cloned pytest-cov-feedstock in a shared folder in order to re-render it.

jakirkham:
If you have some time, I'd be very curious to know what happens when you do the following.

1. Clone the [`libgfortran` feedstock]( https://github.com/conda-forge/libgfortran-feedstock ).
2. Change the [`skip`]( https://github.com/conda-forge/libgfortran-feedstock/blob/21abaaa1d8de17357c954a6d6ed761a99dc10513/recipe/meta.yaml#L12 ) to be Windows only.
3. [Re-render]( https://github.com/conda-forge/conda-smithy#re-rendering-an-existing-feedstock ) with `conda-smithy`
4. Inspect the permissions of `ci_support/run_docker_build.sh` and `ci_support/checkout_merge_commit.sh`.

If they differ, this would be very interesting to know. If they are the same, this would also be good to know (though less surprising).

nicoddemus:
Done. I cloned the `libgfortran` on my shared folder (`/vagrant`).

```diff
diff --git a/recipe/meta.yaml b/recipe/meta.yaml
index b2f928d..8e43c49 100644
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -9,7 +9,7 @@ package:

 build:
   number: 0
-  skip: true                                                  # [not osx]
+  skip: true                                                  # [win]
   always_include_files:
     - lib/libgfortran.dylib                                   # [osx]
     - lib/libgfortran.{{ libgfortran_version[0] }}.dylib      # [osx]
```

```
vagrant@thdengops-ubuntu-14:/vagrant/libgfortran-feedstock$ ls -la ci_support/
total 7
drwxrwxrwx 1 vagrant vagrant    0 Oct 17 21:57 .
drwxrwxrwx 1 vagrant vagrant 4096 Oct 17 21:57 ..
-rwxrwxrwx 1 vagrant vagrant  649 Oct 15 19:10 checkout_merge_commit.sh
-rwxrwxrwx 1 vagrant vagrant 1366 Oct 17 21:57 run_docker_build.sh
```

Please let me know if you would like me to do some other tests.

[...]

PR #332, opened 2016-10-17
WIP: Run with `bash` explicitly

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/330

Sometimes weird things happen and these scripts don't have executable permissions. This could be due to subtleties when working with VMs and/or Windows. To avoid everything going pear-shaped, don't assume these scripts have the right permissions and just run them through `bash`.

### DIFF
--- a/conda_smithy/templates/circle.yml.tmpl
+++ b/conda_smithy/templates/circle.yml.tmpl
@@ -11,7 +11,7 @@ general:
 {%- if matrix -%}
 checkout:
   post:
-    - ./ci_support/checkout_merge_commit.sh
+    - bash ci_support/checkout_merge_commit.sh
 
 machine:
   services:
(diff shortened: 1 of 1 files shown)


jakirkham:
I don't think this is necessary in light of PR ( https://github.com/conda-forge/conda-smithy/pull/335 ) and PR ( https://github.com/conda-forge/conda-smithy/pull/336 ) in particular. Though I need to get feedback from more users before I can say. In any event, let's hold off on this for the time being.

jakirkham:
It looks like PR ( https://github.com/conda-forge/conda-smithy/pull/336 ) has [fixed]( https://github.com/conda-forge/conda-smithy/issues/330#issuecomment-257111570 ) the underlying issue. Going to close this out for now. If we determine it is still needed, we can always reopen.

PR #335, opened 2016-10-18
Fix permissions

jakirkham:
Adds executable permissions to `ci_support/checkout_merge_commit.sh` in a few places. It should have them anyways because the file already has executable permissions. However, there are some oddities on Windows that need to be addressed.

Also makes sure executable permissions are added for everyone, group, and user on both `ci_support/checkout_merge_commit.sh` and `ci_support/run_docker_build.sh`. We were just adding executable permissions for the user before however it seems `os.chmod` conferred them to group and everyone anyways. This appears to be inadvertent AFAICT. So we now do explicitly in this PR.

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -41,10 +41,24 @@ def init_git_repo(target):
 def create_git_repo(target, msg):
     init_git_repo(target)
     subprocess.check_call(['git', 'add', '-A'], cwd=target)
-    if sys.platform == "win32":
-        # Ensure shell scripts have executable permissions.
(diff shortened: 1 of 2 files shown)


jakirkham:
Pulled in changes from PR ( https://github.com/conda-forge/conda-smithy/pull/331 ) into this one and closed that PR out.

jakirkham:
Putting this in to handle some funky permissions issues we have run into. This primarily useful for local work. Mainly on Windows or when interacting with a Windows backed file system (e.g. VMs with shared directories). Shouldn't have any effect on our automation bots that are already on Linux and thus track permissions the same way git does anyways.

PR #336, opened 2016-10-18
Stage re-rendering changes

jakirkham:
This automatically stages changes from a re-rendering run in the index if the feedstock is a git repo. If it is not a git repo, it proceeds exactly the same as it would otherwise. By putting things directly into the index, we should be able to avoid file permission getting lost like in issue ( https://github.com/conda-forge/conda-smithy/issues/330 ).

Also provides the option to commit either with the editor popped up or automatically with a default message. Note that committing is not default behavior unlike staging.

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -198,7 +198,6 @@ class Regenerate(Subcommand):
             configure_feedstock.main(args.feedstock_directory)
             print("\nCI support files regenerated. These need to be pushed to github!")
             print("\nYou can commit the changes with:\n\n"
-                  "    git add -A\n"
                   "    git commit -m 'MNT: Re-rendered with conda-smithy %s'" % __version__)
         except RuntimeError as e:
             print(e)
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -192,14 +192,51 @@ class Regenerate(Subcommand):
         scp = self.subcommand_parser
         scp.add_argument("--feedstock_directory", default=os.getcwd(),
                          help="The directory of the feedstock git repository.")
+        scp.add_argument("-c", "--commit", nargs='?', choices=["edit", "auto"], const="edit",
+                         help="Whether to setup a commit or not.")
 
     def __call__(self, args):
         try:
             configure_feedstock.main(args.feedstock_directory)
-            print("\nCI support files regenerated. These need to be pushed to github!")
-            print("\nYou can commit the changes with:\n\n"
-                  "    git commit -m 'MNT: Re-rendered with conda-smithy %s'" % __version__)
+            print("\nRe-rendered with conda-smithy %s'." % __version__)
+
+            is_git_repo = os.path.exists(os.path.join(args.feedstock_directory, ".git"))
+            if is_git_repo:
+                has_staged_changes = subprocess.call(
+                    [
+                        "git", "diff", "--cached", "--quiet", "--exit-code"
+                    ],
+                    cwd=args.feedstock_directory
+                )
+                if has_staged_changes:
+                    if args.commit:
+                        git_args = [
+                            'git',
+                            'commit',
+                            '-m',
+                            'MNT: Re-rendered with conda-smithy %s' % __version__
+                        ]
+                        if args.commit == "edit":
+                            git_args += [
+                                '--edit',
+                                '--status',
+                                '--verbose'
+                            ]
+                        subprocess.check_call(
+                            git_args,
+                            cwd=args.feedstock_directory
+                        )
+                    else:
+                        print(
+                            "\nYou can commit the changes with:\n\n"
+                            "    git commit -m 'MNT: Re-rendered with conda-smithy %s'\n" % __version__
+                        )
+                    print("\nThese changes need to be pushed to github!\n")
+                else:
+                    print("\nNo changes made. This feedstock is up-to-date.\n")
         except RuntimeError as e:
+            print(e)
+        except subprocess.CalledProcessError as e:
             print(e)
 
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -32,6 +32,77 @@ def meta_config(meta):
     return config
 
 
+def get_repo(path, search_parent_directories=True):
+    repo = None
+    try:
+        import git
+        repo = git.Repo(
+            path,
+            search_parent_directories=search_parent_directories
+        )
+    except ImportError:
+        pass
+    except git.InvalidGitRepositoryError:
+        pass
+
+    return repo
+
+
+def get_file_blob(repo, filename):
+    idx = repo.index
+    rel_filepath = os.path.relpath(filename, repo.working_dir)
+    blob = idx.iter_blobs(lambda _: _[1].path == rel_filepath).next()[1]
+    return blob
+
+
+def get_mode_file(filename):
+    repo = get_repo(filename)
+    if repo:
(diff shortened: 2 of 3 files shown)


jakirkham:
I think you may find this of interest, @minrk .

jakirkham:
Also raising to your attention @pelson . Should add this allows one the mind–numbing option to do `conda smithy rerender --commit=auto` and just push. 😄

conda_smithy/cli.py
```diff
@@ -179,14 +179,34 @@ def __init__(self, parser):
         scp = self.subcommand_parser
         scp.add_argument("--feedstock_directory", default=os.getcwd(),
                          help="The directory of the feedstock git repository.")
+        scp.add_argument("-c", "--commit", nargs='?', choices=["edit", "auto"], const="edit",
```
jakirkham: Thanks. I'm pretty happy with it myself. 😄 This was inspired by my experiences contributing to Homebrew. 

Yeah, was thinking the same things last night when I wrote it. My main considerations for these choices are as follows.

First, not all re-renderings are simple feedstock updates. So it is nice to be able to explain what the re-rendering was intended to accomplish. While one can always amend after, IMHO it is nice to have one command to do the whole thing.

Second, the default edit mode feels a little nicer from the novice `git`/`conda-smithy` user perspective. Mainly because it is obvious a commit is happening. Also it lists the files changed and shows the diff of what is being committed in the editor. This can be great for catching something that shouldn't have been committed ( handy even for an experienced `git` user 😉 ). Though not being a novice with either, I may be a bad judge of what helps.

That all being said, I'm not terribly attached to which one is default. So if `auto` is a better fit, am happy to change it.

conda_smithy/cli.py
```diff
@@ -179,16 +179,50 @@ def __init__(self, parser):
...
+                         help="Whether to setup a commit or not.")
 
     def __call__(self, args):
         try:
             configure_feedstock.main(args.feedstock_directory)
             print("\nCI support files regenerated. These need to be pushed to github!")
```
jakirkham: We should also move this into the `if has_staged_changes` branch below. No point in telling someone to push unless changes were committed.

conda_smithy/cli.py
nicoddemus: (Sorry, I thought I had replied to this earlier)

I say let's go with your gut feeling then, we can change that behavior later if need to. 👍

conda_smithy/cli.py
jakirkham: Fixed below. Now only states to push if there were changes. Also if there were no changes, notes that everything is up-to-date.

conda_smithy/cli.py
jakirkham: No worries. It was a bit long winded. 😜

SGTM. Am happy to iterate based on user experience. 😄

conda_smithy/configure_feedstock.py
```diff
@@ -32,6 +32,77 @@ def meta_config(meta):
...
+
+        os.remove(filename)
+
+
+def copy_file(src, dst):
+    shutil.copy2(src, dst)
+
+    repo = get_repo(dst)
+    if repo:
+        repo.index.add([dst])
```
jakirkham: TBH this is a lot of utility functions that are all similarly themed. Have been thinking awhile that they should move into a different module. Maybe an `io` module? Thoughts?

jakirkham:
Also have been using this myself for the better part of a week and it has been a good experience thus far. Would be curious if anyone has tried using this in development and has any suggestions.

conda_smithy/configure_feedstock.py
jakirkham: Moved all of these utility functions to a module called `io` and imported them into `configure_feedstock`. Makes things a little cleaner and easier to follow.

jakirkham:
Alright, I'm going to go ahead and get this out there. Would certainly welcome suggestions and feedback. Though I feel the easiest way to get that is put this in people's hands. This will be tagged as `1.4.0`. That being said, it should not have any fundamentally different results from `1.3.3` (unless this fixes a subtle permission issue). Certainly the bot should have no different results.

conda_smithy/feedstock_io.py
```diff
@@ -0,0 +1,74 @@
...
+    except git.InvalidGitRepositoryError:
+        pass
+
+    return repo
+
+
+def get_file_blob(repo, filename):
+    idx = repo.index
+    rel_filepath = os.path.relpath(filename, repo.working_dir)
+    blob = idx.iter_blobs(lambda _: _[1].path == rel_filepath).next()[1]
```
jakirkham: Turns out the `next` method here is trouble. Particularly on Python 3. Need to use the `next` function instead.

jakirkham: Appears to be fixed by PR ( https://github.com/conda-forge/conda-smithy/pull/340 ).

jakirkham:
Adding some docs for this feature in PR ( https://github.com/conda-forge/conda-smithy/pull/341 ).

conda_smithy/cli.py
```diff
@@ -192,16 +192,52 @@ def __init__(self, parser):
...
+                         help="Whether to setup a commit or not.")
 
     def __call__(self, args):
         try:
             configure_feedstock.main(args.feedstock_directory)
-            print("\nCI support files regenerated. These need to be pushed to github!")
-            print("\nYou can commit the changes with:\n\n"
-                  "    git add -A\n"
-                  "    git commit -m 'MNT: Re-rendered with conda-smithy %s'" % __version__)
+            print("\nRe-rendered with conda-smithy %s'." % __version__)
```
jakirkham: Missed that there was still a `'` in this string. Removing it in PR ( https://github.com/conda-forge/conda-smithy/pull/342 ).

conda_smithy/cli.py
```diff
@@ -192,16 +192,52 @@ def __init__(self, parser):
...
+                            cwd=args.feedstock_directory
+                        )
+                    else:
+                        print(
+                            "\nYou can commit the changes with:\n\n"
+                            "    git commit -m 'MNT: Re-rendered with conda-smithy %s'\n" % __version__
+                        )
+                    print("\nThese changes need to be pushed to github!\n")
+                else:
+                    print("\nNo changes made. This feedstock is up-to-date.\n")
```
jakirkham: Made some spacing adjustments with these `print` statements. Seems there were a couple cases that got a bit clumped together. PR ( https://github.com/conda-forge/conda-smithy/pull/343 ) seems to fix this issue.

jakirkham:
Adding some docs for this feature in PR ( https://github.com/conda-forge/conda-smithy/pull/341 ).

conda_smithy/feedstock_io.py
```diff
@@ -0,0 +1,74 @@
...
+    repo = get_repo(filename)
+    if repo:
+        repo.index.add([filename])
+
+
+def remove_file(filename):
+    if os.path.exists(filename):
+        repo = get_repo(filename)
+        if repo:
+            repo.index.remove([filename])
```
jakirkham: In some cases, the file may already be removed from the file system, but not the index. Imagine a re-rendering that goes bad and is partially backed out with `git reset`. Alternatively the file may never have existed in the index even though we are trying to remove it. To fix the former case and handle the latter one, we can create the file beforehand and add it to the index before trying to remove it. As a result, we can simply go about removal before checking for the file as we are always guaranteed to have it. This exactly what we propose in PR ( https://github.com/conda-forge/conda-smithy/pull/345 ).

PR #334, opened 2016-10-18
Lint Redundant License

jakirkham:
Have the linter raise if the `license` includes the word "license" in it.

cc @goanpeca

PR #339, opened 2016-10-28
Linter: check the license_family is valid

kynan:
conda-build actually checks for that. Relates to #88.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -6,7 +6,7 @@ import re
 import jinja2
 import ruamel.yaml
 
-from conda_build.metadata import allowed_license_families
+from conda_build.metadata import ensure_valid_license_family
 
 # patch over differences between PY2 and PY3
 try:
@@ -137,10 +137,10 @@ def lintify(meta, recipe_dir=None):
                          'file.')
 
     # 12: License family must be valid (conda-build checks for that)
-    license_family = about_section.get('license_family', '')
-    if license_family and not license_family in allowed_license_families:
-        lints.append('The recipe license_family `{}` is invalid: must be one of {}.'
-                     ''.format(license_family, ', '.join(allowed_license_families)))
+    try:
+        ensure_valid_license_family(meta)
+    except RuntimeError as e:
+        lints.append(str(e))
 
     return lints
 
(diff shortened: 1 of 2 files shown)


conda_smithy/lint_recipe.py
```diff
@@ -86,7 +88,13 @@ def lintify(meta, recipe_dir=None):
     if 'unknown' == license.strip():
         lints.append('The recipe license cannot be unknown.')
 
-    # 6: Selectors should be in a tidy form.
+    # 6: License family must be valid (conda-build checks for that)
```
jakirkham: I don't know that we want to be messing with the number. Maybe @pelson can comment.

jakirkham:
It's a bit redundant as `conda-build` does check it. That being said, am in favor of not having to hunt through logs for information like this. So having the linter say it is an improvement.

kynan:
I agree it's a redundant check, but it will be detected at lint rather than build time.

conda_smithy/lint_recipe.py
jakirkham: Let's go ahead and put it at the end. May need to rebase though as I just merged PR ( https://github.com/conda-forge/conda-smithy/pull/334 ).

kynan:
@jakirkham done!

conda_smithy/lint_recipe.py
```diff
@@ -122,6 +124,12 @@ def lintify(meta, recipe_dir=None):
         lints.append('The recipe `license` should not include the word '
                      '"License".')
 
+    # 11: License family must be valid (conda-build checks for that)
+    license_family = about_section.get('license_family', '')
+    if license_family and not license_family in allowed_license_families:
+        lints.append('The recipe license_family `{}` is invalid: must be one of {}.'
+                     ''.format(license_family, ', '.join(allowed_license_families)))
```
jakirkham: Actually why don't we just use `ensure_valid_license_family` to verify this?

jakirkham: That's certainly an option and should be easy to do. 👍

kynan:
@jakirkham now using `ensure_valid_license_family` directly. The build was cancelled, can't tell why.

jakirkham:
> The build was cancelled, can't tell why.

Weird. 😕 Restarted and seems to have worked.

However, there appears to be a merge conflict. Likely due to another linter change.

kynan:
@jakirkham rebased on upstream master!

PR #341, opened 2016-10-29
Note re-render commit option

jakirkham:
Add documentation on re-renders new commit option to the README.


jakirkham:
@conda-forge/core , have recently added a feature in PR ( https://github.com/conda-forge/conda-smithy/pull/336 ) to commit the re-rendering changes with a default message as explained here. This should allow one to just do `conda smithy rerender --commit` (or `-c` for lazy people like myself 😉) as explained here. Try version `1.4.2` if you would like to take it for a spin. It would be good to get some feedback on this change so we can iron out any kinks before we document it.

PR #349, opened 2016-10-30
Set default Python to 3.5 in appveyor.yml template

hadim:
Why not 3.5 by default ?

@ocefpaf

### DIFF
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -13,7 +13,7 @@ environment:
 
   # We set a default Python version for the miniconda that is to be installed. This can be
   # overridden in the matrix definition where appropriate.
-  CONDA_PY: "27"
+  CONDA_PY: "35"
 
   {%- for name, hashed_secure in (appveyor.secure or {}) | dictsort %}
   {{ name }}:


ocefpaf:
Don't really have an opinion here. I believe the stack of tools we use are ready for `PY35` but I will leave this one for @pelson.

jakirkham:
This should be safe to merge. Also we don't currently use `conda-build-all` or anything in the feedstocks (though there have been discussions at times).

That being said, I don't think this variable setting really matters. I believe it is just overridden by the matrix. So the only times it should matters is when we don't have Python as a build time dependency.

jakirkham:
As reference, @pelson planned to [drop this line]( https://github.com/conda-forge/conda-smithy/pull/193/files#diff-a21d424d600d24e538028a242ff2da78L16 ) entirely as it doesn't really serve a purpose.

jakirkham:
So there is a default still on these [lines]( https://github.com/conda-forge/conda-smithy/blob/ecfb5a71fd07214048b522e0840ebc3cb33e697f/conda_smithy/templates/appveyor.yml.tmpl#L14-L15 ). However, they will almost certainly be overridden. In fact, those lines probably should be removed as they should never be used.

To really get the intended effect, one should change this [line]( https://github.com/conda-forge/conda-smithy/blob/ecfb5a71fd07214048b522e0840ebc3cb33e697f/conda_smithy/configure_feedstock.py#L309 ) to `else`.

jakirkham:
> To really get the intended effect, one should change this [line]( https://github.com/conda-forge/conda-smithy/blob/ecfb5a71fd07214048b522e0840ebc3cb33e697f/conda_smithy/configure_feedstock.py#L309 ) to `else`.

If you still are interested in this change @hadim , please submit a new PR with the change described.

Issue #314, opened 2016-09-29
Pruning empty directories


jakirkham:
In the upcoming release of `conda-smithy`, we are going to have an empty directory. IIUC this is the first time this has occurred. In particular, the `ci_support` will be empty when re-rendering feedstocks that don't build on Linux.

This has no effect on `git` as it has no way to track empty directories. So as far as re-rendering is concerned, the directory will be removed, but locally it will remain.

Now one can do `git clean -fd` to remove these locally. However, it would be nice if we handled this for them.

jakirkham:
Fixing in PR ( https://github.com/conda-forge/conda-smithy/pull/353 ).

PR #353, opened 2016-11-01
Handle empty directories

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/314
Improves on https://github.com/conda-forge/conda-smithy/pull/345

This handles two cases. First it handles a case where one tries to write to a file when its parent directory does not exist. Second it handles the case where the last file in a directory is removed leaving empty directory/directories.

The first case is handled by making all directories up to the parent directory of the file recursively until they all exist. This is of particular importance for `remove_file`, which runs `touch_file` before removal to ensure the file exists. More details about why that is done can be found in PR ( https://github.com/conda-forge/conda-smithy/pull/345 ). As a result, we can be sure the file can be written to whether or not those directories existed before.

The second case is handled during `remove_file` by checking to see if the directory is now empty. If it is, then run `os.removedirs`, which will remove empty directories recursively until there are no more directories or a non-empty directory is encountered. As the feedstock directory should always be non-empty (at least containing a recipe), this should never go out past the feedstock directory.

### DIFF
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -49,6 +49,10 @@ def set_mode_file(filename, mode):
 
 @contextmanager
 def write_file(filename):
+    dirname = os.path.dirname(filename)
+    if not os.path.exists(dirname):
+        os.makedirs(dirname)
+
     with open(filename, "w") as fh:
         yield fh
 
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -75,6 +75,10 @@ def remove_file(filename):
 
     os.remove(filename)
 
+    dirname = os.path.dirname(filename)
+    if not os.listdir(dirname):
+        os.removedirs(dirname)
+
 
 def copy_file(src, dst):
     shutil.copy2(src, dst)


jakirkham:
Would love to have feedback from all of you on this one, @shadowwalkersb @minrk @nicoddemus .

nicoddemus:
The patch seems straightforward, good work! 👍 

How about adding some tests? It seems it should simple to add some tests for those functions. The test for the first change could be simply (using pytest):

```python
def test_auto_create_dirs(tmpdir):
   fn = tmpdir / 'foo' / 'bar.txt'
    with write_file(str(fn)) as f:
        f.write('contents')
    assert fn.read() == 'contents'
```

jakirkham:
So staged-recipes appears to be [failing]( https://travis-ci.org/conda-forge/staged-recipes#L256 ) without this change FYI.

cc @synapticarbors

jakirkham:
Yep, adding tests seems totally reasonable and desirable. TBH `conda-smithy` is pretty weak in the testing department. So having these sorts of simple utility functions should help make it easier to test as you have demonstrated.

That being said, it will probably have to be later today or in the week depending on what time I have.

shadowwalkersb:
The PR looks reasonable to me. 👍

jakirkham:
Adding a bunch of tests in PR ( https://github.com/conda-forge/conda-smithy/pull/357 ).

jakirkham:
So am definitely in favor of adding a test for this and will do so momentarily in the aforementioned testing PR. However, this is also fixing a bug that is blocking staged-recipes from converting anything to feedstocks. Therefore, I'm going to merge this into a patch release so that we can fix staged-recipes. Then we can return to the testing discussion.

Issue #350, opened 2016-10-30
Issues on windows (conda smithy regenerate)


janschulz:
I tried to fix the problems in https://travis-ci.org/conda-forge/pypandoc-feedstock/jobs/171846025 by rerendering the [pypandoc-feedstock](https://github.com/conda-forge/pypandoc-feedstock), but I ran into two issues (I'm on windows 7):

```
λ conda smithy regenerate
Traceback (most recent call last):
  File "C:\portabel\miniconda\Scripts\conda-smithy-script.py", line 5, in <module>
    sys.exit(conda_smithy.cli.main())
  File "C:\portabel\miniconda\lib\site-packages\conda_smithy\cli.py", line 289, in main
    args.subcommand_func(args)
  File "C:\portabel\miniconda\lib\site-packages\conda_smithy\cli.py", line 200, in __call__
    configure_feedstock.main(args.feedstock_directory)
  File "C:\portabel\miniconda\lib\site-packages\conda_smithy\configure_feedstock.py", line 455, in main
    copy_feedstock_content(forge_dir)
  File "C:\portabel\miniconda\lib\site-packages\conda_smithy\configure_feedstock.py", line 361, in copy_feedstock_content
    'README'
  File "C:\portabel\miniconda\lib\site-packages\conda_smithy\configure_feedstock.py", line 352, in copytree
    copy_file(s, d)
  File "C:\portabel\miniconda\lib\site-packages\conda_smithy\feedstock_io.py", line 74, in copy_file
    repo.index.add([dst])
  File "C:\portabel\miniconda\lib\site-packages\git\index\base.py", line 745, in add
    entries_added.extend(self._entries_for_paths(paths, path_rewriter, fprogress, entries))
  File "C:\portabel\miniconda\lib\site-packages\git\util.py", line 55, in wrapper
    return func(self, *args, **kwargs)
  File "C:\portabel\miniconda\lib\site-packages\git\index\util.py", line 84, in set_git_working_dir
    return func(self, *args, **kwargs)
  File "C:\portabel\miniconda\lib\site-packages\git\index\base.py", line 631, in _entries_for_paths
    entries_added.append(self._store_path(filepath, fprogress))
  File "C:\portabel\miniconda\lib\site-packages\git\index\base.py", line 599, in _store_path
    istream = self.repo.odb.store(IStream(Blob.type, st.st_size, stream))
  File "C:\portabel\miniconda\lib\site-packages\gitdb\db\loose.py", line 230, in store
[...]

PR #358, opened 2016-11-02
NO MRG: Demonstrate get_file_blob bug

jakirkham:
Please do **not** merge this PR.

Discovered a bug in `get_file_blob` when writing a unit test for it. This PR exists solely to demonstrate that bug. The fix will be shown in a subsequent PR.

I think this may be related to second problem described by the OP in issue ( https://github.com/conda-forge/conda-smithy/issues/350 ). However, I don't have enough details to confirm whether this is the exact same case or just another case that throws the same exception. In any event, fixing this problem is helpful in either event.


jakirkham:
Weird. While this definitely reproduces the problem for me locally on my Mac, it does not seem to on Travis CI. In any event, going to close this out as it was for testing anyways.

PR #359, opened 2016-11-02
Fix get_file_blob bug

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/350 (?)
Closes https://github.com/conda-forge/conda-smithy/pull/358

Came across the second problem described by the OP in issue ( https://github.com/conda-forge/conda-smithy/issues/350 ) when writing some unit tests. Demonstration of its recreation can be seen in PR ( https://github.com/conda-forge/conda-smithy/pull/358 ).

Cannot confirm that this fixes that issue as I don't have enough details to determine whether they are the same problem or two different problems with the same error. In any event, fixing this problem is helpful in either event.

cc @janschulz

### DIFF
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -21,8 +21,11 @@ def get_repo(path, search_parent_directories=True):
 
 def get_file_blob(repo, filename):
     idx = repo.index
-    rel_filepath = os.path.relpath(filename, repo.working_dir)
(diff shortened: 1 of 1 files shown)


jakirkham:
In the end, we rewrote this with `BlobFilter`, which appears to be much more robust and much simpler in the end.

xref: https://github.com/conda-forge/conda-smithy/pull/363

Issue #364, opened 2016-11-03
UnicodeEncodeError during rerender if meta.yaml contains unicode chars [Python2]


marscher:
when calling rerender in a feedstock, I get the following error, when my root env is running Python 2.7. Note that I do not have created CI tokens yet, but since the re-render step should not rely on the tokens and is already writing/updating the readme, I'd guess it is a bug.
```
marscher@jigsaw:~/sources/conda-forge/pyemma-feedstock$ conda smithy rerender
No circle token.  Create a token at https://circleci.com/account/api and
put it in ~/.conda-smithy/circle.token
No appveyor token. Create a token at https://ci.appveyor.com/api-token and
Put one in ~/.conda-smithy/appveyor.token
No anaconda token. Create a token via
  anaconda auth --create --name conda-smithy --scopes "repos conda api"
and put it in ~/.conda-smithy/anaconda.token
Using Anaconda API: https://api.anaconda.org
Fetching package metadata ...........
Fetching package metadata ...........
Fetching package metadata ...........
Fetching package metadata ...........
Fetching package metadata ...........
Traceback (most recent call last):
  File "/home/marscher/anaconda/bin/conda-smithy", line 6, in <module>
    sys.exit(conda_smithy.cli.main())
  File "/home/marscher/anaconda/lib/python2.7/site-packages/conda_smithy/cli.py", line 289, in main
    args.subcommand_func(args)
  File "/home/marscher/anaconda/lib/python2.7/site-packages/conda_smithy/cli.py", line 200, in __call__
    configure_feedstock.main(args.feedstock_directory)
  File "/home/marscher/anaconda/lib/python2.7/site-packages/conda_smithy/configure_feedstock.py", line 461, in main
    render_README(env, config, forge_dir)
  File "/home/marscher/anaconda/lib/python2.7/site-packages/conda_smithy/configure_feedstock.py", line 204, in render_README
    fh.write(template.render(**forge_config))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 158: ordinal not in range(128)
marscher@jigsaw:~/sources/conda-forge/pyemma-feedstock$ python --version
Python 2.7.12 :: Continuum Analytics, Inc.
```

marscher:
conda-smithy version 1.4.6

jakirkham:
Is this with other feedstocks or just `pyemma`?

jakirkham:
[...]

PR #366, opened 2016-11-05
Use unicode on Python 2

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/364

This uses UTF-8 in all cases. Files are opened with UTF-8 encoding and unicode literals are used throughout. Fixes some issues on Python 2 in particular where metadata from the recipe included unicode characters. This should now correctly support them. Have tested against the [`pyemma` feedstock]( https://github.com/conda-forge/pyemma-feedstock ) with both Python 2 and Python 3 locally. Works fine in both cases. Should add that this ends up already being well covered in the existing test suite thanks to the use of the `touch_file` function, which had to receive a unicode literal to behave correctly.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
 
 from contextlib import contextmanager
 import os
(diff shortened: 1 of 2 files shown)


jakirkham:
cc @marscher

PR #371, opened 2016-11-07
PR: Check there is one empty line at the end of the file

goanpeca:
Checks that there is one and only one empty line in the `meta.yaml` file

@jakirkham


conda_smithy/lint_recipe.py
```diff
@@ -122,6 +122,20 @@ def lintify(meta, recipe_dir=None):
         lints.append('The recipe `license` should not include the word '
                      '"License".')
 
+    # 11: There should be one empty line at the end of the file.
+    if recipe_dir is not None and os.path.exists(meta_fname):
+        with open(meta_fname, 'r') as f:
+            lines = f.read().split('\n')
```
jakirkham: I think `splitlines` behaves a little better when it comes to Python 2/3 compatibility and unicode. However, I don't know that we will get the behavior we want then. Not sure what we should do about it.

Maybe we need start using unicode literals with the linter. That has fixed some problems in other areas. Also I expect this will fix some other linter issues with unicode that we have experienced.

conda_smithy/lint_recipe.py
```diff
@@ -122,6 +122,20 @@ def lintify(meta, recipe_dir=None):
...
+
+        end_empty_lines_count = 0
+        for i, line in enumerate(reversed(lines)):
```
jakirkham: Could use `start=1` with `enumerate` to simplify this a bit.

conda_smithy/lint_recipe.py
```diff
@@ -122,6 +122,20 @@ def lintify(meta, recipe_dir=None):
...
+            if line != '':
+                break
```
jakirkham: Wondering if we can reduce this some application of a function from `itertools`.

conda_smithy/lint_recipe.py
```diff
@@ -122,6 +122,20 @@ def lintify(meta, recipe_dir=None):
...
+            end_empty_lines_count = i + 1
```
jakirkham: With `start=1` this is then just `end_empty_lines_count = i`.

conda_smithy/lint_recipe.py
```diff
@@ -122,6 +122,20 @@ def lintify(meta, recipe_dir=None):
...
+
+        if lines[-1] != '' or end_empty_lines_count != 1:
```
jakirkham: Why do we need both checks? Are they not always the same?

conda_smithy/tests/test_lint_recipe.py
```diff
@@ -206,6 +206,46 @@ def test_redundant_license(self):
...
+                    fh.write(textwrap.dedent(content))
+                meta = {}
+                lints = linter.lintify(meta, recipe_dir=recipe_dir)
+                expected_message = ('There should be one empty line at the '
+                                    'end of the file.')
+
+                if content == valid_content:
+                    self.assertNotIn(expected_message, lints)
+                else:
+                    self.assertIn(expected_message, lints)
```
jakirkham: Nice tests! 👍

conda_smithy/lint_recipe.py
goanpeca: No, check 1 takes into account that there is no empty line at the end

Check 2 takes into account more than 1 empty line at the end

conda_smithy/lint_recipe.py
MSeifert04: You mean something like `sum(1 for _ in itertools.takewhile(lambda x: x == '', reversed(lines)))` but if that's really better? 😄

goanpeca: Hehe, yeah, maybe there is a super pythonic and readable way of making that a one liner, but yeah, I dont know if  it is really better. Thanks for the example @MSeifert04 :-)

conda_smithy/lint_recipe.py
jakirkham: So the trick I found at least is `splitlines` with one terminal newline and zero ends up being the same. 😞 Not sure if it is better in that case.

We can keep the the end line characters though. I'm just not sure if that makes it better or not.

conda_smithy/lint_recipe.py
```diff
@@ -122,6 +122,20 @@ def lintify(meta, recipe_dir=None):
         lints.append('The recipe `license` should not include the word '
                      '"License".')
 
+    # 11: There should be one empty line at the end of the file.
+    if recipe_dir is not None and os.path.exists(meta_fname):
+        with open(meta_fname, 'r') as f:
```
jakirkham: We should switch to `io.open`.

conda_smithy/lint_recipe.py
MSeifert04: Maybe two lines:
```
trailing_empty_lines = itertools.takewhile(lambda x: x == '', reversed(lines))
num_trailing_empty_lines = sum(1 for _ in trailing_empty_lines) # or len(list(trailing_empty_lines))
```
😅

goanpeca: :+1: :-)

conda_smithy/lint_recipe.py
goanpeca: Ah yes you are right check 2 is enough :-p

conda_smithy/tests/test_lint_recipe.py
```diff
@@ -206,6 +206,38 @@ def test_redundant_license(self):
...
+            b'extra:\r\n  recipe-maintainers:\r\n    - goanpeca',
+            # Two empty lines at the end of the file
+            b'extra:\n  recipe-maintainers:\n    - goanpeca\n\n',
+            b'extra:\r  recipe-maintainers:\r    - goanpeca\r\r',
+            b'extra:\r\n  recipe-maintainers:\r\n    - goanpeca\r\n\r\n',
+            # Three empty lines at the end of the file
+            b'extra:\n  recipe-maintainers:\n    - goanpeca\n\n\n',
+            b'extra:\r  recipe-maintainers:\r    - goanpeca\r\r\r',
+            b'extra:\r\n  recipe-maintainers:\r\n    - goanpeca\r\n\r\n\r\n',
+        ]
```
goanpeca: Do you feel safer now @jakirkham ?

goanpeca:
@jakirkham made the suggested changes

conda_smithy/lint_recipe.py
jakirkham: No worries. Was wondering again if I was missing something. 😄

conda_smithy/tests/test_lint_recipe.py
jakirkham: Sorry, was trying to save you from going down this road by trying to handle unicode cleanly throughout the linter and associated tests. ( https://github.com/conda-forge/conda-smithy/pull/372 )

I think it should be ok to make these explicitly unicode. Though I'll let you be the judge.

The `\r*` tests open an interesting question of should we permit this. 😄 Though I see no harm in allowing it for now and tightening down later. For a related problem, please see issue ( https://github.com/conda-forge/conda-forge-maintenance/issues/8 ).

goanpeca: Why not add `.gitattributes` to all repos upon creation and enforce a particular line ending?

Do you want me to change anything else at this point?

jakirkham:
Thanks @goanpeca .

conda_smithy/tests/test_lint_recipe.py
jakirkham: Nope. Looks good. Thanks.

Adding `.gitattributes` sounds like a good idea. Maybe we should write up an issue. Alternatively, if you would like to lead the way with a PR, that would be welcome too.

PR #379, opened 2016-11-11
Set all permissions

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/424

Sets all file permissions whether copied or written from scratch explicitly to non-executables. As we already have to explicit set execute permissions for Windows, it is not such a burden to have everything default to non-executable permissions. This should avoid a few more odd cases with file permissions.

cc @grlee77 @CJ-Wright @nicoddemus

### DIFF
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -54,6 +54,8 @@ def write_file(filename):
     if repo:
         repo.index.add([filename])
 
+    set_exe_file(filename, set_exe=False)
+
 
 def touch_file(filename):
     with write_file(filename) as fh:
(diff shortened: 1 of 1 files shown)


jakirkham:
Closing this for now.

jakirkham:
Resurrecting this PR because we are finding we need to ensure executable permissions are not set in certain cases unless we request them.

xref: https://github.com/conda-forge/conda-smithy/issues/424

jakirkham:
Refreshed this PR. Would appreciate if people experiencing these permission issues give it a look over, try it out, and share feedback. Thanks in advance.

jakirkham:
As there are less lines of code, the coverage obviously drops. However, the same functions are still covered. So the Coveralls failure can be ignored. Travis still passes.

jakirkham:
Think I'm going to close this. Can revisit if it is still important to others.

PR #380, opened 2016-11-11
Add gitattributes

jakirkham:
Adds a `.gitattributes` file for specifying how to handle line endings with recipe content.

cc @goanpeca

xref: https://github.com/conda-forge/staged-recipes/pull/1920

### DIFF
--- a/conda_smithy/feedstock_content/.gitattributes
+++ b/conda_smithy/feedstock_content/.gitattributes
@@ -1,0 +1,5 @@
+* text=auto
+
+meta.yaml text eol=lf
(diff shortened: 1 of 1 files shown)


jakirkham:
Strangely am encountering some issues when re-rendering where this file ends up with executable permissions. However, I don't think we are actually setting them. 😕

jakirkham:
Adding this [SO question]( http://serverfault.com/questions/429594/is-it-safe-to-write-batch-files-with-unix-line-endings ) as food for thought.

PR #390, opened 2016-11-22
Move build status and version info badges near the top of the README

nicoddemus:
When checking if packages are available in `conda-forge`, I often look in https://conda-forge.github.io/feedstocks.html and if positive, click on the link. That lands in the feedstock page, and often I also need to check the version, in which case I have to scroll all the way down to see the version badges.

Also I check the badges of feedstocks I maintain to see if all builds passed after merging a PR to update the version as sometimes CI might fail on `master` for flaky reasons even if the builds passed during the PR.

I was going to open an issue and propose to write a PR if the response was positive, figured out easier to just open the PR and discuss it here.


ocefpaf:
I like that.

jakirkham:
Thanks @nicoddemus .

PR #399, opened 2016-12-18
Fix source channel order in CIs

jakirkham:
Appears the source channel order in AppVeyor and Travis CI config files was backwards. This reverses them to get the channels in the right priority order. Namely channels showing up first in `channels/sources` have highest priority and those showing up last in `channels/sources` have lowest priority. Also the CircleCI docker build was sorting the channels instead of preserving the order. This also drops that sort.

### DIFF
--- a/conda_smithy/templates/run_docker_build.tmpl
+++ b/conda_smithy/templates/run_docker_build.tmpl
@@ -13,7 +13,7 @@ docker info
 config=$(cat <<CONDARC
 
 channels:
-{%- for channel in channels.get('sources', []) | sort %}
+{%- for channel in channels.get('sources', []) %}
  - {{ channel }}
 {%- endfor %}
  - defaults # As we need conda-build
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -45,7 +45,7 @@ install:
     - cmd: set "OLDPATH=%PATH%"
     - cmd: set "PATH=%CONDA_INSTALL_LOCN%\\Scripts;%CONDA_INSTALL_LOCN%\\Library\\bin;%PATH%"
     - cmd: conda config --set show_channel_urls true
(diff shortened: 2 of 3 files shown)


jakirkham:
This may be of some interest to you, @shadowwalkersb .

shadowwalkersb:
Good catch. Was getting to a point that would probably have bitten me. Right on time! Thanks!

conda_smithy/templates/run_docker_build.tmpl
```diff
@@ -13,7 +13,7 @@ docker info
 config=$(cat <<CONDARC
 
 channels:
-{%- for channel in channels.get('sources', []) | sort %}
+{%- for channel in channels.get('sources', []) %}
```
jakirkham: Seems I'm changing something you had added [before]( https://github.com/conda-forge/conda-smithy/pull/19/files#diff-00e77a6004c558c75ab3d1be6bfa56d9L16 ), @ericdill . Would be great to hear your feedback on this.

ericdill:
As long as the order is deterministic I think respecting the channel order makes sense. Iirc adding the `sorted` here  was an overzealous "sort all the dicts so they're deterministic" thing. Given that this is a list, sorting seems like a poor decision. :+1: for leaving the order as specified

jakirkham:
Thanks Eric. That was my impression of the change. FWIW all the other `sort`s seem good to me.

ericdill:
Thanks for the careful check. I hope not too much chaos ensued from mucking with the channel order...

jakirkham:
No chaos that I know of. Discovered as a product of trying to improve handling of channel sources.

ericdill:
Hah I guess we just got super luck then since the desired channel order is
```
- conda-forge
- defaults
```

which also happens to be in alphabetical order :relieved:

jakirkham:
Yep, we got super lucky in all of these cases. 🍀

PR #400, opened 2016-12-18
Handle defaults explicitly

jakirkham:
This adds `defaults` to the default set of channel sources for the feedstock. Also it makes sure that `defaults` does not get included in the CI unless it is explicitly added to channel sources for the feedstock.

Note: The constraint of requiring `defaults` to be added to channel sources explicitly (if the default setting are not used) is very important to ensure a proper interpretation of `defaults` as will be discussed in a latter PR.

Requires https://github.com/conda-forge/conda-smithy/pull/402

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -529,7 +529,8 @@ def main(forge_file_directory):
               'travis': {},
               'circle': {},
               'appveyor': {},
-              'channels': {'sources': ['conda-forge'], 'targets': [['conda-forge', 'main']]},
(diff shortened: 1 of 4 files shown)


jakirkham:
Would also be curious to hear your thoughts on this, @shadowwalkersb .

shadowwalkersb:
Defaults that can't be changed, not cool. Full control over channels without loosing sensible defaults, `conda-forge` and `defaults`, sounds pretty good to me! 👍

jakirkham:
Thanks for your feedback @shadowwalkersb . 😄

Issue #351, opened 2016-10-31
AppVeyor VS2008 slow startup/memory issues


jakirkham:
Seems there is a rather strange issue happening on the VS2008 builds unrelated to the recipe here. They seem to be taking a really long time to run and may be running out of memory as this build shows. Not sure if this is a `conda` problem (infinite recursion and so on) or a VM memory problem.

xref: https://ci.appveyor.com/project/conda-forge/iteration-utilities-feedstock/build/1.0.10/job/m42dxwpke0i6e5pt#L128
xref: https://ci.appveyor.com/project/conda-forge/conda-smithy-feedstock/build/1.0.89/job/1xvlseif9w9se8tr#L120

cc @conda-forge/core

jakirkham:
Should add the setup time for these VS2008 builds is right now ~10mins which is really going to hurt us on the larger scale. This is exceeding our ~2min or less startup time by 5-fold. It is really going to start costing us on the AppVeyor queue soon.

MSeifert04:
@jakirkham I have a Job already running for 50 minutes (a bit longer than the mentioned 10 minutes): https://ci.appveyor.com/project/conda-forge/iteration-utilities-feedstock/build/1.0.11.

The queue is empty right now so I haven't cancelled it. But if it holds up other jobs feel free to cancel it.

mwcraig:
FWIW, I am not seeing delays in python 2.7 jobs in other (non-conda-forge-related) appveyor jobs...

MSeifert04:
@jakirkham Here's the copy you asked for (https://github.com/conda-forge/iteration_utilities-feedstock/pull/3):

See for example (today):
- https://ci.appveyor.com/project/conda-forge/pytest-mock-feedstock/build/1.0.15
- https://ci.appveyor.com/project/conda-forge/hpc05-feedstock/build/1.0.10
- https://ci.appveyor.com/project/conda-forge/ipycache-feedstock/build/1.0.10

2 days ago:
- https://ci.appveyor.com/project/conda-forge/sentinels-feedstock/build/1.0.2

4 days ago:
- https://ci.appveyor.com/project/conda-forge/keepalive-feedstock/build/1.0.1

9 days ago:
- https://ci.appveyor.com/project/conda-forge/django-grappelli-feedstock/build/1.0.9

And those are just those I've looked into. So far a 100% hit rate.

11 days ago (**>>>no delay<<<**):
[...]

PR #329, opened 2016-10-14
Update `conda` on AppVeyor before activation

jakirkham:
Required to address issue ( https://github.com/conda-forge/conda-forge.github.io/issues/252 ).

Adds a hack to update `conda` on AppVeyor to a version that can handle a long path issue before activation. Please see PR ( https://github.com/conda/conda/pull/3349 ) for details about the fix.

Testing in PR ( https://github.com/conda-forge/conda-feedstock/pull/9 ).

cc @msarahan

PR #402, opened 2016-12-19
Tidy up AppVeyor conda update

jakirkham:
Closes https://github.com/conda-forge/setuptools-feedstock/pull/47

Some of the changes included in here install a very particular version of `conda`. This was needed to fix some path manipulation bugs. ( https://github.com/conda-forge/conda-smithy/pull/329 ) Since then newer version have been released that we use successfully. So it makes sense to remove the pinning. However, the update is still needed as AppVeyor's included `conda` is so old.

Also we found ourselves in a situation where Python 2.7 jobs where running out of memory with particular configurations of the root environment when running `conda-forge-build-setup`. ( https://github.com/conda-forge/conda-smithy/issues/351 ) Some hacks were added to adjust the `root` environment to fix this. ( https://github.com/conda-forge/conda-smithy/pull/370 ) Though these appear to be no longer necessary.

These changes are tried out in PR ( https://github.com/conda-forge/setuptools-feedstock/pull/47 ).

### DIFF
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -41,17 +41,10 @@ install:
     # Cywing's git breaks conda-build. (See https://github.com/conda-forge/conda-smithy-feedstock/pull/2.)
     - cmd: rmdir C:\cygwin /s /q
 
-    # Add our channels.
+    # Add a hack to update conda as the included copy is too old.
     - cmd: set "OLDPATH=%PATH%"
     - cmd: set "PATH=%CONDA_INSTALL_LOCN%\\Scripts;%CONDA_INSTALL_LOCN%\\Library\\bin;%PATH%"
-    - cmd: conda config --set show_channel_urls true
-    {%- for channel in channels.get('sources', [])|reverse %}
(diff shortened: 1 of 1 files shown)


jakirkham:
Could you please take a look at this, @MSeifert04 ?

MSeifert04:
Seems straight-forward, the title is a bit misleading the only change was unpinning `conda`, right?

Was the 4.1.12 conda version was responsible for the python 2.7 memoryerror or has it been found and fixed somewhere else?

jakirkham:
> Seems straight-forward, the title is a bit misleading the only change was unpinning `conda`, right?

Sort of. It boils down to the unpinning and moving the channel changes after `conda`.

> Was it the 4.1.12 conda version that lead to the python 2.7 memoryerror or has it been found and fixed somewhere else?

Unfortunately I can't claim we found the issue. 😞 There did seem to be some difference between `defaults` and `conda-forge`. If we installed the full stack (or as much as possible) from `conda-forge` to begin with, the issue seemed to go away. However, now the issue appears to not be reproducible. Perhaps something in `defaults` has been fixed.

jakirkham:
Thanks for looking at this @MSeifert04 . 😄

PR #403, opened 2016-12-25
pinned versions - lint and update on rerender

croth1:
Merry Christmas everyone,

I put together a short proof of principle of linting the pinned versions. The pinnings are defined in a json file and conda-smithy would become the primary place to define pinned versions.

The result would be e.g.

```
[...]
requirements:
  build:
    - zlib 1.2.*
    - freetype 2.6.*
    - fontconfig 2.12.*
  run:
    - zlib 1.2.*
    - freetype 2.6.*
    - fontconfig foooo
[...]
```

```
recipe has some lint:
  The build dependency 'freetype' should be pinned to version 2.7|2.7.*
  The run dependency 'freetype' should be pinned to version 2.7|2.7.*
  The run dependency 'fontconfig' should be pinned to version 2.12.*
```

❗️ The code might still crash if sections are missing. It's not meant for merging, rather for discussion.

This has a couple of implications:

* the code in [pin_the_slow_way.py](https://github.com/conda-forge/conda-forge.github.io/blob/master/scripts/pin_the_slow_way.py) would have to be adapted to import the dict of pinning instead of defining it
* smithy's version has to be bumped every time we want to make new pinnings visible to the update script.

Couple of things that might need further discussion:
* I feel that it may be worth considering moving`pin_the_slow_way.py` to smithy. At first sight it doesn't seem to be so different from the other utility scripts.


jakirkham:
Thanks for giving this some thought, @croth1 . This is definitely an important problem to solve. Linting pinnings is definitely a better solution than having reviewers check them alone. It should also make sure pinnings get update in a timely manner.

The only thing I'm unsure of is whether this behavior should live in `conda-smithy` or whether it should live in the linter webservice. What do you think? Maybe @pelson can share some thoughts as well.

xref: https://github.com/conda-forge/conda-forge-webservices

croth1:
Personally, I would put the pinnings in smithy as part of the general description of feedstocks. It would then probably be a good idea to ship the pinning updates directly with smithy rerender.

croth1:
@jakirkham, I ported more code from pin_the_slow_way. With the second commit the pinnings would be updated automatically when rerendering.

croth1:
@conda-forge/core, this is ready for a first round of comments and reviews

These are the changes:
* added a json file that contains the pinnings and a simple way to import it
* allowed for separate build/run dependency pins
* the linter now reports incorrectly pinned dependencies
* smithy rerender now also updates pinned dependencies

jakirkham:
So I'm still not sold on including the pinnings in `conda-smithy`. It means that releases of `conda-smithy` will need to be done to update the pinnings. This can have a negative impact on development as we may become unwilling to make big changes in `conda-smithy` (even if needed for bug fixes like we did with channels recently) because we want to ensure `master` is in a releasable state at any moment for a pinning change.

ATM `conda-smithy` feels too crowded to me and is already struggling with these sorts of problems. For instance, changes have been made to the linter, but we struggle with when those will be released or not as those changes may not line up with core changes to re-renderings or repo generation. Thus I'm eager to move the linter out of `conda-smithy` ( `conda-lint` ? ) and have not been the only person to suggest this. The initial goal of `conda-smithy` was just to be a cookie cutter for feedstocks, but everything we add to it that is not inline with that objective makes it a little harder to purse.

Could you please let me know your thoughts on the following questions? Where else could the pinnings live? Why should the pinning functionality live in `conda-smithy`? How does it fit with `conda-smithy`'s objective? If you think it doesn't fit with `conda-smithy`'s objective, where else should this functionality live?

croth1:
>  Where else could the pinnings live? 

I think there are two basic options to consider here: Do we want  1) the pinnings part of a python package, or 2) load them dynamically from a json document somewhere on the web?

cons: dynamic loading:
  * all tools dependent on pinnings require internet connection

cons: python package
 * making updates visible requires pushing a new release

> Why should the pinning functionality live in conda-smithy?

It was the closest match for now. I agree that if we want to go with the python package solution, creating a new package is also an option.

> How does it fit with conda-smithy's objective?  If you think it doesn't fit with conda-smithy's objective, where else should this functionality live?

I chose smithy mainly because it was easy accessible by the linter and in conda-forge.github.io. Didn't spend too much thought about other places yet.

---

What do you think of this alternative:

We go with basic option 2 and define the pinnings in some git repository. Loading the pinnings then requires downloading this json file. This logic could either stay in smithy or go in a new package. No smithy updates needed when pinnings are updated.

Downsides:
* linter will only work with internet connection

jakirkham:
Having the pinnings in another package sounds fine. Could live with the pin verification being part of the linter, but would want to make sure it would cleanly separate from `conda-smithy` when a separate linting program is made.

PR #407, opened 2017-01-03
Fix EOL formatting on Windows

jakirkham:
This configures git to normalize more line endings to avoid differences caused by a user's development environment (e.g. Windows vs. Unix). By doing this, we can be sure to get LF in all the cases we were already accustomed to having them. With Windows specific files (e.g. Batch files), we use CRLF as Windows would expect. This extends it beyond simply `bld.bat` as was the case before.

PR #408, opened 2017-01-04
Use shell quotes for the current platform on the commit hint after re-rendering

nicoddemus:
As discussed in #407

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -229,8 +229,8 @@ class Regenerate(Subcommand):
                         print("")
                     else:
                         print(
-                            "You can commit the changes with:\n\n"
-                            "    git commit -m 'MNT: Re-rendered with conda-smithy %s'\n" % __version__
+                            'You can commit the changes with:\n\n'
+                            '    git commit -m "MNT: Re-rendered with conda-smithy %s"\n' % __version__
                         )
                     print("These changes need to be pushed to github!\n")
                 else:


nicoddemus:
The Travis failure seems to be the same as in #407:

```
AttributeError: 'module' object has no attribute 'ci_register'
```

conda_smithy/cli.py
```diff
@@ -228,9 +228,11 @@ def __call__(self, args):
                         )
                         print("")
                     else:
+                        quote = "'" if not sys.platform.startswith('win') else '"'
                         print(
                             "You can commit the changes with:\n\n"
-                            "    git commit -m 'MNT: Re-rendered with conda-smithy %s'\n" % __version__
+                            "    git commit -m {q}MNT: Re-rendered with conda-smithy {version}{q}\n"
```
jakirkham: Honestly I'd be ok just using `"` if that is better for Windows. Both should work fine on Unix as there is nothing fancy going on in the message.

jakirkham: I know this was your contribution originally @minrk . Do you care if we just `"` or was there some special reason for `'` that is not coming to me now?

minrk: I only used `'` because the surrounding string is `"`, so I wouldn't have to escape. No deeper reason than that. Switching the inner and outer quotes should work fine, too, or escape the inner quotes.

nicoddemus: Thanks @jakirkham and @minrk for the feedback.

Updated.

jakirkham: Cool, thanks for confirming.

I couldn't remember if we were concerned about shell substitution at one point or not. Certainly shouldn't be a concern for this string.

jakirkham:
Toggling for CI.

jakirkham:
Thanks @nicoddemus and @minrk .

PR #385, opened 2016-11-15
Use Pytest

nicoddemus:
As discussed in #357.

I suggest to review each commit individually to see the progress. 

Initially I just changed the runner to `pytest` and all tests pass:

```
$ coverage run --source . -m pytest conda_smithy/tests
============================= test session starts ==============================
platform linux -- Python 3.5.2, pytest-3.0.4, py-1.4.31, pluggy-0.4.0
rootdir: /home/travis/build/nicoddemus/conda-smithy, inifile: 
collected 28 items 
conda_smithy/tests/test_configure_feedstock.py ..
conda_smithy/tests/test_feedstock_io.py ......
conda_smithy/tests/test_lint_recipe.py ....................
========================== 28 passed in 6.59 seconds ===========================
```

Next I updated all `self.assert*` methods to plain asserts using [unittest2pytest](pypi.org/project/unittest2pytest).

~~Now I will work on using some of pytest features to simplify the tests.~~ 😁 

Converted all tests to use pytest's features like parametrization and fixtures. 😁

PR #416, opened 2017-01-06
ENH: make too many/too few empty lines more verbose

CJ-Wright:


jakirkham:
@goanpeca , any thoughts on this tweak to the linter's end of line check?

goanpeca:
ehhh sure why not. but tests should be fixed, no?

jakirkham:
Yep, you're right. It will need test updates and possibly more tests to ensure we are covering both branches (though I thought we covered both cases before).

CJ-Wright:
I guess I had ran into the case where I had too many lines and didn't know it and was a bit confused by the message for a little while. The idea was that we have the data needed to make a more verbose comment and it doesn't cost us much to write it out so we might as well be more verbose, it could save someone the confusion over of they have too few or too many lines. In retrospect we could even output exactly how many extra/fewer lines they have, although that was not in this PR.

goanpeca:
> In retrospect we could even output exactly how many extra/fewer lines they have, although that was not in this PR.

I did not pursue this originally (I introduced the change you are improving) because of the way we run the test cases (we expect a fixed message in the tests not a variable one...) but I guess we can tweak the tests a bit to make this work.

jakirkham:
Yep, I'm ok with this sort of change. Just need to update the tests.

goanpeca:
@jakirkham this way of running tests is turning a bit cumbersome. We should make these messages warning classes so it is easier to test.

Thoughts?

jakirkham:
Seems like a good idea. Could you please add it to a new issue?

CJ-Wright:
If you are using pytest you might also look at [pytest.mark.parametrize ](http://doc.pytest.org/en/latest/parametrize.html)

goanpeca:
@CJ-Wright, what I meant is that we can no longer user messages and test based on that. We need to move to warnings/exceptions. So this is not a refactor for the tests only, but for how the code is written.

jakirkham:
Sort of tangential, PR ( https://github.com/conda-forge/conda-smithy/pull/385 ) would switch us to `pytest`. Currently we are using `unittest`.

goanpeca:
cool I really prefer pytest over unittest

conda_smithy/tests/test_lint_recipe.py
```diff
@@ -233,16 +233,25 @@ def test_end_empty_line(self):
...
+                                        'at the end of the '
+                                        'file.'.format(lines - 1))
+                else:
+                    expected_message = ('There are too few lines.  '
+                                        'There should be one empty line at'
+                                        ' the end of the file.')
                 if content == valid_content:
                     self.assertNotIn(expected_message, lints)
                 else:
+
```
jakirkham: Can we drop this empty line?

jakirkham:
Excepting one small nit. This looks fine to me.

@goanpeca, do you have any more thoughts?

goanpeca:
LGTM :+1:

jakirkham:
Once the nit is fixed, @CJ-Wright , will happily merge.

jakirkham:
Thanks @CJ-Wright. This won't be in `2.0.0`, but it will be in whatever release we do next.

jakirkham:
Thanks @CJ-Wright. This will be in `2.0.0`, but it will be in whatever release we do next.

jakirkham:
Thanks @CJ-Wright. This will be in `2.0.0`.

jakirkham:
Have released this to the linter webservice. So we should start seeing warnings for issues like this.

CJ-Wright:
Thank you @jakirkham !

PR #421, opened 2017-01-09
Drop Python 3.4 support

jakirkham:
Note: This does not need to be done right away. Just putting it up here for consideration.

As Python 3.6 is being included throughout feedstocks and Python 3.5 is still in use, it makes sense for us to drop our oldest Python 3 that still remains. The purpose of doing this is to cutdown on the CI burden from building so many things and maintenance burden associate with this and the older Python. So here we drop Python 3.4 from the matrix.

xref: https://github.com/conda-forge/staged-recipes/pull/2179


ocefpaf:
Complementary to #2179

Issue #431, opened 2017-01-14
`--user github_user_name` not passed to ci_support/run_docker_build.sh 


guyer:
Despite conda-forge.yml having
```
channels:
targets:
- [github_user_name, main]
```
ci_support/run_docker_build.sh still says
```
upload_or_check_non_existence /recipe_root conda-forge --channel=main || exit 1
```

jakirkham:
Did you [re-render]( https://github.com/conda-forge/conda-smithy#re-rendering-an-existing-feedstock )? Just tested this myself and am not seeing this behavior.

guyer:
Yes, I rerendered. Nothing changed.

jakirkham:
Assuming this is the same feedstock as the other issue, this is also behaving correctly. Please see this [line]( https://github.com/guyer/pysparse-feedstock/blob/master/ci_support/run_docker_build.sh#L50 ) for an example. Am I missing something else? 😕

guyer:
Like the other, I [added that manually](https://github.com/guyer/pysparse-feedstock/commit/a25c5e52dc491bfe6db107bd6d511bbcd8e034fb#diff-072cea419befcfca383c45968d4c00b5) so that I could actually get things working.

jakirkham:
The `conda-forge.yml` doesn't appear to be formatted correctly. The `channels` section needs to be properly indented.

ref: https://github.com/guyer/pysparse-feedstock/blob/master/conda-forge.yml#L2

guyer:
Ah, thanks. I copied from https://github.com/conda-forge/conda-smithy/blob/master/README.md, but it's not indented by github markdown.

jakirkham:
Good point. Feel free to submit a PR to fix that.

In light of the indentation fix, do these problems go away or are there other issues?

guyer:
Indentation fixed both issues, thanks. 

I can't tell why README.md doesn't indent properly. The raw text looks fine.

jakirkham:
Looks like they were indented using one space instead of two. Using two spaces seems to fix it. Thanks for point that out. Fixing in PR ( https://github.com/conda-forge/conda-smithy/pull/432 ).

jakirkham:
Have corrected the README. Thanks again for catching that.

PR #432, opened 2017-01-15
DOC: Fix indentation in target channel example

jakirkham:
Appears there was one space missing from each indented line. This appears to have resulted in a weird rendering of the code block by GitHub, which misses any of the indentation. Adding one more space to each indented line as done here should fix that.

Thanks to @guyer for [spotting this]( https://github.com/conda-forge/conda-smithy/issues/431#issuecomment-272671303 ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,8 +70,8 @@ of organization with `--user github_user_name`.
 Optionally, you can choose a channel to upload to in `conda-forge.yml`.
   ```
   channels:
-   targets:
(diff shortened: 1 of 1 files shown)


jakirkham:
The fix can be seen [here]( https://github.com/conda-forge/conda-smithy/blob/7b80915274d6b2d0f83994507802cd59e4ac0a54/README.md#making-a-new-feedstock ) under item 4.

PR #446, opened 2017-01-23
feedstocks list: organisation -> organization

jakirkham:
Fix help message to match actual flag name.


jakirkham:
Got fooled by this earlier and wanted to make sure nobody else made the same mistake.

PR #454, opened 2017-02-08
verify that docker script reached its end

minrk:
protects against early termination appearing as success.

This is one small piece of #337, but might be reviewed and merged more easily because it doesn't change anything, just adds a check that the script finished executing.

This would at least turn the cases fixed by #337 into failures, rather than mysteriously missing packages.

### DIFF
--- a/conda_smithy/templates/run_docker_build.tmpl
+++ b/conda_smithy/templates/run_docker_build.tmpl
@@ -58,4 +58,5 @@ EOF
 # double-check that the build got to the end
 # see https://github.com/conda-forge/conda-smithy/pull/337
 # for a possible fix
+set -x
 test -f "$FEEDSTOCK_ROOT/build_artefacts/conda-forge-build-done" || exit 1
--- a/conda_smithy/templates/run_docker_build.tmpl
+++ b/conda_smithy/templates/run_docker_build.tmpl
@@ -25,6 +25,8 @@ show_channel_urls: true
 CONDARC
 )
 
+rm -f "$FEEDSTOCK_ROOT/build_artefacts/conda-forge-build-done"
+
 cat << EOF | docker run -i \
                         -v "${RECIPE_ROOT}":/recipe_root \
                         -v "${FEEDSTOCK_ROOT}":/feedstock_root \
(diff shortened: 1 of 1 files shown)


jakirkham:
I think if we even just did `|| exit 1` after the `docker run` command we should get the same behavior more simply, no?

minrk:
I think that there already is an `|| exit $?` on the `docker run`. One of the reasons this has slipped through is that whatever it was that mpiexec does, results in a *clean* (status=0) premature exit, rather than a nonzero status.

minrk:
My rough understanding of what's happening:

- mpiexec consumes (or closes) stdin when it is called
- when `bash` resumes control, it reaches EOF without ever seeing commands issued after mpiexec, and exits as if everything is AOK

jakirkham:
Could it be that using `$?` instead of `1` is part of the problem. I'm not entirely sure why we bother with `$?` as I don't believe this is not visible to the user.

minrk:
I don't think so. The `||` clause never executes because `docker run` exits with status 0 in these cases, indicating success.

minrk:
[this build](https://circleci.com/gh/minrk/petsc-feedstock/17) was run after rendering with this PR, confirming that `docker run` exits cleanly and the `|| exit 1` is never reached. Only the canary check prevents the build from claiming to have succeeded.

jakirkham:
Thanks for the link. I agree with your conclusion.

jakirkham:
@gqmelo, what do you think about this change?

jakirkham:
Thanks @minrk. Thanks everyone who helped review.

conda_smithy/templates/run_docker_build.tmpl
```diff
@@ -50,5 +52,11 @@ conda build /recipe_root --quiet || exit 1
 {%- endfor -%}
 {%- endblock -%}
 
+touch /feedstock_root/build_artefacts/conda-forge-build-done
 EOF
```
jakirkham: Missed that we did not include another empty line below. As a result, this means the terminal newline would be dropped from this file. Fixing it in PR ( https://github.com/conda-forge/conda-smithy/pull/459 ).

conda_smithy/templates/run_docker_build.tmpl
```diff
@@ -50,5 +52,11 @@ conda build /recipe_root --quiet || exit 1
...
 
+# double-check that the build got to the end
+# see https://github.com/conda-forge/conda-smithy/pull/337
+# for a possible fix
+set -x
+test -f "$FEEDSTOCK_ROOT/build_artefacts/conda-forge-build-done" || exit 1
```
jakirkham: An empty line is added after here in PR ( https://github.com/conda-forge/conda-smithy/pull/459 ) so the re-rendered file has a terminal newline.

PR #457, opened 2017-02-12
Combine CircleCI re-rendering steps

jakirkham:
Previously the re-rendering of `ci_support/run_docker_build.sh` and `circle.yml` occurred in two different steps. This combines these two steps into one. Meaning one less retrieval from the index to determine the contents of the matrix. As a result, there are exactly as many calls to the index as there are target architectures. This should speed up re-rendering a little bit and simplify the codebase a bit as well.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -67,7 +67,7 @@ def meta_config(meta):
     return config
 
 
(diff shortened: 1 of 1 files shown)


jakirkham:
After a bit of testing locally, this seems to work fine. It basically preserves the ordering of the two re-rendering steps. Merely combines them into one function and deduplicates some things that they shared in common.

PR #458, opened 2017-02-12
fix typos in clone functions

kain88-de:
There as a mixup in the lines. This caused the `clone_feedstock` function to fail because the clone variable doesn't exist. This also caused a silent error in `clone_all`  in the multiprocessing. Now it should work. Btw if we replace the explicit multiprocessing code with joblib the exception will always be thrown for us in the main thread.

### DIFF
--- a/conda_smithy/feedstocks.py
+++ b/conda_smithy/feedstocks.py
@@ -73,7 +73,7 @@ def fetch_feedstocks(feedstock_directory):
 
 
 def feedstocks_list_handle_args(args):
-   for repo in feedstock_repos(args.organization):
+    for repo in feedstock_repos(args.organization):
         print(repo.name)
 
 
@@ -83,7 +83,7 @@ def clone_feedstock(feedstock_gh_repo, feedstocks_dir):
     clone_directory = os.path.join(feedstocks_dir, repo.name)
     if not os.path.exists(clone_directory):
         print('Cloning {}'.format(repo.name))
-        new_repo = Repo.clone_from(repo.clone_url, clone_directory)
+        clone = Repo.clone_from(repo.clone_url, clone_directory)
         clone.delete_remote('origin')
     clone = Repo(clone_directory)
     if 'upstream' in [remote.name for remote in clone.remotes]:


jakirkham:
Could you please take a look @dkillick?

conda_smithy/feedstocks.py
```diff
@@ -83,9 +83,9 @@ def clone_feedstock(feedstock_gh_repo, feedstocks_dir):
     clone_directory = os.path.join(feedstocks_dir, repo.name)
     if not os.path.exists(clone_directory):
         print('Cloning {}'.format(repo.name))
-        new_repo = Repo.clone_from(repo.clone_url, clone_directory)
-        clone.delete_remote('origin')
+        Repo.clone_from(repo.clone_url, clone_directory)
     clone = Repo(clone_directory)
+    clone.delete_remote('origin')
```
jakirkham: Are we sure this is a good idea? It seems like the remote could potentially be undefined here (was already cloned and remote removed); whereas, before we knew `origin` is defined. If we do move it down here, there should at least be a `try...except`. Would be a `GitCommandError` that we need to catch IIRC.

jakirkham:
TBH I don't really use this script. Instead I use the [feedstocks repo]( https://github.com/conda-forge/feedstocks ), which gets updated daily. So I don't have strong feelings about the changes here (excepting the one issue noted above). Would be nice if someone that did could weigh in.

conda_smithy/feedstocks.py
kain88-de: I could also remove the remote in the if-clause then the origin remote would exist for sure.

kain88-de:
This script is used by your `MNT: rerender` bot. It works now because the repositories are likely already checked out but could fail with downloading new repositories.

jakirkham:
Correct, am aware of that. Was just mentioning the other one in case you didn't already know.

Yeah, this is part of rebuilding the service. It runs `feedstocks` to clone all repos. Actually we are getting close to size limitations on Heroku. So we may need to change how the service is built in the future.

Though TBH am not clear why we need to remove the remote.

conda_smithy/feedstocks.py
```diff
@@ -84,7 +84,7 @@ def clone_feedstock(feedstock_gh_repo, feedstocks_dir):
     if not os.path.exists(clone_directory):
         print('Cloning {}'.format(repo.name))
         new_repo = Repo.clone_from(repo.clone_url, clone_directory)
-        clone.delete_remote('origin')
+        new_repo.delete_remote('origin')
```
jakirkham: Maybe it would just be better to make `new_repo` -> `clone` in both cases. It would be overwritten in the end, but by nothing different.

kain88-de:
I'm not to sure either. The other code just wants there to be a remote called upstream. It doesn't care if an origin remote exists.

jakirkham:
> I'm not to sure either. The other code just wants there to be a remote called upstream. It doesn't care if an origin remote exists.

Sorry, not to trying to send us down that path. Just musing on this code as it is a bit old and probably could benefit from some cleanup.

Another oddity is a few lines down we create a process pool where we find the minimum between 10 * # CPUs and # of feedstocks. As if there is a computer with enough CPUs to be larger than feedstocks. 😆

kain88-de:
yeah there are some oddities in there but i wanted to keep my change small.

kain88-de:
@jakirkham the builds have passed now.

jakirkham:
Thanks @kain88-de.

PR #460, opened 2017-02-19
Simplify feedstock repo initialization

jakirkham:
As we already stage changes to a feedstock's git repo (should it exist), it makes sense that we actually create a git repo if we will need one before we add content to the feedstock. By doing this, we can simplify and cut redundant code from `conda_smithy.cli` that is already handled effectively in `conda_smithy.configure_feedstock`. Thus reducing our chances of code rot and also avoiding contaminating any other repo.

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -11,6 +11,7 @@ from conda_build.metadata import MetaData
 
 from . import ci_register
 from . import configure_feedstock
+from . import feedstock_io
 from . import lint_recipe
 from . import __version__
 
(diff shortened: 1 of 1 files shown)


jakirkham:
cc @shadowwalkersb

shadowwalkersb:
This seems to have resolved #437 .

conda_smithy/cli.py
```diff
@@ -114,9 +92,12 @@ def __call__(self, args):
         msg = 'Initial feedstock commit with conda-smithy {}.'.format(__version__)
 
         try:
+            os.makedirs(feedstock_directory)
```
jakirkham: In addition to ensure there is a directory to create a repo in and add stuff to, should add this ensures we don't accidentally stomp on a preexisting feedstock as it will raise an `OSError` if we do.

jakirkham:
Thanks for taking a look, @shadowwalkersb.

jakirkham:
FYI this went out in the 2.1.1 release.

PR #463, opened 2017-02-25
Allow to tweak docker executable in run_docker_build

sdvillal:
This is a minimal change that adds *docker.executable* to the configurable parameters of *run_docker_build*. The goal is to be able to change the command to *nvidia-docker*, so that gpus get exposed during build, and specially test, whenever needed.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -645,7 +645,9 @@ def main(forge_file_directory):
     build_config.CONDA_PY = 10
 
     recipe_dir = 'recipe'
-    config = {'docker': {'image': 'condaforge/linux-anvil', 'command': 'bash'},
+    config = {'docker': {'executable': 'docker',
+                         'image': 'condaforge/linux-anvil',
+                         'command': 'bash'},
               'templates': {'run_docker_build': 'run_docker_build_matrix.tmpl'},
               'travis': {},
               'circle': {},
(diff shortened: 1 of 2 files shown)


jakirkham:
LGTM. Thoughts @minrk @gqmelo?

minrk:
:+1:

gqmelo:
Cool, but when the docker executable is set to nvidia-docker? Will this be done on a later step?

sdvillal:
In ```conda-forge.yml```. We are using this for building some cuda based packages. For example, [here](https://github.com/loopbio/caffe-feedstock/blob/40eaf60f34d1ac664aee86c4acd780f70f5e195f/conda-forge.yml), that renders to [this](https://github.com/loopbio/caffe-feedstock/blob/cdd72e000e7271aa5c3ad0934c72518fee2dc306/ci_support/run_docker_build.sh#L30-L35). We then just build internally, as the public CIs are of no use to test cuda code.

As a side note, when finished I will report a bit about our experiences building packages with cuda/cudnn support.

jakirkham:
> As a side note, when finished I will report a bit about our experiences building packages with cuda/cudnn support.

Please do.

We were looking into this early on. Though we were a bit nervous about licensing and distribution of CUDA, CUDNN, or things linked to them at the time. Unfortunately we didn't hear back from our NVIDIA contact on these points, so it got backlogged. Would be great to get that discussion going again.

jakirkham:
Thanks everyone.

PR #466, opened 2017-02-27
Manage teams

jakirkham:
The goal of this change is to handle team management at the level of `conda-smithy`. Particularly as part of GitHub registration. Currently team management is handled by two different scripts. One in staged-recipes, which populates the maintainer teams on feedstock creation. The other is part of the team and feedstocks update script. As we may need to replace the latter with something soon and there is already code duplication, it seems better to consolidate the working implementation in `conda-smithy` and have both scripts leverage it.

In addition to cutting down repeated code, this provides us the option going forward of handling team updates in a different way than we have thus far. In particular, we could use a web service to update teams as changes are made. This should provide us options that don't require scanning and updating all teams over the org in one cron job, which has been extremely problematic (subject to frequent failure, easily messed up by feedstock local changes, often running into GitHub rate limit, very rarely completing). However, these changes don't necessitate going in that particular direction, it does open the door for allowing us to explore different more viable options easily.

xref: https://github.com/conda-forge/conda-forge.github.io/issues/214
xref: https://github.com/conda-forge/conda-forge-webservices/issues/63

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,9 @@ directory called `foo-feedstock`, populates it with CI setup skeletons, adds the
 2. **Create a github repo:** `conda smithy register-github --organization conda-forge ./foo-feedstock`.
 This requires a github token. You can try it out with a github user account
 instead of an organization by replacing the organization argument with
-`--user github_user_name`.
+`--user github_user_name`. If you are interested in adding teams for your feedstocks,
+you can provide the `--add-teams` option to create them. This can be done when creating
+the feedstock or after.
 3. **Register the feedstock with CI services:**
 `conda smithy register-ci --organization conda-forge --feedstock_directory ./foo-feedstock`.
 This requires tokens for the CI services. You can give the name of a user instead
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -111,6 +111,10 @@ class RegisterGithub(Subcommand):
         #  conda-smithy register-github ./ --organization=conda-forge
         super(RegisterGithub, self).__init__(parser, "Register a repo for a feedstock at github.")
         scp = self.subcommand_parser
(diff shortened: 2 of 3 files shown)


jakirkham:
@shadowwalkersb @sdvillal, would you mind taking a look? Basically this here to simplify team management at conda-forge, but I want to make sure it is not going to mess something up for external users.

To simply explain, it creates a team for every feedstock named after the package. These teams get read and write permissions to the feedstock in question. In addition, an `all-members` team is created that holds everyone that is a feedstock maintainer.

Was debating about requiring a flag to enable this feature.

shadowwalkersb:
This looks like it is very conda-forge specific, `all-members`, certainly, is. Not sure that external users would care about teams, unless, it is an org. Even then, they can do it manually. So, I wouldn't make it the default. Requiring a flag sounds more reasonable.

jakirkham:
Just FYI teams are for orgs only. So this will have no effect on a user doing this sort of thing to their own account.

jakirkham:
Broke out some minor changes from this PR and put together as PR ( https://github.com/conda-forge/conda-smithy/pull/478 ).

jakirkham:
Thanks for the feedback @shadowwalkersb. Have changed this to only run if a flag `--add-teams` is provided to `register-github`. FYI this can be done either at the same time as the feedstock is created or after the fact. If you do become interested in this feature, we can probably play with it to allow for turning off the `all-members` team while keeping the individual feedstock teams, but do not plan on addressing this now.

jakirkham:
Included a little bit of documentation in the README to explain how to use this feature.

jakirkham:
@conda-forge/core, would appreciate your feedback on this change. Thanks. 😄

PR #475, opened 2017-03-10
Comments

CJ-Wright:
Search for comments in the meta.yaml

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -149,6 +149,14 @@ def lintify(meta, recipe_dir=None):
     except RuntimeError as e:
         lints.append(str(e))
 
+    # 13: There should be no comments
+    if recipe_dir is not None and os.path.exists(meta_fname):
+        with io.open(meta_fname, 'r') as f:
+            lines = f.read().split('\n')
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@conda-forge/conda-smithy Ready for review!

conda_smithy/lint_recipe.py
```diff
@@ -149,6 +149,14 @@ def lintify(meta, recipe_dir=None):
     except RuntimeError as e:
         lints.append(str(e))
 
+    # 13: There should be no comments
```
ocefpaf: We must allow, and in cases like patches even require, for some comments.

CJ-Wright: Ok is there a good way to limit that down?
I have gotten a fair amount of boilerplate requests from reviewers to remove comments from my recipes.

jakirkham:
I think the challenge is people keep comments from the `example` recipe by accident. Thus reviewers spend time ask that this be removed as they are clearly remnants. Maybe there is a better way we could document this so that people understand to do this cleanup themselves? Alternatively a better tool of some form for generating compliant recipes would remove the need for comment cleanup. Just some options to consider.

CJ-Wright:
I am going to close this and move any relevant changes over to conda-verify

PR #476, opened 2017-03-11
Remove homebrew by just moving /usr/local

gqmelo:
Sometimes the uninstall script takes too long, I've seen it taking ~130s
on some jobs.

### DIFF
--- a/conda_smithy/templates/travis.yml.tmpl
+++ b/conda_smithy/templates/travis.yml.tmpl
@@ -39,10 +39,9 @@ before_install:
     - |
       echo ""
       echo "Removing homebrew from Travis CI to avoid conflicts."
-      curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall > ~/uninstall_homebrew
-      chmod +x ~/uninstall_homebrew
-      ~/uninstall_homebrew -fq
-      rm ~/uninstall_homebrew
+      sudo mv /usr/local /usr/local.old
+      hash -r
+      if brew --version; then exit 1; fi
 
 
 install:


gqmelo:
I've created a branch on psutil-feedstock to test this changes: https://github.com/gqmelo/psutil-feedstock/tree/travis-homebrew

And the travis jobs: https://travis-ci.org/gqmelo/psutil-feedstock/builds/210134915

jakirkham:
Are you sure it was the uninstall script or was it the old removal operations? That said, I get that it sucks to spend time removing things. Though I think fiddling with this every few months is not very productive. Especially if we get wrong and have to start debugging bad packages. Nor will it address key issues like why we spend so much time waiting for builds to start. We should put pressure on Travis CI to give us an option to exclude Homebrew. Especially if we start paying them for services.

xref: https://github.com/travis-ci/travis-ci/issues/7190
xref: https://github.com/conda-forge/conda-forge.github.io/issues/306

jakirkham:
To provide another near term course of action, maybe it is worth taking a look at that uninstall script to make it faster. For instance, can we work with Homebrew to parallelize removal of files.

jakirkham:
Also could you please show where you are getting the ~130s from? I'm seeing ~20s typically (often less).

gqmelo:
> Also could you please show where you are getting the ~130s from? I'm seeing ~20s typically (often less).

I've noticed the problem when updating psutil:

https://travis-ci.org/conda-forge/psutil-feedstock/builds/209500926

On that PR, uninstalling homebrew took from 61s up to 133s. After that I created a branch to try some optimizations directly in the feedstock and got large durations too:

https://travis-ci.org/gqmelo/psutil-feedstock/builds/209865383

That is why I was motivated to submit this PR. But now looking better on the time spent on other builds, you are right, it normally takes around 20s. So I'm not sure what happened on these builds, probably travis had a degraded performance by the time, specially if we consider that everything was slower (conda build command took 170s on one case when it normally takes 70s).

> Though I think fiddling with this every few months is not very productive. Especially if we get it wrong and have to start debugging bad packages.

Ok, but how confident we are that the uninstall script works? Is it used so often? If homebrew only installs things on ```/usr/local```, moving or removing the directory seems even safer than relying on the script.

> Nor will it address key issues like why we spend so much time waiting for builds to start.

Yeah, but I suppose we probably wait so much because we have too many builds and travis algorithm won't let us monopolize their machines. If this is true, eliminating any significant overhead could help on the wait time.

> We should put pressure on Travis CI to give us an option to exclude Homebrew. Especially if we start paying them for services.

This would be the right thing, and would also solve the problem I've seen when using the cache on #477, where a homebrew update is triggered after a cache is downloaded. But how likely is this to happen? They do not seem to be very happy in implementing this.

> To provide another near term course of action, maybe it is worth taking a look at that uninstall script to make it faster. For instance, can we work with Homebrew to parallelize removal of files?

I think it would be nice to take a look in the script and try to optimize it, but I'm afraid I can't help much with this as I don't even have a Mac. Besides that, an official uninstall script needs to uninstall everything cleanly, something we don't need to worry about, as we don't care about anything installed on ```/usr/local```

That said, I'm ok if you want to hold this PR, I really created it in the context where uninstalling homebrew could take more than 2 minutes. That not being true makes this issue becomes much less urgent. But if travis does not provide images without homebrew, I do think it is worth the risk to just move ```/usr/local``` as ```20s``` is still too much time and can be 10% of the time spent on building a lot of packages (psutil for example).

jakirkham:
So I don't have time to respond to all of your points directly, but have tried to respond to many of them generally below.

It seems the moral of the story is this is not as big of a waste as expected. Given that, it seems to me that there are other worthwhile efforts to spend time on that will net us bigger savings. Happy to provide more info on known issues if you are interested. That said, I think your caching PR is definitely approaching one of them. Would be interested in working with you on that. Especially given the install and upgrade of Miniconda takes ~60s. Perhaps caching will also mitigate some of the (download?) issues you experienced before.

Beyond that I think you have raised some points about what is required when clearing out Homebrew. We initially did our own sort of "uninstall" step for Homebrew. From my notes, it appears this was somewhat variable in time, but could typically take ~30s. So the uninstall script was an improvement, albeit a small one. It also reduced the log size, which is helpful for any build running into log size limits.

That said, the real motivation was for switching to the uninstall script was to have a clean, stable way to clear Homebrew. Ideally something that we wouldn't need to muck with. The uninstall script is [recommend for uninstalling Homebrew]( https://github.com/Homebrew/brew/blob/master/docs/FAQ.md#how-do-i-uninstall-homebrew ) and is maintained by Homebrew. They often recommend it to Homebrew users wanting to do a fresh install in their forums. Had raised an issue about something that wasn't being uninstalled and they fixed it right away. Given all of this, the Homebrew uninstall script really fits our needs.

The history of uninstalling Homebrew at conda-forge is one that has been driven by making sure our build environment was not contaminated. Other than you and I, I'm not aware of anyone complaining about the time spent on this step. We had to rebuild several packages the last time Homebrew was an issue. While it was relatively simple and conda-forge was much smaller then, it still took a bit of coordination to fix things. Having something go wrong with that uninstall step now would be catastrophic in terms of debugging and rebuilding things. As such, I see a much greater detriment to changing this step than leaving it alone.

gqmelo:
Ok, no problem, I understand you don't want to risk wasting time debugging any issue might cause, specially considering you would be one of the most involved people in debugging this.

Sorry for my lack of knowledge about the homebrew uninstall process. As it was a separate script I thought it was more a workaround than an official and supported way to do it.

And I was just curious if anything was left in ```/usr/local``` after uninstalling homebrew and added a ```find /usr/local``` on my branch. This is the result:

https://travis-ci.org/gqmelo/psutil-feedstock/jobs/210581632

There are some files, although not related to homebrew. Anyway, do you think it is a good idea moving ```/usr/local``` or maybe some specific files like ```/usr/local/lib/node_modules``` after executing the uninstall script? I could change this PR or create another one for that.

jakirkham:
No worries at all. It is always good to revisit the assumptions we have made.

That's concerning. FWICT the node modules, at least, appear to be a leftovers from Homebrew. Raised issue ( https://github.com/Homebrew/install/issues/96 ) with Homebrew.

It looks like there are some other things in there that might be due to Homebrew. If you could follow up with them about these, that would be great. Basically I'd rather Homebrew uninstall work be done by the Homebrew team in that script if at all possible.

We can then sort out whatever is left after Homebrew has addressed what they can. Probably worth putting it in a new issue here once we are ready.

gqmelo:
Ok, looking more closely at the files, it seems that only ```oclint*``` are homebrew leftovers. I suspect it is because it was installed with ```brew cask```. I raised an [issue](https://github.com/Homebrew/install/issues/97) for that.

Besides that and the node stuff I think the other files are ok, as they are located on ```etc``` and ```var``` and I wouldn't expect them to be removed.

I'm also closing this PR for now.

jakirkham:
Ok, that makes sense. Cask removal was added only very recently. So it is possible some subtleties were missed. Though, as you note in that issue, this might not be easy to solve.

I think you are right w.r.t. the other stuff. It looks like remnants of `openssl` and `postgres`. Went ahead and pointed them out in issues ( https://github.com/Homebrew/install/issues/98 ) and ( https://github.com/Homebrew/install/issues/99 ) just so Homebrew is aware. Though it shouldn't matter to us if they are removed or not.

Thanks @gqmelo.

PR #482, opened 2017-03-17
lint pinnings and add conda smithy update-pinning

croth1:
This will work if we decide to merge  https://github.com/conda-forge/staged-recipes/pull/2267

the commit
  * fixes https://github.com/conda-forge/conda-forge.github.io/pull/310 and implements https://github.com/conda-forge/conda-forge.github.io/issues/270
  * gives the linter the capability to check if the pinning is correct
  * adds a new `update-pinning` command to conda smithy to automatically update all pins

### DIFF
--- a/conda_smithy/__init__.py
+++ b/conda_smithy/__init__.py
@@ -1,3 +1,4 @@
 from ._version import get_versions
+
 __version__ = get_versions()['version']
 del get_versions
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -1,7 +1,6 @@
 from __future__ import print_function, absolute_import
 
 import os
-import requests
 import subprocess
 import sys
 import time
@@ -13,10 +12,14 @@ from . import ci_register
 from . import configure_feedstock
 from . import feedstock_io
 from . import lint_recipe
+from . import update_pinning
 from . import __version__
+
+from cf_pinning import __version__ as pinning_version
 
 
 PY2 = sys.version_info[0] == 2
+
 
 def generate_feedstock_content(target_directory, source_recipe_dir, meta):
     target_directory = os.path.abspath(target_directory)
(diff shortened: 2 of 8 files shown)


jakirkham:
Toggling to restart CI.

jakirkham:
I like this solution, @croth1. This is very cool. I'll try to take some time to look at it more in the coming week or so.

Tried to make some changes to how we are doing testing on CI to hopefully make it more clear what is causing the failure here.

jakirkham:
Also, in light of the movement of this code to `conda-smithy`, do we actually need code to read the pinnings external from this?

croth1:
We had a similar discussion a couple of months ago in which we argued that the pinnings should be independent of smithy. IIRC correctly the argument was that this would speed up smithy's release cycle too much. I'll link the discussion when I find it.

We then decided to move the pinnings to an extra package. We can revert this. I actually don't have a strong opinion on that.

Having the pinnings as json might have a future application at the moment all tools that require pinnings already depend on smithy.

jakirkham:
Sorry, I should clarify. What I mean is the code reading the pinnings could be moved to `conda-smithy`, but the pinnings themselves could stay a separate package. Or is there something else you are thinking about here?

croth1:
Ah again either way is fine - we can move them to smithy or leave them in the pinning package.

We will reuse these functions in the web service that sends pull requests after pin updates - but IIRC it already depends on smithy, so we should be fine either way.

jakirkham:
> We will reuse these functions in the web service that sends pull requests after pin updates - but IIRC it already depends on smithy...

Yep, I would like to use the nice functionality you have provide here anywhere it can help including the pinning service.

croth1:
@jakirkham shall I move the helper functions to smithy or leave them in the pinning package?

jakirkham:
I think they should move to `conda-smithy`.

croth1:
@jakirkham should be done

croth1:
xref: https://github.com/conda-forge/aubio-feedstock/pull/6

PR #484, opened 2017-03-19
Specify channel for conda-forge-build-setup and obvious-ci installations

shadowwalkersb:
When `conda-forge` is not listed in targets/sources, `conda-forge-build-setup` and `obvious-ci` can't be installed. This PR adds `conda-forge` channel to installation commands explicitly. These tools will always be downloaded from `conda-forge`, regardless of the channel list.

### DIFF
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -48,8 +48,8 @@ install:
     - cmd: conda config --add channels {{ channel }}{% endfor %}
 
     # Configure the VM.
-    - cmd: conda install -n root --quiet --yes obvious-ci -c conda-forge
-    - cmd: conda install -n root --quiet --yes conda-forge-build-setup -c conda-forge
+    - cmd: conda install -n root --quiet --yes -c conda-forge obvious-ci
+    - cmd: conda install -n root --quiet --yes -c conda-forge conda-forge-build-setup
     {% if build_setup -%}
     {{ build_setup }}{% endif %}
 
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -48,8 +48,8 @@ install:
     - cmd: conda config --add channels {{ channel }}{% endfor %}
 
     # Configure the VM.
-    - cmd: conda install -n root --quiet --yes obvious-ci
-    - cmd: conda install -n root --quiet --yes conda-forge-build-setup
+    - cmd: conda install -n root --quiet --yes obvious-ci -c conda-forge
+    - cmd: conda install -n root --quiet --yes conda-forge-build-setup -c conda-forge
     {% if build_setup -%}
     {{ build_setup }}{% endif %}
 
--- a/conda_smithy/templates/run_docker_build.tmpl
+++ b/conda_smithy/templates/run_docker_build.tmpl
@@ -41,7 +41,7 @@ echo "$config" > ~/.condarc
 # A lock sometimes occurs with incomplete builds. The lock file is stored in build_artefacts.
 conda clean --lock
 
-conda install --yes --quiet conda-forge-build-setup -c conda-forge
+conda install --yes --quiet -c conda-forge conda-forge-build-setup
 {% if build_setup -%}
 {{ build_setup }}{% endif -%}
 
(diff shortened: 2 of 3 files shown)


conda_smithy/templates/appveyor.yml.tmpl
```diff
@@ -48,8 +48,8 @@ install:
     - cmd: conda config --add channels {{ channel }}{% endfor %}
 
     # Configure the VM.
-    - cmd: conda install -n root --quiet --yes obvious-ci
-    - cmd: conda install -n root --quiet --yes conda-forge-build-setup
+    - cmd: conda install -n root --quiet --yes obvious-ci -c conda-forge
+    - cmd: conda install -n root --quiet --yes conda-forge-build-setup -c conda-forge
```
jakirkham: Might be better to put the channel before the package being installed so it is easier to read.

shadowwalkersb: Done.

shadowwalkersb:
> This seems fine. That said, why not just add these packages to your own channel?

In general, why duplicate? But, the specific case that prompted this was that I needed a package to be pulled from `defaults` and not from `conda-forge`. I specified only `defaults` in my `conda-forge.yml`, but couldn't get `conda-forge-build-setup`. And, I can't add these packages to `defaults`, right?

jakirkham:
We could ask Continuum about getting `obvious-ci` added. I don't think it will see too many more updates. So they might agree to add it given the low maintenance burden. Also they did add `conda-build-all`. Though we are planning to drop `obvious-ci` from usage here. So I don't know how much longer this will matter.

I don't see `conda-forge-build-setup` being added to `defaults` (nor would I really want it to be). For that matter, I wouldn't encourage anyone to reuse our package for `conda-forge-build-setup`. That package merely exists as a way of making quick fixes to the CIs. So if conda-forge needs a quick fix, we will make it as we see fit. That could have ramifications for outside users, but generally I don't see that blocking us from making those changes. Hence I would recommend making your own feedstock/package for this instead. It is easy for you to pick up our fixes if you agree with them or not, but you won't be forced into decisions you may not agree with.

So I guess ATM my feeling is while this change doesn't hurt so much, it doesn't feel necessary either.

jakirkham:
As to dropping `obvious-ci`, please see PR ( https://github.com/conda-forge/conda-smithy/pull/392 ).

shadowwalkersb:
> I don't see conda-forge-build-setup being added to defaults (nor would I really want it to be). For that matter, I wouldn't encourage anyone to reuse our package for conda-forge-build-setup. That package merely exists as a way of making quick fixes to the CIs.

Exactly. So, why would I add it to my channel, then? It makes more sense to decouple tools (conda-forge-build-setup and associated channels) from the package sources and their channels. According to your suggestion any outside user would have to add conda-forge-build-setup to their channels. I think this PR is improvement to conda-smithy regardless of outside users. If you think it is not, feel free to close it. I would rather fork conda-smithy than fork packages for this purpose.

jakirkham:
>> I don't see conda-forge-build-setup being added to defaults (nor would I really want it to be). For that matter, I wouldn't encourage anyone to reuse our package for conda-forge-build-setup. That package merely exists as a way of making quick fixes to the CIs.

> Exactly. So, why would I add it to my channel, then?

I think you have misunderstood me. To be clear, I'm saying that you should feel free to take that feedstock/package and use it to configure your CIs as you see fit. Otherwise you will be subject to what conda-forge sees fit and you may not agree with it. Or even if you like a change we are considering, you may have to wait until we enact it. These problems simply don't exist if you have your own `conda-forge-build-setup` package and feedstock. For instance, we stayed on `conda-build` 1.x for a long time because it took a while to prepare for that upgrade. In your org, you could have gone ahead with `conda-build` 2.x if you wished if you had this package and feedstock up and running there. However if you stuck to our package you would have been stuck just as long as we were. This is merely one case.

jakirkham:
Am going to close this out.

Have asked to have `obvious-ci` added to `defaults` in issue ( https://github.com/ContinuumIO/anaconda-recipes/issues/95 ). Feel free to follow-up with them. Though, as noted, `obvious-ci` on its way out at conda-forge. ( https://github.com/conda-forge/conda-smithy/pull/392 )

Specifying that `conda-forge-build-setup` only comes from `conda-forge` misses the point that this is an org specific package and each org should specify their own as they see fit.

Beyond these two issues is the problem of re-rendering ~2000 feedstocks with this change. It is not a small ask and doesn't actually make an improvement at conda-forge.

Happy to further discuss alternative solutions.

Issue #487, opened 2017-03-21
Fix curl


jakirkham:
Seems `curl` is not on the path on AppVeyor. This breaks the fast finish script. We need to either add it back to the path or use something else.

xref: https://github.com/appveyor/ci/issues/1426

PR #486, opened 2017-03-21
Setup path for curl on appveyor

SylvainCorlay:
Fixes https://github.com/conda-forge/conda-forge.github.io/issues/357
Fixes https://github.com/conda-forge/conda-smithy/issues/487

The update of git in the latest build image of appveyor broke the windows build. Cf http://help.appveyor.com/discussions/problems/6312-curl-command-not-found

PR #489, opened 2017-03-21
Download fast finish script with PowerShell

jakirkham:
Fixes https://github.com/conda-forge/conda-forge.github.io/issues/357
Fixes https://github.com/conda-forge/conda-smithy/issues/487
Closes https://github.com/conda-forge/conda-smithy/pull/486
Closes https://github.com/conda-forge/setuptools-feedstock/pull/62 (for testing)

It appears `curl` has moved off of the path on AppVeyor. ( https://github.com/appveyor/ci/issues/1426 ) While it is possible to move `curl` back onto the path, there are a couple of concerns. First this may result in including other things on the path that we don't intend to. Second those other things may get involved in builds corrupting packages we are trying to deploy. Third there was never any guarantee that `curl` would be included on Windows nor is there any guarantee it will still be there. So this changes the download command to make use of PowerShell to perform the download. This is something we can rely on. Therefore PowerShell should be a safe choice for long term usage.

Edit: A similar change has been proposed at staged-recipes ( https://github.com/conda-forge/staged-recipes/pull/2648 ).

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -471,7 +471,7 @@ def render_appveyor(jinja_env, forge_config, forge_dir):
         if os.path.exists(cfbs_fpath):
             fast_finish_script += "{recipe_dir}\\ff_ci_pr_build".format(recipe_dir=forge_config["recipe_dir"])
         else:
-            get_fast_finish_script += "curl https://raw.githubusercontent.com/conda-forge/conda-forge-build-setup-feedstock/master/recipe/ff_ci_pr_build.py > ff_ci_pr_build.py"
+            get_fast_finish_script += '''powershell -Command "(New-Object Net.WebClient).DownloadFile('https://raw.githubusercontent.com/conda-forge/conda-forge-build-setup-feedstock/master/recipe/ff_ci_pr_build.py', 'ff_ci_pr_build.py')"'''
             fast_finish_script += "ff_ci_pr_build"
             fast_finish += "del {fast_finish_script}.py"
 


conda_smithy/configure_feedstock.py
```diff
@@ -471,7 +471,7 @@ def render_appveyor(jinja_env, forge_config, forge_dir):
         if os.path.exists(cfbs_fpath):
             fast_finish_script += "{recipe_dir}\\ff_ci_pr_build".format(recipe_dir=forge_config["recipe_dir"])
         else:
-            get_fast_finish_script += "curl https://raw.githubusercontent.com/conda-forge/conda-forge-build-setup-feedstock/master/recipe/ff_ci_pr_build.py > ff_ci_pr_build.py"
+            get_fast_finish_script += '''powershell -Command "(New-Object Net.WebClient).DownloadFile('https://raw.githubusercontent.com/conda-forge/conda-forge-build-setup-feedstock/master/recipe/ff_ci_pr_build.py', 'ff_ci_pr_build.py')"'''
```
ocefpaf: Why not the much cleaner and guaranteed to work `appveyor DownloadFile`. (or better yet, the solution we had before and was working without depending on any external tool.)

jakirkham: > Why not the much cleaner and guaranteed to work `appveyor DownloadFile`.

Short answer people have tried it and they run into intermittent issues. The fix for their case is to use PowerShell like we are doing here. I'm not seeing similar reports with PowerShell.

ref: http://help.appveyor.com/discussions/problems/2753-rawgithubusercontentcom-gives-server-unavailable-sporadically

> ...or better yet, the solution we had before and was working without depending on any external tool.

This line has only ever used `curl`. Not sure what you are think of, but I'm pretty sure it is something else. Though please feel free to clarify.

ocefpaf:
Wow a 2015 issue! Nice googling skills!
btw I always use that and never saw it happen.

jakirkham:
Well, they still use the same [construct]( https://www.appveyor.com/docs/how-to/rdp-to-build-worker/ ) in their docs when working with `raw.githubusercontent.com`. If `appveyor DownloadFile` is preferred and works effectively now with `raw.githubusercontent.com`, why are they still recommending PowerShell?

mwcraig:
TL;DR -- I'd suggest we merge this, as it uses the OSes built-in method for downloading. 

Longer:

Looked a bit at the history of this, and I assume the previous solution to which @ocefpaf is referring is this https://github.com/jakirkham/conda-smithy/commit/ed97b465d9b20d94e6baa914dcda76a97614ffee#diff-a21d424d600d24e538028a242ff2da78L36

While I might have had a mild preference for not  changing anything, since that had been working, I (personally) find that old way to be an impenetrable mess I would never be able to debug 😅 if it failed. Or perhaps, more accurately, I'd never want to get to the point that I understood powershell that well. 😱 

Abstracting the fast-finish concept out into a python script that works across CIs seems like a nice idea because it means other projects could potentially use it instead of copy/pasting PowerShell (as happened with the `%CMD_IN_ENV%` bit of magic).

The question of whether to use `appveyor Download` or a PowerShell-native solution seems relatively inconsequential to me, so let's just pick one and run with it. All else being even, I'd lean towards the Windows-provided solution over the appveyor-provided solution, but both would be reasonable.

nicoddemus:
I agree with @mwcraig, we should adopt a solution ASAP. The built-in method seems good enough IMHO.

jakirkham:
Thanks @mwcraig. Will tag and release.

Issue #233, opened 2016-07-19
Error during registering CI


jakirkham:
Not sure if this is worth looking into further. This happened once. Saw this while trying to enable the Travis CI setting "Build only if .travis.yml is present". Other info can be found in the log along with the error at the end.

```
$ conda smithy --version
No circle token.  Create a token at https://circleci.com/account/api and
put it in ~/.conda-smithy/circle.token
No appveyor token. Create a token at https://ci.appveyor.com/api-token and
Put one in ~/.conda-smithy/appveyor.token
1.0.1
$ conda info
Current conda install:

             platform : osx-64
        conda version : 4.1.8
    conda-env version : 2.5.2
  conda-build version : 1.21.6
       python version : 2.7.12.final.0
     requests version : 2.10.0
     root environment : /zopt/conda2  (writable)
  default environment : /zopt/conda2/envs/nanshenv
     envs directories : /zopt/conda2/envs
        package cache : /zopt/conda2/pkgs
         channel URLs : https://conda.anaconda.org/t/<TOKEN>/conda-forge/osx-64/
                        https://conda.anaconda.org/t/<TOKEN>/conda-forge/noarch/
                        https://conda.anaconda.org/t/<TOKEN>/nanshe/osx-64/
                        https://conda.anaconda.org/t/<TOKEN>/nanshe/noarch/
                        https://repo.continuum.io/pkgs/free/osx-64/
                        https://repo.continuum.io/pkgs/free/noarch/
                        https://repo.continuum.io/pkgs/pro/osx-64/
                        https://repo.continuum.io/pkgs/pro/noarch/
                        https://conda.anaconda.org/t/<TOKEN>/jakirkham/osx-64/
                        https://conda.anaconda.org/t/<TOKEN>/jakirkham/noarch/
                        https://conda.anaconda.org/t/<TOKEN>/javascript/osx-64/
                        https://conda.anaconda.org/t/<TOKEN>/javascript/noarch/
          config file : /Users/kirkhamj/.condarc
         offline mode : False
    is foreign system : False
$ conda smithy register-ci --feedstock_directory=check-feedstock --org  conda-forge
No circle token.  Create a token at https://circleci.com/account/api and
put it in ~/.conda-smithy/circle.token
No appveyor token. Create a token at https://ci.appveyor.com/api-token and
Put one in ~/.conda-smithy/appveyor.token
CI Summary for conda-forge/check-feedstock (can take ~30s):
Traceback (most recent call last):
[...]

PR #498, opened 2017-04-01
Let the travis-ci response fail to be json.

ericdill:
Also print out what the response actually is.

Maybe addresses #233.

Any idea how one would go about testing this?

### DIFF
--- a/conda_smithy/ci_register.py
+++ b/conda_smithy/ci_register.py
@@ -194,8 +194,8 @@ def add_project_to_travis(user, project):
             # https://github.com/conda-forge/conda-smithy/issues/233
             # ValueError: No JSON object could be decoded
             # Maybe trying again in a few seconds will fix this.
-            print('travis-ci says: %s' % response)
-            pass
+            print('travis-ci says: %s' % response.text)
+            raise
         try:
             found = [hooked for hooked in content['hooks']
                      if hooked['owner_name'] == user and hooked['name'] == project]
--- a/conda_smithy/ci_register.py
+++ b/conda_smithy/ci_register.py
@@ -177,7 +177,7 @@ def appveyor_configure(user, project):
 def add_project_to_travis(user, project):
     headers = travis_headers()
     endpoint = 'https://api.travis-ci.org'
-    
+
     url = '{}/hooks'.format(endpoint)
 
     found = False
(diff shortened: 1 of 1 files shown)


ericdill:
> Any idea how one would go about testing this?

Oh maybe what this should do is *not* eat the exception, but just print out whatever `response` is and then let the exception bubble up. That way this is not modifying the current behavior of smithy, but potentially giving us some more information about what is going on since this seems to be occurring again on staged-recipes master

jakirkham:
Thanks for giving this some thought, @ericdill. Yeah the story with automatically configuring CIs with feedstocks has not been a good one (particularly of late). Travis, in particular, has been translucent about helping clarify what is going on so we can fix. Maybe they will be better if we end up paying for services? Though certainly Travis CI is not the only one with this issue. AppVeyor has been flaky at times too, but we do have a better idea of some of the problems in that case.

TBH I'm leaning towards having some auto restart with staged-recipes as described in issue ( https://github.com/conda-forge/staged-recipes/issues/2619 ) at this point. After all that seems to be the source of this flakiness (manual use of `conda-smithy` rarely has issues). We already do this with partial conversion failure. So it shouldn't be difficult to extend to total conversion failure. Since we suppress retries on completed conversion, this shouldn't run us into any trouble. What do you think?

ericdill:
I think an auto-restart makes sense. It is not the most elegant way to fix this issue, but it should definitely be better than the current state of things. Can you go ahead and enable auto restarting and I can iterate on a more granular fix as time and interest permits?

3 or 5 retries seems like a sensible starting point.

jakirkham:
So this will take a little coding on my part. It's not bad, but it's not flipping a switch either. Just letting you know.

FWIW I'm suspicious that Travis CI is doing something really silly when one pushes the sync button (as we need to do here). So I'm not sure if there is an elegant solution on our end, but there might be on there's. ( https://github.com/travis-ci/travis-ci/issues/6320 ) This also raises some questions about long term solutions as well.

ericdill:
> So this will take a little coding on my part. It's not bad, but it's not flipping a switch either. Just letting you know.

Definitely. Thanks for clarifying. I misunderstood what you were proposing. I literally thought there was an "auto restart" feature on the travis-ci configs somewhere. 🤦‍♂️

ericdill:
I'm curious what the text of the response is. Do you think we could get this in since it doesn't change the behavior of conda-smithy?

ericdill:
> Was more trying to share some thoughts on this problem generally

For sure. Let's keep discussing this issue

> we might want to check it doesn't print out the token

yeesh, good point. I sure hope travis doesn't do this...

minrk:
I noticed that error codes aren't being checked (#503) and there's a pretty good chance that Travis is responding with an error here. The text is just 'access denied' for an incorrect token, for instance.

My *guess* is that Travis may have flagged the access token for some reason and blocked it. This ought to show up as a 403 error. If this is the case, issuing a new token should solve the issue.

It could also be Travis failing to deal with the size of the response for some reason, in which case there would hopefully be a 5XX error. Not as much we could do about that, though.

Potentially useful: All responses from Travis include a `X-Request-Id` header containing a UUID identifying the request, presumably for their own debugging purposes. If this is a regular thing, logging the request ID for failed requests may be useful for getting support from Travis.

Issue #494, opened 2017-03-25
KeyError: 'python-3.3.0-pro0.tar.bz2'


astrofrog:
I tried to rerender the [glue-core-feedstock](https://github.com/conda-forge/glue-core-feedstock) feedstock with conda-smithy 2.2.2 and ran into the following error:

```
$ conda smithy rerender
No circle token.  Create a token at https://circleci.com/account/api and
put it in ~/.conda-smithy/circle.token
No appveyor token. Create a token at https://ci.appveyor.com/api-token and
Put one in ~/.conda-smithy/appveyor.token
No anaconda token. Create a token via
  anaconda auth --create --name conda-smithy --scopes "repos conda api"
and put it in ~/.conda-smithy/anaconda.token
Fetching package metadata .............
Traceback (most recent call last):
  File "/Users/tom/miniconda3/bin/conda-smithy", line 6, in <module>
    sys.exit(conda_smithy.cli.main())
  File "/Users/tom/miniconda3/lib/python3.5/site-packages/conda_smithy/cli.py", line 275, in main
    args.subcommand_func(args)
  File "/Users/tom/miniconda3/lib/python3.5/site-packages/conda_smithy/cli.py", line 186, in __call__
    configure_feedstock.main(args.feedstock_directory)
  File "/Users/tom/miniconda3/lib/python3.5/site-packages/conda_smithy/configure_feedstock.py", line 709, in main
    render_circle(env, config, forge_dir)
  File "/Users/tom/miniconda3/lib/python3.5/site-packages/conda_smithy/configure_feedstock.py", line 77, in render_circle
    forge_config.get('channels', {}).get('sources', tuple())
  File "/Users/tom/miniconda3/lib/python3.5/site-packages/conda_smithy/configure_feedstock.py", line 628, in compute_build_matrix
    mtx = special_case_version_matrix(meta, index)
  File "/Users/tom/miniconda3/lib/python3.5/site-packages/conda_build_all/version_matrix.py", line 223, in special_case_version_matrix
    py_vn = minor_vn(index[python_pkg.fn]['version'])
KeyError: 'python-3.3.0-pro0.tar.bz2'
```

jakirkham:
I'm not able to reproduce this error, @astrofrog. Do you get this error consistently or is it sporadic? Could you please also provide some more information like `conda list -n root`? Also what OS are you using? Have you change `conda-forge.yml`? What does `.condarc` contain? Thanks.

[...]

PR #502, opened 2017-04-06
check for conda < 4.3

jcb91:
as discussed in https://github.com/conda-forge/conda-smithy/issues/494
check for conda < 4.3, and if not found, print a message more informative than the exception which would otherwise arise (since currently we're incompatible with 4.3+)

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -8,6 +8,7 @@ import time
 import argparse
 
 import conda
+from distutils.version import LooseVersion
 from conda_build.metadata import MetaData
 
 from . import ci_register
@@ -274,8 +275,7 @@ def main():
         args = parser.parse_args()
 
     # Check conda version for compatibility
-    conda_version_tuple = tuple(map(int, conda.__version__.split('.')[:3]))
-    if conda_version_tuple > (4, 2):
+    if LooseVersion(conda.__version__) < LooseVersion('4.3'):
         print('You appear to be using conda {}, but conda-smithy {} is\ncurrently only compatible with conda versions < 4.3.'.format(
             conda.__version__, __version__))
         sys.exit(2)
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -7,6 +7,7 @@ import sys
 import time
 import argparse
 
+import conda
 from conda_build.metadata import MetaData
 
 from . import ci_register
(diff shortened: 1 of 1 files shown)


conda_smithy/cli.py
```diff
@@ -272,6 +273,13 @@ class Rerender(Regenerate):
     else:
         args = parser.parse_args()
 
+    # Check conda version for compatibility
+    conda_version_tuple = tuple(map(int, conda.__version__.split('.')[:3]))
+    if conda_version_tuple > (4, 2):
```
jakirkham: `distutils` has some version comparison stuff that might be helpful. Would check that out.

jcb91:
sure, could do something like

```python
from distutils.version import LooseVersion, StrictVersion
if LooseVersion(conda.__version__) < LooseVersion('4.3')
```

?

jakirkham:
Yep, something like that could work.

jcb91:
Ok, done. However, as noted in [the stackoverflow post](http://stackoverflow.com/questions/11887762) where I found that distutils comparison, 

> `distutils.version` is built in but __is undocumented and conformant only to the superseded PEP 386__

(emphasis mine)

jakirkham:
Mainly suggested it because the previous check was failing and this would be an easy fix.

While it is true that they are undocumented and rely on an old superseded PEP, there is a lot of usage of both [`LooseVersion`]( https://github.com/search?utf8=%E2%9C%93&q=distutils+LooseVersion&type=Code ) and [`StrictVersion`]( https://github.com/search?utf8=%E2%9C%93&q=distutils+StrictVersion&type=Code ). In fact, I think the main reason they haven't been fixed up to match PEP440 and documented is because there is fear of breaking existing people's code. I don't think `distutils.version` is going to disappear quickly and quietly. Maybe by Python 4 it will be gone. Hopefully by that time, `conda` does something smarter and we don't need this warning at all. If not, as you have noted, it is easy enough to fix when the time comes.

TL;DR Noted the concern, but am not too worried about it.

jcb91:
Yep, fair enough, I think it's probably sufficient...

conda_smithy/cli.py
```diff
@@ -272,6 +274,12 @@ class Rerender(Regenerate):
     else:
         args = parser.parse_args()
 
+    # Check conda version for compatibility
+    if LooseVersion(conda.__version__) >= LooseVersion('4.3'):
+        print('You appear to be using conda {}, but conda-smithy {} is\ncurrently only compatible with conda versions < 4.3.'.format(
+            conda.__version__, __version__))
+        sys.exit(2)
```
jakirkham: Why 2?

jcb91:
Well, I saw 1 being used already (I think for linting errors?), so thought it might be good to distinguish this failure case. But of course, it needn't be 2, if you'd rather keep it simpler!

jakirkham:
Sure that's fine. Just wanted to know if there was some reason.

jakirkham:
Thanks @jcb91.

PR #503, opened 2017-04-10
check http response codes in more places

minrk:
check for error codes with `response.raise_for_status()` before assuming responses are valid JSON, etc.

related to #498 where there's a pretty good chance that travis is responding with an error.

### DIFF
--- a/conda_smithy/ci_register.py
+++ b/conda_smithy/ci_register.py
@@ -188,8 +188,9 @@ def add_project_to_travis(user, project):
 
         response = requests.get(url, headers=headers)
         try:
+            response.raise_for_status()
             content = response.json()
-        except ValueError:
+        except (requests.HTTPError, ValueError):
             # We regularly seem to hit this issue during automated feedstock registration on github.
             # https://github.com/conda-forge/conda-smithy/issues/233
             # ValueError: No JSON object could be decoded
@@ -207,6 +208,7 @@ def add_project_to_travis(user, project):
                 print(" * Travis doesn't know about the repo, synching (takes a few seconds).")
                 synch_url = '{}/users/sync'.format(endpoint)
                 response = requests.post(synch_url, headers=headers)
+                response.raise_for_status()
             time.sleep(3)
 
         if count > 20:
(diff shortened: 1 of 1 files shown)


conda_smithy/ci_register.py
```diff
@@ -188,8 +188,9 @@ def add_project_to_travis(user, project):
 
         response = requests.get(url, headers=headers)
         try:
+            response.raise_for_status()
             content = response.json()
-        except ValueError:
+        except Exception:
```
jakirkham: Couldn't we do better than a generic exception here? Looks like this should be an [`HTTPError`]( http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status ) for the status. I don't know if the `ValueError` from trying to parse the JSON still needs to be handled, but I guess we could do that too if needed.

jakirkham:
Thanks @minrk. Barring one minor comment, LGTM.

conda_smithy/ci_register.py
minrk: Good call, done.

jakirkham:
Mind taking a quick look, @ericdill?

conda_smithy/ci_register.py
```diff
@@ -188,8 +188,9 @@ def add_project_to_travis(user, project):
 
         response = requests.get(url, headers=headers)
         try:
+            response.raise_for_status()
             content = response.json()
-        except ValueError:
+        except (requests.HTTPError, ValueError):
```
ericdill: Ah, yes of course. That's a critical addition here

jakirkham:
Is there a way for us to tell Travis CI to send the response in chunks? Maybe trying to send a big glob of data is causing the failure.

minrk:
> Is there a way for us to tell Travis CI to send the response in chunks?

I don't believe so. GitHub's API has a paged-result structure for all list results, but the Travis API only appears to have a similar limit for logs and requests (basically the endpoints they expect to have a lot of results). Looks like #505 circumvents the problem, anyway.

Issue #527, opened 2017-06-16
Upgrade to CircleCI 2.0


jakirkham:
CircleCI has introduced breaking changes in [CircleCI 2.0]( https://circleci.com/docs/2.0/migrating-from-1-2/ ). This will allow for real matrix builds on CircleCI, which should help some of our long running builds that are timing out.

jakirkham:
This upgrade to the template is available in PR ( https://github.com/conda-forge/conda-smithy/pull/530 ) and ready to go.

PR #530, opened 2017-06-17
Support CircleCI 2.0

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/527

Closes https://github.com/conda-forge/pyelftools-feedstock/pull/14
Closes https://github.com/conda-forge/gnureadline-feedstock/pull/8

Upgrades to new [CircleCI 2.0]( https://circleci.com/docs/2.0/ ) by borrowing from this [migration guide]( https://circleci.com/docs/2.0/migrating-from-1-2/ ), the [reference spec]( https://circleci.com/docs/2.0/configuration-reference/ ), and a [bunch of examples]( https://github.com/CircleCI-Public ). This should make it easier to explore matrix builds in CircleCI and provide us access to all the other latest features.

Test cases for the [Linux build]( https://github.com/conda-forge/pyelftools-feedstock/pull/14 ) and [non-Linux build]( https://github.com/conda-forge/gnureadline-feedstock/pull/8 ) can be seen. Note that CircleCI still doesn't build in the non-Linux case (and if it does it just exits quietly). Also despite looking like a big overhaul, the behavior in the CircleCI 2.0 case is basically the same. Though this will change when matrix builds are considered.

In the upgrade, we make sure to clean out the old `circle.yml` file too.

### DIFF
--- a/conda_smithy/templates/circle.yml.tmpl
+++ b/conda_smithy/templates/circle.yml.tmpl
@@ -1,3 +1,5 @@
+version: 2
+
 {%- block env_branches -%}
 {%- if not matrix -%}
 general:
--- a/conda_smithy/templates/circle.yml.tmpl
+++ b/conda_smithy/templates/circle.yml.tmpl
@@ -1,11 +1,11 @@
 version: 2
 
 jobs:
+{%- block env_test -%}
+{%- if not matrix %}
   build:
     working_directory: ~/test
     machine: true
-{%- block env_test -%}
-{%- if not matrix %}
     branches:
       ignore:
         - /.*/
@@ -14,6 +14,9 @@ jobs:
           # The Circle-CI build should not be active, but if this is not true for some reason, do a fast finish.
           command: exit 0
 {%- else %}
+  build:
+    working_directory: ~/test
+    machine: true
     steps:
       - checkout
       - run:
--- a/conda_smithy/templates/circle.yml.tmpl
+++ b/conda_smithy/templates/circle.yml.tmpl
@@ -13,7 +13,7 @@ jobs:
       - run:
           # The Circle-CI build should not be active, but if this is not true for some reason, do a fast finish.
           command: exit 0
-{% else %}
+{%- else %}
     steps:
       - checkout
       - run:
@@ -26,6 +26,22 @@ jobs:
       - run:
           # Run, test and (if we have a BINSTAR_TOKEN) upload the distributions.
           command: ./ci_support/run_docker_build.sh
-{# #}
 {%- endif -%}
 {%- endblock %}
+
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+{%- block env_deps -%}
+{%- if not matrix %}
+      - build:
+          filters:
+            branches:
+              ignore:
+                - /.*/
+{%- else %}
+      - build
+{%- endif -%}
+{%- endblock %}
+{# #}
--- a/conda_smithy/templates/circle.yml.tmpl
+++ b/conda_smithy/templates/circle.yml.tmpl
@@ -5,7 +5,15 @@ jobs:
     working_directory: ~/test
     machine: true
 {%- block env_test -%}
-{%- if matrix %}
+{%- if not matrix %}
+    branches:
+      ignore:
+        - /.*/
+    steps:
+      - run:
+          # The Circle-CI build should not be active, but if this is not true for some reason, do a fast finish.
+          command: exit 0
+{% else %}
     steps:
       - checkout
       - run:
@@ -18,14 +26,6 @@ jobs:
       - run:
           # Run, test and (if we have a BINSTAR_TOKEN) upload the distributions.
           command: ./ci_support/run_docker_build.sh
-{% else %}
-    branches:
-      ignore:
-        - /.*/
-    steps:
-      - run:
-          # The Circle-CI build should not be active, but if this is not true for some reason, do a fast finish.
-          command: exit 0
 {# #}
 {%- endif -%}
 {%- endblock %}
--- a/conda_smithy/templates/circle.yml.tmpl
+++ b/conda_smithy/templates/circle.yml.tmpl
@@ -1,41 +1,35 @@
 version: 2
 
+jobs:
+  build:
+    working_directory: ~/test
+    machine: true
 {%- block env_branches -%}
-{%- if not matrix -%}
-general:
-  branches:
-    ignore:
-      - /.*/
-
-{% endif -%}
-{% endblock -%}
-{% block env_dependencies -%}
-{%- if matrix -%}
-checkout:
-  post:
-    - ./ci_support/fast_finish_ci_pr_build.sh
-    - ./ci_support/checkout_merge_commit.sh
-
-machine:
-  services:
-    - docker
-
-dependencies:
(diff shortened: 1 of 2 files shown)


conda_smithy/templates/circle.yml.tmpl
```diff
@@ -1,39 +1,31 @@
...
+      - run:
+          name: Fast finish outdated PRs and merge PRs
+          command: |
+            ./ci_support/fast_finish_ci_pr_build.sh
+            ./ci_support/checkout_merge_commit.sh
+      - run:
+          command: docker pull condaforge/linux-anvil
+      - run:
+          # Run, test and (if we have a BINSTAR_TOKEN) upload the distributions.
+          command: ./ci_support/run_docker_build.sh
```
gqmelo: It would be nice to check how the cache works in this new version. So we could place this command in a section which could be cached and need fewer changes to use ccache.

gqmelo: The cache is much more flexible and or seems we choose exactly when to save and restore caches: https://circleci.com/docs/2.0/caching/

jakirkham: We should resurrect the package caching work you did in this context. There are many pure Python or metapackages now floating around that have huge lists of dependencies that slow down the test phase due to downloading. Having a package cache there will certainly help the CIs cope with our ever increasing number of feedstocks.

gqmelo: Sure, we should definitely revisit this. I can take a look after you merge this.

jakirkham:
So a couple things to consider that I could use feedback on.

First we could move the `ci_support` scripts into `.circleci`. All of these scripts are CircleCI specific and it seems in some of CircleCI's examples they do the same thing. This would keep things pretty nicely organized. That said, it would move the `run_docker_build.sh` script, which has lived in this directory for a long time. Not sure if that would impact offline build usage. Also as the `.circleci` directory would appear hidden on Unix platforms, it may make this script less visible and could cause some confusion. For now I have done nothing on this front. Thoughts on this would definitely be appreciated.

Second technically CircleCI 2.0 is beta. That said, the closed beta started on [Nov 7]( https://circleci.com/blog/say-hello-to-circleci-2-0/ ) (last year). Also it looks like they opened the beta up on [Mar 17]( https://circleci.com/blog/circleci-2-0-beta-is-ready-for-your-builds/ ). Assuming a similar time range for ending beta, I would expect this to be official by sometime in July (i.e. next month). Presumably some time after that CircleCI 1.0 is going to be phased out. Given the number of users who have already tried out the beta and time elapsed (not to mention our slow rate of adoption for things like this), I don't think this is risky and it is something we are going to need to do anyways. Combined with better features for caching and matrix builds, this seems very desirable to adopt soon. That said, it would still be good to get feedback from others on this.

Given both of these changes can be seen as breaking, I'd be ok marking this as `conda-smithy` 3.0.0. As we need to start phasing out NumPy 1.11 as well to lighten the CIs (and in some cases already are), making this a breaking release seems reasonable. If it seems necessary, we could discuss some maintenance support for 2.x, but the NumPy issue would need to be taken into account as well.

jakirkham:
Would appreciate if @conda-forge/core could please take a look at this. Also feedback on the items noted above would be very helpful. Thanks in advance.

jakirkham:
Also a CircleCI matrix implementation based off of this can be seen in PR ( https://github.com/conda-forge/conda-smithy/pull/532 ) with links to example re-renderings.

gqmelo:
> That said, it would move the run_docker_build.sh script, which has lived in this directory for a long time.

I would prefer to not hide run_docker_build.sh under ```.circleci```. Even though it's only used on CircleCI it's generic enough to be used standalone (personally I always run it to reproduce linux builds)

> Second technically CircleCI 2.0 is beta.

Given the amount of time since it entered beta it seems pretty safe to make those changes now.

jakirkham:
Yeah, I think I agree on both points.

Though the `run_docker_build.sh` script will need to change for CircleCI matrix builds. ( https://github.com/conda-forge/conda-smithy/pull/532 ) Still can use it externally, but one will want to set `CONDA_*` variables before running it.

Any thoughts on the other scripts in `ci_support`? Is it worth moving them over to `.circleci`? Could declutter things a bit.

jakirkham:
As @gqmelo noted elsewhere, it does seem that CircleCI 2.0 is not correctly grouping pull requests. Have opened a discussion issue on this point. I remember having similar issues on CircleCI 1.0 before when they made changes to pull requests, but it didn't affect anything beyond how things were grouped. Since the build number always increases and is unique to the repo build not how it was created, this issue ends up only being a UI one. Nevertheless it is worth noting.

ref: https://discuss.circleci.com/t/circleci-2-0-incorrectly-groups-pull/13505

gqmelo:
> Though the run_docker_build.sh script will need to change for CircleCI matrix builds. ( #532 ) Still can use it externally, but one will want to set CONDA_* variables before running it.

Yeah, no problem. This is already necessary when trying to reproduce osx and windows builds. It would be better to have a script for each platform to build all packages locally (like the current run_docker_build.sh), but I know we would probably have to duplicate code as the matrices are on CI config files.

> Any thoughts on the other scripts in ci_support? Is it worth moving them over to .circleci? Could declutter things a bit.

Well, the other seems very specific to CircleCI and something one would rarely want to run locally. So I would be fine with moving them.

jakirkham:
> It would be better to have a script for each platform to build all packages locally (like the current run_docker_build.sh), but I know we would probably have to duplicate code as the matrices are on CI config files.

Not to digress, but to mention this brief. Have been wanting to do something like this for a while. If we go for setting `CONDA_*` variables, it shouldn't be too bad. The main trick is we don't want to mess up the user's `conda` install. Also we probably don't want to add stuff to their home directory. So we might want to download `conda` to a temporary directory and go from there.

gqmelo:
> The main trick is we don't want to mess up the user's conda install. Also we probably don't want to add stuff to their home directory. So we might want to download conda to a temporary directory and go from there.

Yeah, if the scripts are going to install packages on the root env it would be better to have a separate miniconda installation. But we could have separate scripts for configuring miniconda and installing needed packages and other scripts to build all packages for the given plataform.

Something like:
```
  feedstock-dir
    |_ build_scripts
        |_ build_linux.sh
        |_ build_osx.sh
        |_ build_windows.bat
        |_ setup_linux.sh
        |_ setup_osx.sh
        |_ setup_windows.bat
```
```setup_*``` scripts would change the user installation, but at least it will be more explicit and ```build_*``` would build all the matrix for the given platform.
So someone trying to reproduce windows builds would just have to run ```setup_windows.bat``` and then ```build_windows.bat``` to build all packages without having to worry about the matrix and the appveyor.yml file.
Anyway, this is just an idea for future refactorings.

jakirkham:
Friendly nudge @conda-forge/core. 😉

jakirkham:
Should add that CircleCI 2.0 is [officially released]( https://circleci.com/blog/launching-today-circleci-2-0-reaches-general-availability/ ). So no longer in beta.

jakirkham:
Given CircleCI 2.0 is no longer in beta, this is well tested, and has sat for a while to get feedback, going to go ahead and merge it to clear it from the list.

mwcraig:
Thanks!

PR #532, opened 2017-06-18
CircleCI 2.0 Matrix

jakirkham:
Fixes https://github.com/conda-forge/conda-smithy/issues/203

Closes https://github.com/conda-forge/patchelf-feedstock/pull/12
Closes https://github.com/conda-forge/pyelftools-feedstock/pull/15
Closes https://github.com/conda-forge/rank_filter-feedstock/pull/27

~~Requires PR ( https://github.com/conda-forge/conda-smithy/pull/530 ) be merged beforehand.~~ Already done.

Creates a proper build matrix on CircleCI 2.0 using the new [Workflows]( https://circleci.com/docs/2.0/workflows/ ). This allows us to have separate builds for Python and NumPy for instance. Though really this is not specific to those two. Instead applies to any environment variables that might be looped over and respected by `conda-build`. This is no-op when Linux builds are skipped. Also it is nearly a no-op for Linux builds that don't require a matrix as it makes some minor changes (whitespace and comments).

Some tests of this functionality can be seen in for the following cases:

1. No matrix build: https://github.com/conda-forge/patchelf-feedstock/pull/12
2. Python only matrix build: https://github.com/conda-forge/pyelftools-feedstock/pull/15
3. Python and NumPy matrix build: https://github.com/conda-forge/rank_filter-feedstock/pull/27

Would note that CircleCI seems to like to add one status per build in the workflow. There is no difference if there is no matrix. However for any case with a non-trivial number of matrix cases, has an equal number of statuses for the PR. This isn't too bad for just Python (3 cases). However for Python and NumPy this can be quite a bit depending on what version are available. There doesn't appear to be any obvious way to change this.


patricksnape:
This would be very useful for Numpy builds +1 from me

jakirkham:
Any comments, @conda-forge/core?

PR #535, opened 2017-06-20
Fix feedstocking with binary files

ocefpaf:
This should help with the binary file copy issue and any other file that fails the `utf-8` encoding.

### DIFF
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -75,10 +75,14 @@ def remove_file(filename):
 
 
 def copy_file(src, dst):
-    with io.open(src, "r", encoding="utf-8") as fh_src:
-        with io.open(dst, "w", encoding="utf-8", newline="\n") as fh_dst:
-            for line in fh_src:
-                fh_dst.write(line)
+    try:
+        with io.open(src, "r", encoding="utf-8") as fh_src:
+            with io.open(dst, "w", encoding="utf-8", newline="\n") as fh_dst:
+                for line in fh_src:
+                    fh_dst.write(line)
+    except UnicodeDecodeError:
+        # Leave any other files alone.
+        shutil.copy(src, dst)
 
     shutil.copymode(src, dst)
 
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -75,6 +75,12 @@ def remove_file(filename):
 
 
 def copy_file(src, dst):
+    """
+    Tried to copy utf-8 text files line-by-line to avoid
+    getting CRLF characters added on Windows.
+
+    If the file fails to be decoed with utf-8 we revert to a regular copy.
+    """
     try:
         with io.open(src, "r", encoding="utf-8") as fh_src:
             with io.open(dst, "w", encoding="utf-8", newline="\n") as fh_dst:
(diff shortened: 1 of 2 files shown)


conda_smithy/feedstock_io.py
```diff
@@ -75,10 +75,14 @@ def remove_file(filename):
...
-            for line in fh_src:
-                fh_dst.write(line)
+    try:
+        with io.open(src, "r", encoding="utf-8") as fh_src:
+            with io.open(dst, "w", encoding="utf-8", newline="\n") as fh_dst:
+                for line in fh_src:
+                    fh_dst.write(line)
+    except UnicodeDecodeError:
+        # Leave any other files alone.
+        shutil.copy(src, dst)
```
ocefpaf: If the line issue exist with text files that fails when opened with `utf-8` that will still exist in this scenario.
But it is better than the current situation and such a scenario would be quite improbable.

conda_smithy/feedstock_io.py
```diff
@@ -75,10 +75,20 @@ def remove_file(filename):
 
 
 def copy_file(src, dst):
-    with io.open(src, "r", encoding="utf-8") as fh_src:
-        with io.open(dst, "w", encoding="utf-8", newline="\n") as fh_dst:
-            for line in fh_src:
-                fh_dst.write(line)
+    """
+    Tried to copy utf-8 text files line-by-line to avoid
+    getting CRLF characters added on Windows.
```
ocefpaf: @nicoddemus can you take a look if this is OK?

jakirkham:
This seems like the simplest most reasonable fix. Totally happy with it. As you have already asked, would be good to have a Windows user try it out.

conda_smithy/feedstock_io.py
nicoddemus: Tried the development version from this branch on a local feedstock and it seemed to work fine: only a `.sh` file was changed, and no CRLF changes at all. 👍

jakirkham: How was the `.sh` file changed?

nicoddemus:
```diff
diff --git a/ci_support/run_docker_build.sh b/ci_support/run_docker_build.sh
index ef556d8..2c6ce73 100755
--- a/ci_support/run_docker_build.sh
+++ b/ci_support/run_docker_build.sh
@@ -58,21 +58,15 @@ conda install --yes --quiet conda-forge-build-setup
 source run_conda_forge_build_setup

 # Embarking on 3 case(s).
-    set -x
     export CONDA_PY=27
-    set +x
     conda build /recipe_root --quiet || exit 1
     upload_or_check_non_existence /recipe_root conda-forge --channel=main || exit 1

-    set -x
     export CONDA_PY=35
-    set +x
     conda build /recipe_root --quiet || exit 1
     upload_or_check_non_existence /recipe_root conda-forge --channel=main || exit 1

-    set -x
     export CONDA_PY=36
-    set +x
     conda build /recipe_root --quiet || exit 1
     upload_or_check_non_existence /recipe_root conda-forge --channel=main || exit 1
 touch /feedstock_root/build_artefacts/conda-forge-build-done
```

Seems legit to me, should have posted it before anyways.

jakirkham: Yeah, that's a change in `master` that is not yet released. So agreed it looks fine.

Thanks for testing @nicoddemus. 😄

PR #539, opened 2017-06-27
Added brew_requirements.txt support

Korijn:
As became apparent on https://github.com/conda-forge/wxpython-feedstock/pull/3 it would be useful to have `brew_requirements.txt` support similar to the `yum_requirements.txt` feature.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -302,11 +302,35 @@ def render_travis(jinja_env, forge_config, forge_dir):
             build_setup += textwrap.dedent("""\
                 # Overriding global conda-forge-build-setup with local copy.
                 source {recipe_dir}/run_conda_forge_build_setup_osx
+
             """.format(recipe_dir=forge_config["recipe_dir"]))
         else:
             build_setup += textwrap.dedent("""\
                 source run_conda_forge_build_setup
+
             """)
+
+        # If there is a "brew_requirements.txt" file in the recipe, we honour it.
+        brew_requirements_fpath = os.path.join(forge_dir, 'recipe',
+                                              'brew_requirements.txt')
+        if os.path.exists(brew_requirements_fpath):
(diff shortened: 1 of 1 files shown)


jakirkham:
The first big question to ask is was this a one off thing or is this clearly becoming a prevalent issue? Right now it looks like a one off thing, but I may be missing other relevant cases. The second one is do we want to become reliant on other package managers? Admittedly this is already the case with `yum`, but `brew` is a bit different. The latter not being a system package manager, but another community packaging project like our own. Third, if we do want to become reliant on other non-system package managers, which ones, why, and what can we use from them?

Depending on the outcome of those questions, there are also some technical issues we will need to discuss for this to work.

IMHO `yum_requirements.txt` is a hack that should be minimized. I say that having been the one who proposed it. 😄 The reason we added `yum_requirements.txt` is people were commonly hacking the CI scripts to install things from `yum` that would be difficult to package. Further we had trouble coming to consensus about what could be installed on the Docker image to start with.

I have yet to see a similar level of issues with macOS as we did with Linux in the beginning. Certainly XQuartz is one such case. That said, there is interest in having a healthy cross platform XOrg stack at conda-forge. As XQuartz would be needed for XOrg on macOS, we have been investigating packaging XQuartz and a considerable amount of work has already be done ( https://github.com/conda-forge/staged-recipes/issues/2587 ). So if XQuartz is the only motivating factor for this change, maybe we should dedicate more resources to finishing the XQuartz packaging project up.

Korijn:
I will be the first to admit that conda packages should rely on conda packages only rather than system packages. I evangelize the same message to my coworkers.

However, sometimes I just need to get a job done. I need that package, and not just for the sake of packaging.

Personally I would open an issue on the feedstock and replace the brew requirement with a conda dependency once XQuartz becomed available on Anaconda.

Heck, you could even include that logic in conda-smithy's rerendering?

Korijn:
In the end we got away with leaving this out of the build.

PR #551, opened 2017-08-09
Honor docker.image in .circle/config.yaml

sdvillal:
Let's not hardcode the docker image we pull in CircleCI.


jakirkham:
Thanks @sdvillal.

It appears that has been hard coded for a long time surprisingly. Definitely time to make use of the configuration parameters from `conda-forge.yml`. 😄

PR #563, opened 2017-09-12
add docs link

ocefpaf:
Our docs have a low visibility and this is a small step to try to change that.

@isuruf can you take a quick look?

Ping @choldgraf who brought this up.


choldgraf:
small comment: I try to avoid hyperlinking only a single word, because some people can't tell if it's hyperlinked when it's part of a sentence. I'd recommend something more like `please see the [conda-forge documentation](<link>)`

my 2 cents :-)

Issue #578, opened 2017-09-20
Error if noarch and selectors are used


jakirkham:
The combination of `noarch` and selectors are at this time incompatible with how `conda` and `conda-build` work. As such, if someone tries to use both, we allow them too, but may build a broken package. It would be better for the linter to error out in this case to stop a potentially broken build from being created.

Came up in this [thread]( https://github.com/conda-forge/conda-forge.github.io/issues/443#issuecomment-330892368 ).

jakirkham:
Proposal in PR ( https://github.com/conda-forge/conda-smithy/pull/602 ).

PR #602, opened 2017-10-19
add test that noarch packages don't have selectors

dougalsutherland:
Fixes https://github.com/conda-forge/conda-smithy/issues/578.

Since this has been happening more often now that the example recipe includes a `noarch: python` instruction, add a linter check that `noarch` packages don't have any selectors in them.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -196,6 +196,14 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
                              'subsection name. {} is not a valid subsection'
                              ' name.'.format(section, subsection))
 
+    # 17: noarch doesn't work with selectors
+    if build_section.get('noarch') is not None:
+        with io.open(meta_fname, 'rt') as fh:
+            if any(True for l in selector_lines(fh)):
+                lints.append("Selectors don't work for `noarch` packages. If "
+                             "the selectors are necessary, please remove "
+                             "`noarch: {}`.".format(build_section['noarch']))
+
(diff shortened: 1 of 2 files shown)


jakirkham:
Thanks @dougalsutherland. Would update the title to something like "Noarch packages can't have selectors".

jakirkham:
Tried to add a fixes note in the OP, but GitHub went bonkers. Tried to cancel out, but I'm not sure if messed something up. Sorry about that. 😞

dougalsutherland:
Maybe my fault, I did a force push earlier. Nothing seems broken now but I'm also not seeing any comments; I'll change the error string, but any other recommended changes?

dougalsutherland:
Oh never mind, it's just the "Fixes" bit which is there now. Will edit the string when I'm out of the talk I'm sitting in now.

Note that tests actually passed, it's just the dumb broken version of conda-build failing as always.

jakirkham:
That's probably safest. I think I'm done trying to editing things for the day.

Cool. Yeah we aren't worried about that old `conda` version. Honestly I think we should just drop it, but that is another discussion.

dougalsutherland:
This just came up in gitter again, so...ping. :)

ocefpaf:
LGTM, the failures seems to be unrelated to the changes here.

Thanks @dougalsutherland!

jakirkham:
Thanks @dougalsutherland and @ocefpaf. Sorry I️ missed the updates @dougalsutherland. Happy to have it in.

PR #609, opened 2017-11-11
[appveyor] update to the new VS2017 image for py3

xoviat:
/cc @isuruf 

This updates the build image to VS2017 for Python 3. VS2017 is ABI compatible with VS2015, and the new image has VS2015 as well, so there shouldn't be any problems. 

That said, I am sure that there is something that I haven't considered.


xoviat:
Actually I'm going to close this because it looks like clang is going to be used. That's OK with me.

PR #636, opened 2018-01-06
Update date range

jschueller:


jakirkham:
Thanks @jschueller. We might want to turn this into a Jinja template that autofills the current year. In any event, this is good to go.

Issue #637, opened 2018-01-07
Detect SELinux and mount volumes appropriately


bollwyvl:
On a recent Fedora host with SELinux, I can't access the mounted volumes for the feedstock and recipe when running `run_docker_build`.

Looks like recent dockers have [fixed this](http://www.projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/).

Because SELinux is a Good Thing, I don't feel like we should punish conscientious developers for running with it.

Haven't gone down the whole rabbit hole, but I think the change would be something like this:
```bash
...
VOLUME_OPTIONS="rw"
if hash getenforce 2> /dev/null && getenforce | grep 'Enforcing' > /dev/null
then
  VOLUME_OPTIONS="${VOLUME_OPTIONS},Z"
fi

cat << EOF | {{ docker.executable }} run -i \
                        -v "${RECIPE_ROOT}":/home/conda/recipe_root:$VOLUME_OPTIONS \
                        -v "${FEEDSTOCK_ROOT}":/home/conda/feedstock_root:$VOLUME_OPTIONS \
                        {{ matrix_env_vars(matrix) }}
                        -a stdin -a stdout -a stderr \
                        {{ docker.image }} \
                        {{ docker.command }} || exit 1
...
```

Barring that, just letting me do `VOLUME_OPTIONS=Z ./scripts/run_docker_build.sh` would do the trick. Happy to PR, with whichever makes the most sense.

bollwyvl:
Indeed, why even mount anything other than `build_artefacts` writeable?

jakirkham:
First I'd like to better understand what you are proposing before we make a change.

Could you help me understand why you want to access the feedstock while the build is running? Not saying there isn't a reason. Just trying to understand your use case is all. :)

Well we actually don't mount `build_artifacts` (recently renamed) as it doesn't necessarily exist. We do mount the feedstock as writable, which is where `build_artifacts` is created. Though we could create `build_artifacts` first and then mount it as writeable. The recipe directory probably can be read only.

Another relevant question is why do we need to mount the recipe and the feedstock directory separately? I don't have a good answer there and we can probably just mount one of them.

What is the `if` statement doing above? Do we need it?

bollwyvl:
My use case is, "locally test the docker build script on an
selinux-enforcing host system". Once inside the container, if the volumes
are not mounted with :z or :Z the recipe can't be found.

[...]

PR #640, opened 2018-01-08
[wip] detect selinux and mount with Z, readonly feedstock

bollwyvl:
This PR:
- mounts the `feedstock_root` as read-only
  - while not strictly necessary, it has a nice _least privilege_ feel to it, though as @jakirkham points out, it _can_ be handy to hack the recipe in-place for a tighter debugging loop, but this can be accomplished by doing your own docker thing. If `ro` is the default, perhaps this could also be parameterized.
- removes the (apparently) extraneous `$RECIPE_ROOT`... 
  - might be ignorance on my part, can certainly be convinced this is still needed
- creates(outside) and mounts the `build_artifacts` as read-write
  - though honestly, I've often wanted different containers to share the _same_ `build_artifacts`, but that's another story, and maybe a recipe (ha) for un-reproducible builds... would save some steps with orchestrating multiple feedstock updates...
- for #637, detects `Enforcing` selinux hosts and adds `Z`
  - This helps people running SELinux from getting mysterious `permission denied` errors, using a similar mechanism to the `docker-machine` check and config.
  - An alternative would be to _not_ detect selinux, and offer local runners of the docker-based scripts to provide additional volume options, which are numerous.
    - In this case, to help folks like myself, it could preflight whether `build_artifacts` is writeable and `feedstock_root` is readable, and perhaps suggest use of the escape hatch.

```
Checking if /home/conda/feedstock_root is readable.... FAIL
Checking if /home/conda/build_artifacts is writeable... FAIL

   It looks like the feedstock is not writeable. If you are running with SELinux 
  (check with `getenforcing`), you can try running:
   
       VOLUME_OPTIONS=Z ./.circle_ci/run_docker_build.sh

   Note that this changes SELinux labels ON YOUR HOST SYSTEM, and can lead to 
   instability if used too broadly, like $HOME or /. This script will only change:

      /home/weg/projects/conda-forge/feedstocks/gawk-feedstock
      /home/weg/projects/conda-forge/feedstocks/gawk-feedstock/build_artifacts

Aborting!
```

This was tested with `gawk-feedstock`, which exhibited a number of exciting permission denied errors before some patches were made to the non-root anvil (which is a very good thing!).

### DIFF
--- a/conda_smithy/templates/run_docker_build.tmpl
+++ b/conda_smithy/templates/run_docker_build.tmpl
@@ -19,7 +19,7 @@
 # benefit from the improvement.
 
 FEEDSTOCK_ROOT=$(cd "$(dirname "$0")/.."; pwd;)
-RECIPE_ROOT=$FEEDSTOCK_ROOT/{{ recipe_dir }}
+ARTIFACTS_ROOT=$FEEDSTOCK_ROOT/build_artifacts
 
 docker info
 
@@ -31,7 +31,7 @@ channels:
 {%- endfor %}
 
 conda-build:
- root-dir: /home/conda/feedstock_root/build_artifacts
+ root-dir: /home/conda/build_artifacts
 
 show_channel_urls: true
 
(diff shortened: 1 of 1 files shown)


jakirkham:
> though honestly, I've often wanted different containers to share the _same_ `build_artifacts`, but that's another story, and maybe a recipe (ha) for un-reproducible builds... would save some steps with orchestrating multiple feedstock updates...

I get the impression that was the original design intent. Though it predates my involvement. :)

> detects `Enforcing` selinux hosts and adds `Z`

Reraising this question from the issue (sorry if I just missed your answer). What happens if we just always add `Z` regardless of whether selinux is used as host? Any downsides you know of?

jakirkham:
Sorry just saw the `Z` comment in the issue. Let's go with an environment variable if that is ok with you.

bollwyvl:
Yeah, that's great.

I created two variables:
```
CF_FEEDSTOCK_VOLUME_OPTIONS
CF_ARTIFACT_VOLUME_OPTIONS
```

Such that now you can solve either the in-container editing or selinux hack (or both).

Here's what my alias looks like now:
```bash
CF_FEEDSTOCK_VOLUME_OPTIONS="ro,Z" CF_ARTIFACTS_VOLUME_OPTIONS="rw,Z" ./.circleci/run_docker_build.sh 
```

Happy to rename, add doc (someplace), etc.

jakirkham:
This looks to be resolved in PR ( https://github.com/conda-forge/conda-smithy/pull/888 ). If not, would you be able to rebase this @bollwyvl?

bollwyvl:
Not sure, I'll have a look!

PR #647, opened 2018-01-29
Appveyor: optimize startup time

xoviat:
This is one of two PRs. The idea is that rather than solving the environment multiple times (which is done currently), the environment is solved once.


jakirkham:
Made a [comment]( https://github.com/conda-forge/conda-forge-build-setup-feedstock/pull/88#issuecomment-361383052 ) about in PR ( https://github.com/conda-forge/conda-forge-build-setup-feedstock/pull/88 ). For simplicity, let's keep the discussion in that PR (arbitrary choice) at least until we make some decision about how to proceed.

conda_smithy/templates/appveyor.yml.tmpl
```diff
@@ -32,7 +32,6 @@ install:
 
     # Add path, activate `conda` and update conda.
     - cmd: call %CONDA_INSTALL_LOCN%\Scripts\activate.bat
-    - cmd: conda.exe update --yes --quiet conda
```
jakirkham: FWIW this is a holdover from when we use to support Python 3.4. Given we don't do that any more and `conda` is generally recent enough, it seems reasonable to drop.

PR #702, opened 2018-03-04
fail rerender if matrix found in conda-forge.yml

### LINKED ISSUE #700: conda-forge.yml matrix ignored on rerender?
I'm trying to test rerendering with conda-smithy master and it's failing in a few cases. The first example I tried was https://github.com/conda-forge/mpi-feedstock which is an empty metapackage with two MPI_VARIANT matrix entries in conda-forge.yml. When I rerendered, I got only a single `config_.yml` without the MPI_VARIANTs.

Is this a bug in rerender or is there an extra step I'm missing to migrate conda-forge.yml matrix entries to conda_build_config.yml?

minrk:
since rerendering will be wrong if it proceeds ignoring the matrix.

point to new matrix location in recipe/conda_build_config.yaml and wiki for reference on migrating

This points to the release notes on the wiki, but in #700 @jakirkham mentioned that this will end up in the docs closer to release. The link should probably be updated if/when that happens.

closes #700

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -583,6 +583,19 @@ def _load_forge_config(forge_dir, exclusive_config_file):
     else:
         with open(forge_yml, "r") as fh:
             file_config = list(yaml.load_all(fh))[0] or {}
+
+        # check for conda-smithy 2.x matrix which we can't auto-migrate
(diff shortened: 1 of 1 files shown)


jakirkham:
LGTM. Thanks @minrk.

PR #740, opened 2018-04-17
Initial rever

scopatz:
Adds rever support in order to issue releases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ conda_smithy.egg-info/
 dist/
 .cache
 .DS_Store
+
+# Rever
+rever/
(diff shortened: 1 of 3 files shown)


ocefpaf:
No need to wait for the CIs here b/c there is new conda in conda-smithy to test, right?

jakirkham:
Thanks.

Having not really used `rever`, how does the tagging process work? Does it give us the option to add things to an annotated tag?

ocefpaf:
Thanks @scopatz for the release! BTW, should we drop the `rc` in the tag? We got `3.0.0rc3`

PR #771, opened 2018-05-13
travis API version handling

dougalsutherland:
Our current `travis_headers` confusingly requests both API V2 and V3.

This changes it to request only V3, except for the auth action which only works on V2 (https://github.com/travis-ci/travis-ci/issues/9273#issuecomment-370474214).

Don't know if this is actually related to https://github.com/conda-forge/staged-recipes/issues/5846, but it could be. :man_shrugging: 

(Note that I haven't gone through all our uses of `travis_headers` to make sure we're actually using the V3 API everywhere. It's very possible this could break something....)

### DIFF
--- a/conda_smithy/ci_register.py
+++ b/conda_smithy/ci_register.py
@@ -45,17 +45,23 @@ travis_endpoint = 'https://api.travis-ci.org'
 
 def travis_headers():
     headers = {
-               # If the user-agent isn't defined correctly, we will recieve a 403.
-               'User-Agent': 'Travis/1.0',
-               'Accept': 'application/vnd.travis-ci.2+json',
-               'Content-Type': 'application/json',
-               'Travis-API-Version': '3'
-               }
-    url = '{}/auth/github'.format(travis_endpoint)
-    data = {"github_token": github.gh_token()}
+       # If the user-agent isn't defined correctly, we will recieve a 403.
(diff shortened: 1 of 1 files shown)


dougalsutherland:
This came up because I was investigating https://github.com/conda-forge/staged-recipes/issues/5846 and got a 404 on `/auth/github` locally when running `conda_smithy.ci_register.travis_headers()` without a `~/.conda-smithy/travis.token`:

```
$ python -c 'from conda_smithy.ci_register import *; print(travis_headers())'
  File "<string>", line 1, in <module>
  File "/nfs/nhome/live/dougals/anaconda/lib/python3.6/site-packages/conda_smithy/ci_register.py", line 60, in travis_headers
    response.raise_for_status()
  File "/nfs/nhome/live/dougals/anaconda/lib/python3.6/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://api.travis-ci.org/auth/github
```

With this change, that at least works, and gets me an appropriate token in `Authorization`.

dougalsutherland:
(Note that this isn't actually the cause of https://github.com/conda-forge/staged-recipes/issues/5846, because `trigger_travis_build.py` patches up the headers appropriately.)

Issue #318, opened 2016-09-29
Specify branch name to associate with feedstock's source repo


shadowwalkersb:
Currently, `conda-smithy` sets up CI badges on the feedstock repository to show current build status of `master` branches on Travis CI and Appveyor and no branch is specified for Circle CI. It would be nice to be able to specify the branch name to be used for the CI badges. At this point, I am not sure if they would be useful for anything else. The discussion first was brought up [here](https://github.com/conda-forge/conda-smithy/pull/292#discussion_r81026056). Initial thoughts were expressed [here](https://github.com/conda-forge/conda-smithy/pull/292#discussion_r81134874). In summary, these are the concerns raised (I tried to extract the important parts, the full comments are in the code review listed above).

1. The name of a branch should not live in the repo itself (otherwise you end up having to make a code change just to take a branch).
2. It would result in redundancy of information, and would require a code change whenever we want a new branch. I would consider automatically determining this, but even then, there are caveats (like having a headless checkout) that we would have to consider carefully.

jakirkham:
Thanks for opening this issue. Saw the conversation that was going on along these lines in PR ( https://github.com/conda-forge/conda-smithy/pull/292 ) and was about to do the same thing to help consolidate the conversation.

There are two other use cases I can already think of right now.

1. White listing branches for release ( https://github.com/conda-forge/conda-forge.github.io/issues/87 )
2. Failing CircleCI builds on merge conflict ( https://github.com/conda-forge/conda-smithy/issues/316 )

For case 1, @asmeurer and I discussed came to the conclusion that it may be better to not have a list of branches to release from, but only specify the branch that we are on for release (or not). Read the associated issue for more details. These aren't totally aligned, but it may be worth thinking about what we want the badges to represent (e.g. just failure for that branch or failure for release branches) and therefore how aligned these cases are.

In the case of 2, knowing the release branch one is merging into may be a good temporary workaround (for the CI PR merge workaround https://github.com/conda-forge/conda-smithy/pull/309 😆) to catch and fail CircleCI builds where there is a merge conflict.

So these are some other related cases for knowing the branch. However, we should discuss more to see how well these overlap with your goals here or not. If you have time, reading the linked issues may help in answering this question.

shadowwalkersb:
> These aren't totally aligned, but it may be worth thinking about what we want the badges to represent (e.g. just failure for that branch or failure for release branches) and therefore how aligned these cases are.

[...]

PR #802, opened 2018-06-12
Allow custom branch in repo for badges

hmaarrfk:
For a project of mine, I'm trying to stay close to the upstream conda-forge-feedstock, but would prefer to have the badges generated for my own branch.

Does this do what I want it to do?

Do I need to add something around these lines of code:
https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/github.py#L99

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -685,7 +685,9 @@ def _load_forge_config(forge_dir, exclusive_config_file):
               'linux': {'enabled': False},
               'channels': {'sources': ['conda-forge', 'defaults'],
                            'targets': [['conda-forge', 'main']]},
-              'github': {'user_or_org': 'conda-forge', 'repo_name': ''},
+              'github': {'user_or_org': 'conda-forge',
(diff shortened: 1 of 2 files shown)


jakirkham:
LGTM. Were you able to try it out?

shadowwalkersb:
Is this related to #318, by any chance?

hmaarrfk:
I would need to figure out how to install and test my spin of smithy. Maybe
I can just clobber my old install with these files.

Seems like #318 could be helped by this. I think he wants the branch name
to be read from git or smithy rerender automatically. Which may be
problematic with the fork/PR model.

jakirkham:
This is what I do to accomplish this. Maybe we should add it to the docs in another PR if it works ok for you.

```bash
conda activate
conda remove -y conda-smithy
conda install -y pip  # if `pip` isn't in `base`
pip install --no-deps -e <path to conda-smithy git repo> 
```

hmaarrfk:
@isuruf did you happen to test the patch?

PR #807, opened 2018-06-18
suggest pip

ocefpaf:
Addressees https://github.com/conda-forge/conda-smithy/pull/669#issuecomment-397668135

Instead of a lint fail if `noarch` is used let's suggest "use pip" globally whenever it is possible to use `pip`.

Ping @isuruf and @CJ-Wright

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -269,17 +269,6 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
                                      "`noarch: {}`.".format(build_section['noarch']))
                         break
 
-    # 18: noarch and python setup.py doesn't work
-    if build_section.get('noarch') == 'python':
-        if 'script' in build_section:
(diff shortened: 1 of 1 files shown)


jakirkham:
We could do both. IOW keep the error for `noarch` and suggest it elsewhere.

It's worth noting that if it the package is not `noarch: python`, it may be using `build.sh`/`bld.bat` or possibly have multiple steps in `script`. Not sure how we should address that. Just something to keep in mind.

ocefpaf:
> We could do both. IOW keep the error for `noarch` and suggest it elsewhere.

The error is wrong so no, we cannot keep both. There is `noarch` without `pip`.

> Not sure how we should address that.

We cannot address that in the linter at the moment and I would not worry about it since these are just a few. Merging this to fix a bad lint that is preventing users from updating their recipes.

jakirkham:
Are there examples where it was problematic?

PR #808, opened 2018-06-18
Allow setting appveyor.image in conda-forge.yml

### LINKED ISSUE #605: Use os: Visual Studio 2017 for Python 3.6 builds
The VS 2017 image has MSVC 2015 and VS2017 is also ABI compatible. There's no reason not to use the newer image.

minrk:
I need this to get Visual Studio 2017 for nodejs 10.

Testing a build rendered from this branch now: https://github.com/conda-forge/nodejs-feedstock/pull/41

At some point, the default image may want to be bumped, but the VS 2017 image doesn't appear to have the old-timey Python 2.7 compilers, so I think it would have to be conditional for a while.

closes #605

### DIFF
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -2,6 +2,10 @@
 # file, make changes to conda-forge.yml and/or recipe/meta.yaml, and run
 # "conda smithy rerender".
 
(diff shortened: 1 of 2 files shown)


jakirkham:
LGTM. Could you please add a note in the [`news` directory]( https://github.com/conda-forge/conda-smithy/tree/master/news )?

PR #838, opened 2018-07-16
Adjust `--no-check-uptodate` to take no argument

jakirkham:
This makes specifying `--no-check-uptodate` without an argument sufficient for skipping the update check. Preserves the default behavior of erroring out.


scopatz:
Thanks @jakirkham!

PR #864, opened 2018-08-06
Fix missing newline in last line of generated readmes and add unit test for it

### LINKED ISSUE #863: Autogenerated README.md missing final newline
## The Problem

As I've confirmed is the case on multiple repos here, including our own ``spyder-feedstock`` and ``spyder-kernels-feedstock`` as well as two arbitrary conda-forge repos I checked conda-forge, the last line in README.md lacks a terminating newline (LF/``x0D``), and is thus ill-formed. I'd be happy to submit a PR to fix it since I imagine it is probably pretty trivial, if someone more knowlegable than me can let me know how to approach it. 

## Proposed Solutions

A naive hack would seem to be just writing an additional ``\n`` [here](https://github.com/conda-forge/conda-smithy/blob/855f23bb96efb1cbdbdc5e60dfb9bbdd3e142d31/conda_smithy/configure_feedstock.py#L718), but editing the [template ](https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/templates/README.md.tmpl) would seem to make far more sense. However, the template *has* a trailing newline and hasn't been edited in a while, so not sure what's going on—is it not writing the last one; is it getting stripped, or what?

Thanks!

CAM-Gerlach:
Fixes #863 .

Currently, as described in the above issue, readmes generated with ``conda smithy rerender`` lack a final terminating newline, likely due to it being stripped from the jinja template somewhere. Therefore, this PR makes a small tweak to the template (``{# #}`` as suggested by @jakirkham ) to ensure a newline is present in the final output, as well as adds a simple unit test to check for the existence of the generated readme and that its end of file character is a newline (LF/``0x0A``), and fixes a couple instances in the repo's own files where that is not the case. Also, takes the opportunity to add pytest and np++ files to the gitignore, for convenience.

Thanks!


CAM-Gerlach:
How did coverage *decrease* a whopping 1.2%? I added a single line in a non-code template file (plus a few lines in the ``.gitignore`` and a newline at the end of two non-code files); the vast majority of lines added was *a unit test* that, if anything, should mean a net *increase* in coverage. Furthermore, looking at the actual[ Coveralls output](https://coveralls.io/builds/18336491), I don't see anywhere it indicates this supposed decrease took place; on every file it shows "no change" except for on ``configure_feedstock.py``, where it shows "coverage increased" by one branch (which is what my test covers). Uh, what?

scopatz:
Thanks for adding this @CAM-Gerlach! Can you please add a new entry.

CAM-Gerlach:
@scopatz Done.

.gitignore
```diff
@@ -10,3 +10,9 @@ dist/
 
 # Rever
 rever/
+
+# Notepad++ files
+nppBackup/
```
dougalsutherland: This kind of editor-specific thing should be in your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). doesn't particularly hurt anything to have it here, but we're not ignoring `*~` or other kinds of backup files....

CAM-Gerlach: Okay, thanks! For some reason I wasn't aware one could have a global ``.gitignore``, and never thought to look it up, so I appreciate the tip.

PR #866, opened 2018-08-06
Upgrade links to HTTPS and update link targets where necessary

CAM-Gerlach:
To conform to modern best practices, this PR upgrades all public-facing links to HTTPS where available, and update link targets if broken. redirected or otherwise clearly out of date. All links were carefully checked by hand to confirm whether they were fully functional both before and after (attempted) HTTPS conversion. HTTP links that were present for example or testing purposes, or that were XML namespace URNs were not altered, as were any links that did not fully function when tested over HTTPS. 

The ``get_conda_hook_info()`` URLs *were* updated in a separate commit as they did automatically redirect on trying them, but please double-check this. Also, in addition to updating the URL for the xkcd webcomic, it was properly named, linked and attributed (as is legally and ethically *required* by its CC-BY-SA 2.5 license), and  the title-text was included for completeness.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -19,5 +19,5 @@ exclude_lines =
     # Ignore continue statement in code as it can't be detected as covered
     # due to an optimization by the Python interpreter. See coverage issue
     # ( https://bitbucket.org/ned/coveragepy/issue/198/continue-marked-as-not-covered )
-    # and Python issue ( http://bugs.python.org/issue2506 ).
+    # and Python issue ( https://bugs.python.org/issue2506 ).
     continue
--- a/README.md
+++ b/README.md
@@ -102,4 +102,9 @@ $ rever 0.1.2
 Conda-smithy in a nutshell
 --------------------------
 
-![tools](http://imgs.xkcd.com/comics/tools.png)
+xkcd 1629: Tools
+~~~~~~~~~~~~~~~~
+
+[![xkcd 1629: Tools](https://imgs.xkcd.com/comics/tools.png)](https://xkcd.com/1629/)
+**Titletext**: *I make tools for managing job-hunting sites for people who make*
+*tools for managing job-hunting sites for people who make tools for ...*
--- a/conda_smithy/ci_register.py
+++ b/conda_smithy/ci_register.py
@@ -317,7 +317,7 @@ def travis_encrypt_binstar_token(repo, string_to_encrypt):
     #    not use this file except in compliance with the License. You may obtain
     #    a copy of the License at
     #
-    #         http://www.apache.org/licenses/LICENSE-2.0
+    #         https://www.apache.org/licenses/LICENSE-2.0
     #
     #    Unless required by applicable law or agreed to in writing, software
     #    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
--- a/conda_smithy/ci_register.py
+++ b/conda_smithy/ci_register.py
@@ -398,25 +398,25 @@ def add_conda_forge_webservice_hooks(user, repo):
 
     hooks = [
         get_conda_hook_info(
-            "http://conda-forge.herokuapp.com/conda-linting/hook",
+            "https://conda-forge.herokuapp.com/conda-linting/hook",
             [
                 "pull_request"
             ]
         ),
         get_conda_hook_info(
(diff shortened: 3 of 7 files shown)


CAM-Gerlach:
Something's clearly going wrong with coveralls, since on both this PR and #864 it shows a decease of 1.2% when there should be no net change, or even a slight increase.

scopatz:
> Something's clearly going wrong with coveralls, since on both this PR and #864 it shows a decease of 1.2% when there should be no net change, or even a slight increase.

Yeah, coveralls and coverage I have found to not be generally trustworthy.

scopatz:
Thanks for putting this in @CAM-Gerlach - can you please add a news entry.  See the following for instructions: http://xon.sh/devguide.html#changelog

CAM-Gerlach:
@scopatz Sorry for the delay (was working on some improvements to a Spyder PR); done as requested. 

I didn't consider doing so initially, since I'm so used to all of that being done automatically without all the repetitive tedium through our very own [loghub](https://github.com/spyder-ide/loghub) package across all our projects (and a number of others also use it). I'd recommend checking it out as an alternative to the current manual system, since it basically updates your NEWS/changelog and announcements automatically each release with titles, numbers and links of the milestoned issues closed and PRs merged  with each release, as well as the total stats, all in a nicely and consistently formatted manner with customization options. 

This avoids almost all the repetitive effort, overhead, inconsistencies and potential for errors/omissions of the current system, while producing better looking, richer and more informative changelogs/news/announcements with essentially no effort on your part. As an extra bonus, it makes it easy to tweak the wording of the news message presented to your liking—just change the desired issue or PR title before you merge or release right from the Github web UI, and you can always modify the result to your liking or, e.g. add a summary of the most important/impactful changes with each release up top (as we do).

Of course, there may be reasons why this may not work as well for you as for us, but just wanted to put that out there. Thanks!

CAM-Gerlach:
> I am not sure that is a direction that I'd like to go in. There is a choice to be made here with regards to whether you fully automate changelogs or not.

@scopatz I 100% agree that a human should be in the loop during the changelog process, and it shouldn't be "fully" automated; furthermore, there are *also* reasons why a loghub-like workflow wouldn't be ideal for all projects. However, I don't believe these sets have much overlap given what Loghub actually does, since it *isn't* "fully automated" (although really, no changelog system truly is)—rather, it only (better) automates the tedious and inconsistent part of *your existing* process, the part you already have some (more primitive) automation involved with i.e. going from a PRs to a NEWS entry. Essentially, it just makes what you *already do* easier, while keeping a human in the loop as much as before (both at the PR stage, and the compilation/release stage).

As implied previously, it *doesn't* work on the basis of commits (which I agree would be a terrible idea for almost all projects) but rather issues and/or PRs (at your option), which naturally group conceptually related changes, *already* seems to be more or less the unit on which NEWS entries are based, and of which the (properly descriptive) titles are already around the form and length of most of your changelog entries (and that latter can easily be edited by both the submitter or org members at any time before release if it needs refinement to be more easily understandable, unlike a commit to NEWS). 

As also discussed, the resulting announcement/changelog can easily be edited (as I assume is currently the case) to add or remove entries as desired, or to e.g. summarize the most user-impactful enhancements/changes/deprecations/removals (as we routinely do every release, although that may not even really be necessary for you most of the time, since you don't have dozens or hundreds of issues/PRs per release as we do and is aimed at a less general audience).

> We are using rever to create the changelog - and more generally perform the release management. So even though we likely won't go the loghub route here, it would be great to have loghub integration in rever itself!

I see, thanks! We were actually looking into a few options to do the same for Spyder, but of the ones we were familiar from previous work, they were all either unmaintained, had problems, were too simple, etc. I'll float the idea and see what the others thing about how ``loghub`` might integrate with ``rever``—or not. Thanks!

scopatz:
That'd be great if Spyder would start using rever! Let me know if there is anything I can do to help.  Rever is all about adding custom release activities and integrations :)

CAM-Gerlach:
@goanpeca , one of our core devs, said he'd be quite open to it, presuming it provides equivalent or better features and is well maintained; it seemed the only likely stumbling block would be integrating ``loghub`` into ``rever`` as a custom activity or the like—not sure about that developer bandwidth being available right now at least on our side, but if it isn't too hard then its definitely something I'd think we'd want to at least consider for sometime in the immediate future. Thanks!

PR #871, opened 2018-08-10
get_section: check ABCs instead of list/dict subclass

dougalsutherland:
`ruamel.yaml` version 0.15.52 [changed `CommentedMap` to no longer subclass from `dict`](https://bitbucket.org/ruamel/yaml/commits/ace40a6a40270b75320971d1c9ff7625fbf6e081#Lcomments.pyF594). This is breaking `get_section` / `get_list_section` and causing all kinds of spurious errors / breaking real checks: [example](https://github.com/conda-forge/conda-forge-webservices/pull/54#issuecomment-412106931). Checking based on ABCs instead should still work.

I also removed a few no-longer-used imports from the top of this file while I was at it.

PR #872, opened 2018-08-10
don't complain about bioconda for packages with no name

dougalsutherland:
Currently, if a recipe doesn't have a `name`, it'll ping `bioconda-recipes` ([example](https://github.com/conda-forge/conda-forge-webservices/pull/54#issuecomment-412106931)). That seems not ideal. :)


dougalsutherland:
Failures all would be fixed by #871.

PR #873, opened 2018-08-10
Drop `vendored` package

jakirkham:
This is a leftover directory from when we vendored `travis_encrypt.py`, which we no longer include. Given this, go ahead and drop this directory.


scopatz:
Thanks @jakirkham

PR #879, opened 2018-08-22
Allow for variants to pass through the linter

hmaarrfk:
I guess we have to decide if we want to support this first.

https://github.com/jjhelmus/selectable_pkg_variants_example


scopatz:
Thoughts @conda-forge/core

PR #888, opened 2018-09-05
conda_smithy/templates/run_docker_build.sh.tmpl: add SELinux flags to Docker volumes

pkgw:
If you're running Docker builds on an SELinux-enabled Linux system, the mounted volumes need to have the appropriate SELinux flags set if they are to be used by their containers. Here we add the magic `z` option to make this happen. We also mount the recipe directory read-only since the container shouldn't need to write to it.

Full disclosure: not tested locally but it *should* work ...


scopatz:
Thanks @pkgw!

PR #895, opened 2018-09-17
Toolchain3 for update-cb3

scopatz:
see https://github.com/regro/cf-scripts/issues/363

### DIFF
--- a/conda_smithy/update_cb3.py
+++ b/conda_smithy/update_cb3.py
@@ -70,7 +70,7 @@ class Section:
                 return Section(sect, start, end)
         return Section(sect, start, self.end)
 
-        
+
 def iterate(tarzip):
     if isinstance(tarzip, zipfile.ZipFile):
         for f in iter(tarzip.infolist()):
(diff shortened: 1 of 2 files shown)


scopatz:
@conda-forge/core This is ready for review!

jakirkham:
Looks good. Were you able to try it locally?

scopatz:
@jakirkham - nope I didn't try it locally, but I can.

scopatz:
Seems to work locally:

```console
scopatz@artemis ~/feedstocks/hunspell master $ git checkout -b compiler_migration2 bot/compiler_migration2
Branch 'compiler_migration2' set up to track remote branch 'compiler_migration2' from 'bot'.
Switched to a new branch 'compiler_migration2'
scopatz@artemis ~/feedstocks/hunspell compiler_migration2 $ conda smithy update-cb3
List of changes done to the recipe:
Renamed build with host
Moving autoconf from host to build
Moving automake from host to build
Moving libtool from host to build
Moving make from host to build
Removing toolchain3 in favour of compiler()
Adding C compiler
Adding C++ compiler
scopatz@artemis ~/feedstocks/hunspell compiler_migration2 $ git diff
diff --git a/recipe/meta.yaml b/recipe/meta.yaml
index d15ddf4..7c35449 100644
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -26,7 +26,9 @@ requirements:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
   host:
-    - toolchain3
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
     - gettext
     - readline
   run:
```

jakirkham:
Hmm...did it add a second `host` section? Also looks like `compiler` was added to `host` as well. 😕

Admittedly as this appears to be a partially migrated recipe, this is a bit more complicated than one that hadn't been migrated at all. Not sure we should worry too much about how this behaves on partially migrated recipes, but I'll leave it up to you.

The code itself seems like it should work on an unmigrated recipe. IDK if there are any of those to still test. Maybe an earlier commit of this one would suffice?

scopatz:
I think in situations like this where there has already been a partial migration (or syntax migration), we'll have to manually edit the files somewhat.  Some of these tools might actually need toolchain3 for the old compilers.  There should not be that many of them.

@mariusvniekerk - what is the selector that is used for the gcc7 label switching?

jakirkham:
Thus far simply removing `toolchain3` from partially migrated recipes has worked nicely in my experience.

scopatz:
Yeah, removing toolchain3 seems to work so far.

jakirkham:
Well I'm ok with merging this if you are. WDYT?

scopatz:
I think it is OK to merge this, because it at least bring toolchain3 in-line with toolchain, even if the solution is to ultimately remove toolchain3 from the recipe.

mariusvniekerk:
@scopatz adding `compiler_stack: comp7` to conda-forge-congif.yaml makes the magic go

jakirkham:
Thanks @scopatz 😄

scopatz:
Thanks @jakirkham @mariusvniekerk!

PR #901, opened 2018-09-26
[Ready for review] Initial work on recipes for azure-pipelines

mariusvniekerk:
This probably needs some more pieces.

For now this is configured to just generate for all platforms.

### DIFF
--- a/conda_smithy.recipe/meta.yaml
+++ b/conda_smithy.recipe/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - pygithub <2
     - ruamel.yaml
     - conda-forge-pinning
-    # - vsts-python-api
+    - vsts-python-api
 
 test:
   requires:
--- a/conda_smithy.recipe/meta.yaml
+++ b/conda_smithy.recipe/meta.yaml
@@ -29,6 +29,7 @@ requirements:
     - pygithub <2
     - ruamel.yaml
     - conda-forge-pinning
+    # - vsts-python-api
 
 test:
   requires:
--- a/conda_smithy/azure_ci_utils.py
+++ b/conda_smithy/azure_ci_utils.py
@@ -1,0 +1,187 @@
+import os
+import typing
+
+from vsts.vss_connection import VssConnection
+from msrest.authentication import BasicAuthentication
+import pprint
+
+
+AZURE_TEAM_INSTANCE = os.getenv("AZURE_INSTANCE", "https://dev.azure.com/conda-forge")
+AZURE_PROJECT_ID = os.getenv("AZURE_PROJECT_ID", "feedstock-builds")
+AZURE_SERVICE_ENDPOINT_NAME = os.getenv("AZURE_SERVICE_ENDPOINT", "conda-forge")
+
+try:
+    # Create a token at https://circleci.com/account/api. Put it in circle.token
+    with open(os.path.expanduser('~/.conda-smithy/azure.token'), 'r') as fh:
+        AZURE_TOKEN = fh.read().strip()
+    if not AZURE_TOKEN:
+        raise ValueError()
+except (IOError, ValueError):
+    print('No circle token.  Create a token at https://dev.azure.com/conda-forge/_usersSettings/tokens and\n'
+          'put it in ~/.conda-smithy/azure.token')
+
+
+credentials = BasicAuthentication("", AZURE_TOKEN)
+connection = VssConnection(base_url=AZURE_TEAM_INSTANCE, creds=credentials)
+
+
+def get_service_endpoint(project_id=AZURE_PROJECT_ID):
+    from vsts.service_endpoint.v4_1.service_endpoint_client import ServiceEndpointClient
+    from vsts.service_endpoint.v4_1.models import ServiceEndpoint
+
+    service_endpoint_client = ServiceEndpointClient(
+        base_url=AZURE_TEAM_INSTANCE, creds=credentials
+    )
+    endpoints: typing.List[
+        ServiceEndpoint
+    ] = service_endpoint_client.get_service_endpoints(project=project_id, type="GitHub")
+    for service_endpoint in endpoints:
+        if service_endpoint.name == AZURE_SERVICE_ENDPOINT_NAME:
+            return service_endpoint
+    else:
+        raise KeyError("Service endpoint not found")
+
+
+def get_queues(project_id=AZURE_PROJECT_ID):
+    from vsts.task_agent.v4_0.task_agent_client import TaskAgentClient
+    from vsts.task_agent.v4_0.models import TaskAgentQueue
+
+    aclient = TaskAgentClient(AZURE_TEAM_INSTANCE, credentials)
+    queues: typing.List[TaskAgentQueue] = aclient.get_agent_queues(project_id)
+    return queues
+
+
+def get_default_queue(project_id):
+    queues = get_queues(project_id)
+    for q in queues:
+        if q.name == "Default":
+            return q
+    else:
+        raise ValueError("Default queue not found")
+
+
+def get_repo_reference(project_id, github_org, repo_name):
+    from vsts.build.v4_1.build_client import BuildClient
+    from vsts.build.v4_1.models import SourceRepositories, SourceRepository
+
+    service_endpoint = get_service_endpoint(project_id)
+    bclient: BuildClient = connection.get_client(
+        "vsts.build.v4_1.build_client.BuildClient"
+    )
+    repos: SourceRepositories = bclient.list_repositories(
+        project=project_id,
+        provider_name="github",
+        repository=f"{github_org}/{repo_name}",
+        service_endpoint_id=service_endpoint.id,
+    )
+
+    repo: SourceRepository = repos.repositories[0]
+    return repo
+
+
+def register_repo(github_org, repo_name, project_id=AZURE_PROJECT_ID):
+    from vsts.build.v4_1.models import (
+        BuildDefinition,
+        BuildDefinitionReference,
+        BuildRepository,
+    )
+    from vsts.build.v4_1.build_client import BuildClient
+    import inspect
+    bclient: BuildClient = connection.get_client(
+        "vsts.build.v4_1.build_client.BuildClient"
+    )
+
+    source_repo = get_repo_reference(project_id, github_org, repo_name)
+
+    new_repo = BuildRepository(
+        type="GitHub",
+        url=source_repo.properties["cloneUrl"],
+        **{
+            k: v
+            for k, v in source_repo.as_dict().items()
+            if k in set(inspect.getfullargspec(BuildRepository).args) - {"url"}
+        }
+    )
+    new_repo.name = source_repo.properties["fullName"]
+    new_repo.properties["cleanOptions"] = "0"
+    new_repo.properties["skipSyncSource"] = "false"
+    new_repo.properties["gitLfsSupport"] = "false"
+    new_repo.properties["checkoutNestedSubmodules"] = "false"
+    new_repo.properties["labelSources"] = "0"
+    new_repo.properties["fetchDepth"] = "0"
+    new_repo.properties["labelSourcesFormat"] = "$(build.buildNumber)"
+    new_repo.properties["reportBuildStatus"] = "false"
+    new_repo.clean = False
+
+    queues = get_queues(project_id)
+    default_queue = get_default_queue(project_id)
+    service_endpoint = get_service_endpoint(project_id)
+
+    build_definition = BuildDefinition(
+        process={
+            "type": 2,
+            # These might be optional;
+            "resources": {
+                "queues": [{"id": q.id, "alias": q.name} for q in queues],
+                "endpoints": [
+                    {"id": service_endpoint.id, "alias": service_endpoint.name}
+                ],
+            },
+            "yamlFilename": "/azure-pipelines.yml",
+        },
+        # queue works
+        queue=default_queue,
+        # now onto this
+        repository=new_repo,
+        name=repo_name,
+        # configure trigger for our builds.
+        triggers=[
+            {
+                "branchFilters": ["+*"],
+                "forks": {"enabled": True, "allowSecrets": False},
+                "pathFilters": [],
+                "isCommentRequiredForPullRequest": False,
+                "triggerType": "pullRequest",
+            },
+            {
+                "branchFilters": ["+*"],
+                "pathFilters": [],
+                "batchChanges": False,
+                "maxConcurrentBuildsPerBranch": 1,
+                "pollingInterval": 0,
+                "triggerType": "continuousIntegration",
+            },
+        ],
+        type="build",
+    )
+
+    # clean up existing builds for the same feedstock if present
+    existing_definitions: typing.List[
+        BuildDefinitionReference
+    ] = bclient.get_definitions(project=project_id, name=repo_name)
+    if existing_definitions:
+        assert len(existing_definitions) == 1
+        ed = existing_definitions[0]
+        bclient.update_definition(
+            definition=build_definition, definition_id=ed.id, project=ed.project
+        )
+    else:
+        bclient.create_definition(
+            definition=build_definition, project=project_id
+        )
+
+
+def repo_registered(github_org, repo_name, project_id=AZURE_PROJECT_ID):
+    from vsts.build.v4_1.models import (
+        BuildDefinitionReference,
+    )
+    from vsts.build.v4_1.build_client import BuildClient
+    bclient: BuildClient = connection.get_client(
+        "vsts.build.v4_1.build_client.BuildClient"
+    )
+
+    existing_definitions: typing.List[
+        BuildDefinitionReference
+    ] = bclient.get_definitions(project=project_id, name=repo_name)
+
+    return bool(existing_definitions)
--- a/conda_smithy/azure_ci_utils.py
+++ b/conda_smithy/azure_ci_utils.py
@@ -114,7 +114,7 @@ def register_repo(github_org, repo_name, project_id=AZURE_PROJECT_ID):
     new_repo.properties["labelSources"] = "0"
     new_repo.properties["fetchDepth"] = "0"
     new_repo.properties["labelSourcesFormat"] = "$(build.buildNumber)"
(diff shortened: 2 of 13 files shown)


mariusvniekerk:
Requires: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/29

scopatz:
Seems good to me, but the tests are failing.

conda_smithy/configure_feedstock.py
```diff
@@ -758,6 +763,52 @@ def render_appveyor(jinja_env, forge_config, forge_dir):
...
+        ],
+        'linux': [
+            os.path.join(forge_dir, '.azure-pipelines', 'run_docker_build.sh'),
+            os.path.join(forge_dir, '.azure-pipelines', 'azure-pipelines-linux.yml'),
+        ],
+        'osx': [
+            os.path.join(forge_dir, '.azure-pipelines', 'azure-pipelines-osx.yml'),
+        ],
+        'win': [
+            os.path.join(forge_dir, '.azure-pipelines', 'azure-pipelines-osx.yml'),
```
dhirschfeld: `azure-pipelines-osx.yml` seems a bit suspect no?

conda_smithy.recipe/meta.yaml
```diff
@@ -29,6 +29,7 @@ requirements:
     - pygithub <2
     - ruamel.yaml
     - conda-forge-pinning
+    # - vsts-python-api
```
scopatz: Should this line be removed?

conda_smithy/azure_ci_utils.py
```diff
@@ -0,0 +1,187 @@
+import os
+import typing
+
+from vsts.vss_connection import VssConnection
```
scopatz: Do we need a conda-forge package for this?

conda_smithy/azure_ci_utils.py
```diff
@@ -0,0 +1,187 @@
...
+from msrest.authentication import BasicAuthentication
```
scopatz: And this?

scopatz:
Overall looks great! Thanks soooo much for your hard work here @mariusvniekerk!

scopatz:
Is there anything we can do for testing this?

dhirschfeld:
I can't wait for this to get in - the wait on appveyor is getting unbearable!

conda_smithy/azure_ci_utils.py
mariusvniekerk: yes, I've started the process

conda_smithy/azure_ci_utils.py
mariusvniekerk: oddly enough nope

isuruf:
Can you re-use `conda-smithy/conda_smithy/templates/run_osx_build.sh.tmpl` for osx?

mariusvniekerk:
@isuruf probably for some of it.  I based most of the osx bit by harvesting pieces of the travis stuff

requirements.txt
```diff
@@ -10,3 +10,4 @@ gitpython
 pygithub <2
 ruamel.yaml
 conda-forge-pinning
+vsts-python-api
```
minrk: The Python package is just called `vsts`

conda_smithy/azure_ci_utils.py
```diff
@@ -0,0 +1,191 @@
...
+    else:
+        raise ValueError("Default queue not found")
+
+
+def get_repo_reference(project_id, github_org, repo_name):
+    from vsts.build.v4_1.build_client import BuildClient
+    from vsts.build.v4_1.models import SourceRepositories, SourceRepository
+
+    service_endpoint = get_service_endpoint(project_id)
+    bclient: BuildClient = connection.get_client(
```
scopatz: Well, I think I found a bug in xonsh :)

scopatz:
This LGTM - are there any barriers to merging this?

isuruf:
One thing I'd like to see is to remove all the duplication. For eg: `run_docker_build.sh` is added twice in a feedstock.

requirements.txt
minrk: True, but this is a requirements.txt, so it should have the Python name, no?

minrk:
When I [tested this on nodejs](https://github.com/conda-forge/nodejs-feedstock/pull/52), `build_setup` in [build_steps.sh](https://github.com/conda-forge/nodejs-feedstock/pull/52/commits/45368390da0967e99ffeb7a42eddb48abf050270) had Windows syntax instead of shell syntax. Maybe a mutable dict getting reused, perhaps?

The main reason I was trying this is that the nodejs appveyor builds keep running out of time and/or memory, and azure apparently has a 6 hour time limit. Interestingly, building on azure results in running out of [disk space](https://dev.azure.com/minrk/toolchain-tests/_build/results?buildId=3&view=logs), which is a new experience.

willsmythe:
@minrk - thanks for pointing out the disk space problem on Windows. I think the built-in Azure Pipelines "conda environment" task is incorrectly using C: for the environment (it should be using D:). I'll find out what's up and report back when I have an answer.

minrk:
@willsmythe thanks! I tested running in D: and [still ran out of space](https://dev.azure.com/minrk/toolchain-tests/_build/results?buildId=7&view=logs). Both C and D start with ~10G free and apparently nodejs needs more than 10G of disk to build on Windows (?!).

willsmythe:
@minrk - thanks. I started digging into this as well. To help isolate the root cause, I forked the nodejs/node repo, created a pipeline that calls `vcbuild.bat` with the same arguments, and hit the same "out of disk space" problem. I have a thread started with the engineering team to figure out what's going on.

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
```diff
@@ -0,0 +1,83 @@
...
+      {{ config }}:
+        CONFIG: {{ config}}
+    {%- endif %}
+    {%- endfor %}
+
+  steps:
+  # TODO: Fast finish on azure pipelines?
+  - script: |
+      echo "Fast Finish"
+      {{ fast_finish }}
```
DavidStaheli: Unfortunately, Azure Pipelines doesn't yet support fast finish of jobs created by a matrix.

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -0,0 +1,94 @@
...
+    # TODO: Fast finish on azure pipelines?
+    - script: |
+        {{ fast_finish }}
+
+    - script: |
+        choco install vcpython27 -fdv -y --debug
+      condition: contains(variables['CONFIG'], 'vs2008')
+      displayName: Install vcpython27.msi (if needed)
+
+    # Cywing's git breaks conda-build. (See https://github.com/conda-forge/conda-smithy-feedstock/pull/2.)
```
DavidStaheli:
```suggestion
    # Cygwin's git breaks conda-build. (See https://github.com/conda-forge/conda-smithy-feedstock/pull/2.)
```

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
```diff
@@ -0,0 +1,83 @@
...
+
+  - script: |
+      echo "Installing Miniconda"
+      set -x -e
+      curl -o $(Build.StagingDirectory)/miniconda.sh https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
+      chmod +x $(Build.StagingDirectory)/miniconda.sh
+      $(Build.StagingDirectory)/miniconda.sh -b -p $(Build.StagingDirectory)/miniconda
+      export PATH=$(Build.StagingDirectory)/miniconda/bin:$PATH
+      echo "Setting up Conda environment"
+    displayName: 'Install miniconda'
```
DavidStaheli: Is installing Miniconda necessary still necessary if version 4.5.x is [already installed](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-Readme.md#package-management) on the macOS image?

willsmythe:
@minrk - the D: drive (where the source gets cloned) is provided 10 GB of storage. Not sure why I didn't notice this earlier, but the "Release" folder produced during the build can easily grow to 8 or 9 GB (from my limited testing) and if other files are being created/dropped into the repo folder, it would explain the out of disk space error on D.

I'm playing around with generating the Release folder elsewhere (e.g. C: which has more available space, but is slower than D:).

minrk:
Thanks for checking. I’ve run builds on both C and D and both have similar amounts of free space and run out before it completes. It seems weird to me that the build for would get so big!

willsmythe:
@minrk - I agree. The "problem" is that the Node.js Windows build (`vcbuild.bat`) creates the `Release` folder (and intermediate folders) at the root of the cloned repo, and doesn't expose an argument to change this. I tweaked vcbuild.bat directly in my local version and set the `OutDir` property:

```
msbuild node.sln %msbcpu% /t:%target% /p:OutDir=c:\rel;Configuration=%config%  ...
                                         ^ here
```

This built correctly (although it took over an hour -- more of this in a minute).

**What I think you can do**...

Copy the repo source (build variable `$Build.SourcesDirectory`) to C: and then run `vcbuild.bat` as usual. `bld.bat` may be a reasonable place to do this, otherwise a custom Windows-only step in azure-pipelines.yml.

Re: build time ... D: is essentially an SSD drive, and is where you generally want to do any intensive disk operations (like building this repo), but, as we have determined, it's not an option for this particular repo so you have to use C:. The Node.js repo built in 1 hour and 1 minute on the VS2017 image even when I ran msbuild with `OurDir` set to `c:\rel`. I also had to add the following attribute to the Windows `Job` in my `azure-pipelines.yml`:

```
timeoutInMinutes: 240
```

(Azure Pipelines has a 6 hour limit on open source build jobs, but the default (if not specified in the .yml) is only 60 minutes)

@mariusvniekerk - you might want to consider having conda-smithy set this attribute in azure-pipelines.yml or think about a way for feedstock authors to override it (in general or on a per-platform basis). https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?tabs=yaml&view=vsts#timeouts

Let me know if that doesn't work / doesn't make sense.

requirements.txt
scopatz: I think this is used for conda, not pip, though.

scopatz:
Hi All, I am going to merge this since I know @mariusvniekerk is slammed this week so that we can move forward in other PRs. 

Thanks for the discussion on node @minrk! 

@DavidStaheli @willsmythe - I don't seem to be able to incorporate your suggested changed from the GitHub UI.  Would you mind please putting in other PRs? Thanks!

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
scopatz: If miniconda is already available, this could be changed to a `conda update --all`. We just need the latest version

isuruf:
@scopatz, can you at least open an issue for my concern on template duplication?

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
```diff
@@ -0,0 +1,83 @@
+jobs:
+- job: osx_64
+  pool:
+    vmImage: xcode9-macos10.13
```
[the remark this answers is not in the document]
DavidStaheli: Here's a list of its contents.  Please let me know if anything is missing that you need.
[Azure Pipelines hosted macOS image](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-Readme.md)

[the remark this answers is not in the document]
isuruf: We currently build for SDK 10.9 and allow it to be changed per feedstock. (Need 10.10 for some)

scopatz:
@isuruf - I am not sure what you wanted here exactly.  It seems like each OS/CI platform might need their own templates fundamentally.

isuruf:
@scopatz, I meant the run_docker_build.sh file. With this PR, there is basically two copies of the same file in two places in the same feedstock. (Only difference being that the azure file disables uploading. This can be avoided by using a env variable)

scopatz:
I guess we can. I'll let @mariusvniekerk jump in with the justification for why not, but I presume it is because these files may diverge significantly.

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
jakirkham: Think there may have been some concerns with the location of the install. Can't remember if it was specific to the the macOS image or a different image. Hopefully @mariusvniekerk remembers better than I.

requirements.txt
jakirkham: Yeah [it's used for CI]( https://github.com/conda-forge/conda-smithy/blob/06722d9553e73fd0610ceb443117d8d1bfda0805/.travis.yml#L26 ). Though can see this is a bit confusing. Opened issue ( https://github.com/conda-forge/conda-smithy/issues/920 ) about dropping it.

conda_smithy/templates/run_docker_build.sh.tmpl
```diff
@@ -36,15 +39,15 @@ mkdir -p "$ARTIFACTS"
 DONE_CANARY="$ARTIFACTS/conda-forge-build-done-${CONFIG}"
 rm -f "$DONE_CANARY"
 
-{{ docker.executable }} run -it \
+{{ docker.executable }} run \
```
jakirkham: Was `-it` dropped intentionally for some reason or could it be kept?

conda_smithy/templates/azure-pipelines-win.yml.tmpl
jakirkham: That's sadly a typo that has been copied around for a while. Fixing with PR ( https://github.com/conda-forge/conda-smithy/pull/921 ).

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
jakirkham: Ah so this actually does something different. Though we probably need to fix it as well. We have [our own fast finish script for CIs]( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/master/recipe/ff_ci_pr_build.py ) (particularly if the provider doesn't support it or the provider implementation fails).

That said, it looks like Marius has made this an empty string for now. So it isn't running the script at all, which is fine to start. We can update this once the CI fast finish script handles the Azure case. Raised as issue ( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/32 ).

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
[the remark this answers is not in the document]
jakirkham: Would it be possible to include 10.9 and 10.10 on the image as well?

conda_smithy/azure_ci_utils.py
```diff
@@ -0,0 +1,191 @@
+import os
+import typing
+
+from vsts.vss_connection import VssConnection
+from msrest.authentication import BasicAuthentication
+
+
+AZURE_TEAM_INSTANCE = os.getenv("AZURE_INSTANCE", "https://dev.azure.com/conda-forge")
+AZURE_PROJECT_ID = os.getenv("AZURE_PROJECT_ID", "feedstock-builds")
+AZURE_SERVICE_ENDPOINT_NAME = os.getenv("AZURE_SERVICE_ENDPOINT", "conda-forge")
```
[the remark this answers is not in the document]
jakirkham: Raised as issue ( https://github.com/conda-forge/conda-smithy/issues/922 ).

conda_smithy/azure_ci_utils.py
```diff
@@ -0,0 +1,191 @@
...
+
+try:
+    # Create a token at https://circleci.com/account/api. Put it in circle.token
```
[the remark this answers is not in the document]
jakirkham: The comment seemed unnecessary since the error message already explains this (and does a better job). So just added PR ( https://github.com/conda-forge/conda-smithy/pull/923 ) to drop this comment (and another one just like it).

conda_smithy/azure_ci_utils.py
jakirkham: Just checking, was this resolved and/or is there an issue/PR we should xref to?

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
```diff
@@ -0,0 +1,83 @@
...
+    {%- endfor %}
+
+  steps:
+  # TODO: Fast finish on azure pipelines?
+  - script: |
+      echo "Fast Finish"
+      {{ fast_finish }}
+
+  - script: |
+      echo "Removing homebrew from Travis CI to avoid conflicts."
```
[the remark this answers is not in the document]
jakirkham: Addressing in PR ( https://github.com/conda-forge/conda-smithy/pull/924 ).

jakirkham:
Have merged a bunch of nit PRs to fix small typos and the like quickly. I don't think any of them should have cause for concern. However if they do, happy to discuss resolving them.

Also have taken the liberty of breaking the duplicate CI file concern out into issue ( https://github.com/conda-forge/conda-smithy/issues/926 ). Hopefully that will make it easier to have a fruitful discussion about addressing this issue (given the PR has been merged). Please feel free to add any other relevant details there that I've neglected in my executive summary. Thanks.

jakirkham:
Thanks @mariusvniekerk for all the work you put into this. It's a nice step forward. Appreciate all the testing you did on it as well.

Also thanks @DavidStaheli and @willsmythe for helping out with the reviewing, answering questions, and fielding user reports. This makes it much easier to make progress on this sort of thing generally.

Finally thanks everyone else who helped test and review this PR. With something of this size that is much needed work and is greatly appreciated. Imagine there will be follow-up work to adjust this is as we move forward and it will be good to hear your feedback as we progress.

Thanks again everyone. Excited to start making use of Azure in conda-forge! 😄 🎉

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
mariusvniekerk: When we did this last time Miniconda was owned by something else and we couldn't alter the base environment correctly

PR #902, opened 2018-09-28
add numfocus badge

ocefpaf:
It would look like this:

screenshot from 2018-09-28 12-24-19

@scopatz what do you think?


scopatz:
Seems fine for now.  I think it could go lower down in the readme, but I don't think it is worth getting into that level of copy-editing.  Folks should feel free to switch this up!

PR #966, opened 2018-12-13
Only fail noarch lint if selectors in runtime dependencies.

jdblischak:
Checklist
* [x] Added a ``news`` entry

PR #622 didn't include a news entry. Since this PR is similar in scope, I assume it also does not warrant a news entry. Please let me know otherwise.

This PR address https://github.com/conda-forge/conda-forge.github.io/issues/683, in which the linter is failing recipes that have selectors in the build requirements. staged-recipes builds noarch packages on all 3 platforms purposefully, in case others want to build the recipe on other platforms.

cc: @CJ-Wright @isuruf @bgruening

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -273,7 +273,7 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
             in_runreqs = False
             for line in fh:
                 line_s = line.strip()
-                if (line_s == "run:"):
+                if (line_s == "host:" or line_s == "run:"):
                     in_runreqs = True
                     runreqs_spacing = line[:-len(line.lstrip())]
                     continue
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -267,24 +267,24 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
                                      ' name.'.format(section, source_subsection))
 
 
-    # 17: noarch doesn't work with selectors
+    # 17: noarch doesn't work with selectors for runtime dependencies
     if build_section.get('noarch') is not None and os.path.exists(meta_fname):
         with io.open(meta_fname, 'rt') as fh:
-            in_requirements = False
+            in_runreqs = False
             for line in fh:
                 line_s = line.strip()
-                if (line_s == "requirements:"):
-                    in_requirements = True
-                    requirements_spacing = line[:-len(line.lstrip())]
+                if (line_s == "run:"):
+                    in_runreqs = True
+                    runreqs_spacing = line[:-len(line.lstrip())]
                     continue
                 if line_s.startswith("skip:") and is_selector_line(line):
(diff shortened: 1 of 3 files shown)


CJ-Wright:
This needs news (that PR is from before news was a thing).

jdblischak:
Also noting that an alternative to changing the linter is to skip Travis and AppVeyor builds on staged-recipes, a WIP PR is already open at https://github.com/conda-forge/staged-recipes/pull/7239

jakirkham:
Are there other things we are accidentally allowing by relaxing this (e.g. selectors in source)?

jdblischak:
> Are there other things we are accidentally allowing by relaxing this (e.g. selectors in source)?

@jakirkham That's already happened. PR #622 limited the linting of selectors in noarch recipes to the requirements section. This PR takes that further and only lints the subsection `run:`.

jakirkham:
Sure I understand relaxing the other sections. Perhaps “revisit” was the wrong word to use here. My only point is relaxing this for some other sections like `source` appears to be unintentional and may actually cause packages that should be the same across platforms to be different.

bgruening:
I think that's a valid approach. @jakirkham do you still have concerns? Currently, most of the r-* packages in staged-recipes are failing due to this.

jakirkham:
Mostly ok with this. Am concerned about allowing selectors into `host`. There could be some subtle cases with `run_exports` that would sneak by this change. Can't really see any benefit to allowing selectors in `host` either. Do others?

jdblischak:
> Am concerned about allowing selectors into host. There could be some subtle cases with run_exports that would sneak by this change. Can't really see any benefit to allowing selectors in host either.

@jakirkham I could try to update the linter to check for selectors in both `host` and `run` dependencies. The current approach is rather fragile, but it'd work as long as the `build` dependencies were always specified first (which should be almost always true since it is the convention for recipes).

scopatz:
Yeah the linter will fail if build is after host. I am happy with this too.  I will merge tomorrow if there are no further objections.

scopatz:
> @scopatz If I understand correctly, you're saying the linter will fail independent of this PR if the requirements are not in the conventional order. Thanks for clarifying.

Yes that is correct!

Adding the change would allay one of @isuruf's concerns, so it is probably a good idea to add it.

jdblischak:
@jakirkham This PR now fails the lint if there are selectors used for either `host` or `run` dependencies. Please let me know if there are additional scenarios you would like me to add to the test suite.

jakirkham:
Thanks @jdblischak :)

> This PR now fails the lint if there are selectors used for either `host` or `run` dependencies. Please let me know if there are additional scenarios you would like me to add to the test suite.

That seems sufficient. Thanks.

> What is the potential problem with having selectors in `source:` and other sections?

With the `source` section specifically, I've raised issue ( https://github.com/conda-forge/conda-smithy/issues/982 ). Though as we discussed this is unrelated to this specific PR (as that change preceded it).

PR #974, opened 2018-12-21
Emit conda-build version in commit message

duncanmmacleod:
This PR adds the `conda-build` version to the commit message generated by `conda smithy rerender`. Ideally this will make tracking down feedstock rendering problems easier.

Checklist
* [ ] Added a ``news`` entry


duncanmmacleod:
I'm not quite sure where to put the `news` entry, is that in `CHANGELOG.rst`?

mariusvniekerk:
in the news directory there is a template.  Make a copy of that and name it something,  Add some text

scopatz:
Basically the same as this: https://xon.sh/devguide.html#changelog

jakirkham:
This seems like a good idea. We might want to adjust the format given the long commit subject line will be wrapped by GitHub. This is a minor point though.

PR #979, opened 2018-12-29
Drop old CHANGELOG file linking to GitHub releases

jakirkham:
As we now use rever to cut releases and that makes a `CHANGELOG.rst` in the top-level directory, there is no need to point users to GitHub releases using this `CHANGELOG.md` as well. So go ahead and drop it for simplicity.

Checklist
* [ ] Added a ``news`` entry


jakirkham:
Thoughts?

scopatz:
LGTM!

PR #1006, opened 2019-01-26
minor azure tweaks

mariusvniekerk:
just removed some --quiet flags


mariusvniekerk:
So tiny that it really does not need news

PR #1007, opened 2019-01-26
Fixed missing secret for azure publishing

mariusvniekerk:
For azure secrets aren't added automatically to all stages, you have to ask to use them

See https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azdevops&tabs=yaml%2Cbatch#systemaccesstoken


scopatz:
Thanks!

PR #1041, opened 2019-03-15
fix broken links to conda-forge docs

ltalirz:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/feedstock_content/.github/PULL_REQUEST_TEMPLATE.md
+++ b/conda_smithy/feedstock_content/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@ Checklist
 * [ ] Used a fork of the feedstock to propose changes
 * [ ] Bumped the build number (if the version is unchanged)
 * [ ] Reset the build number to `0` (if the version changed)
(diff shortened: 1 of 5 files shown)


ltalirz:
happy to add a news entry, if really desired

scopatz:
This is basically all wrong and should be reverted. We don't use read the docs

croth1:
This confusion was created by having two doc versions in parallel. The issue is already reported here: https://github.com/conda-forge/conda-forge.github.io/issues/735. Unfortunately, @jakirkham seems to be very busy at the moment, so we couldn't take the other version down yet. Sorry for the confusion.

ltalirz:
Sorry about this - I just came across a broken doc link suggested to me by the linter bot on conda-forge and then thought you had moved to rtd.

Take the first link changed in the PR, for example:
This original link is still broken: https://conda-forge.org/docs/conda_smithy.html#how-to-re-render
This one on rtd works: https://conda-forge.readthedocs.io/en/latest/conda_smithy.html#how-to-re-render

I.e. reverting will not really make things better. If I understand correctly, the problem is that the documentation on conda-forge.org was updated while the one on RTD is still old and therefore the links still exist (?)

Should I work on a separate PR to try moving links back to the docs on conda-forge.org?

scopatz:
See #1048, which should fix all of the links themselves too

PR #1077, opened 2019-05-21
Clean up azure table whitespace in README

bollwyvl:
This cleans out some extra whitespace that trigger a code block in the azure table:
Screenshot from 2019-05-20 20-05-56

I haven't _fully_ tested this... but ran into it while looking at other stuff...


scopatz:
Thanks! Can you please add a news entry, see https://regro.github.io/rever-docs/news.html

scopatz:
Thanks!

bollwyvl:
Yay!

PR #1078, opened 2019-05-21
[wip] publish test reports

### LINKED ISSUE #1076: Opt-in to capturing coverage and test reports?
So as a maintainer, despite the heartache getting it working, building test dependencies, burning coal, etc., running the unit tests during the build gives me a warm fuzzy feeling when i press merge. When they _do_ fail, it's always a bit of a hunt to find the fail... `exited with code 1` is... opaque.

A number of test runners can generate the xunit files needed to generate really nice, targeted error reports that bubble all the way up to github.... while you'd still need to look at the logs, at least you'd have something to <kbd>ctrl+f</kbd> to get started.

It would be great to be able to signal in `extras` that after a build that there is structured data someplace which _could_ be submitted, and then have ci-vendor specific approaches to uploading done in a consistent fashion.

While at it, coverage would be cool, too: while the absolute coverage would be pretty much meaningless, a rapid decline in coverage could point to some new dependency that isn't being tested, etc.

So, for a python-kinda thing, something like:
```yaml
test:
  requires: [pytest-cov]
  commands: pytest --cov-report xml:cov.xml --junitxml junit.xml
extras:
  coverage:
    - cov.xml
  junit:
    - junit.xml
```

no doubt this would be more complicated for more bespoke `make check`s, but you gotta start somewhere!

I can't promise that I can work on this immediately, but would happily hazard a PR if it wasn't outright, _this isn't conda-forge's job_ed, but it does seem like another interesting way to increase the value of the coal we're already burning.

once this could, at the very least, be captured in each ci vendor's ui, and maybe at the github ui, investigating third-party providers would be pretty neat to provide an even broader look at the health of the ecosystem.

bollwyvl:
**Checklist**
* [ ] Added a ``news`` entry

**Issues**
- fixes #1076 

**Why**
The hope is this will become something that can capture structured data test reports, increasing the maintainer value of running actual tests. It seems the de facto format is the `xunit`/`junit` ~~standard~~, supported by at least circle, azure and appveyor.

Python-town has about ~200 packages that are presently running tests with `pytest` or `nosetests` and a smattering from other language groups. For `make check`-style tests, the situation is less clear, though there appear to be various _x_`2junit` utilities... who knows.

**Implementation sketch**
- [ ] decide where to put report info, thinking in `meta.yaml`:
```yaml
extras:
  test_reports:
    - *.xunit.xml   # relative to test dir?
```
- [ ] decide where to render report info (still learning)
- [x] run `conda-build` with `--keep-old-work`
- [ ] find `test_reports` and `coverage_reports` after build
- [ ] use ci-specific tools to upload test results, tagged with config
  - [ ] [azure](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml#publishtestindocker)
  - [ ] [circle](https://circleci.com/docs/2.0/collect-test-data/)
  - [ ] [appveyor](https://www.appveyor.com/docs/running-tests/#uploading-xml-test-results)
- [ ] render test badges to README

Thinking on this more, I probably won't do coverage on this first go-around, as its value is a bit dubious, and often hard to reconcile between multiple platforms... something like codecov or coveralls would be better, but even that might not be worth the work.

I can't promise I'll be able to drive this home quickly, as I'm mostly nights-and-weekends of late. Happy or any insights/discussion!

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1144,6 +1144,8 @@ def _load_forge_config(forge_dir, exclusive_config_file):
             "branch_name": "master",
         },
         "recipe_dir": "recipe",
+        "test_reports": None,
+        "coverage_report": None
     }
 
     # An older conda-smithy used to have some files which should no longer exist,
--- a/conda_smithy/feedstock_content/build-locally.py
+++ b/conda_smithy/feedstock_content/build-locally.py
@@ -12,6 +12,8 @@ from argparse import ArgumentParser
 def setup_environment(ns):
     os.environ["CONFIG"] = ns.config
     os.environ["UPLOAD_PACKAGES"] = "False"
+    os.environ["UPLOAD_TEST_REPORTS"] = "False"
+    os.environ["UPLOAD_COVERAGE_REPORT"] = "False"
 
 
 def run_docker_build(ns):
--- a/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
@@ -15,6 +15,8 @@ jobs:
       {{ config_name }}:
         CONFIG: {{ config_name }}
         UPLOAD_PACKAGES: {{ upload }}
+        UPLOAD_TEST_REPORTS: "False"
+        UPLOAD_COVERAGE_REPORT: "False"
         DOCKER_IMAGE: {{ config["docker_image"][-1] }}
     {%- endif %}
     {%- endfor %}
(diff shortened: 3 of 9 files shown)


scopatz:
Thanks so much for putting this in! 

I think these might be better suited to configuration variables (ie set in the `conda-forge.yaml` file) rather than as environment variables.  See https://conda-forge.org/docs/maintainer/conda_forge_yml.html and https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/configure_feedstock.py#L1092

scopatz:
I know it is more work do to that, but that is the standard place for configuration

PR #1091, opened 2019-06-25
bump conda version check (from 4.7 to 5.0)

msarahan:
Checklist
* [x] Added a ``news`` entry

I did a successful rerender with conda 4.7.5.  I don't think this needs to be constrained so tightly.  We'll be much more conservative about semver under my watch, so the next major version should be appropriate here.

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -472,7 +472,7 @@ def main():
         args = parser.parse_args()
 
     # Check conda version for compatibility
-    CONDA_VERSION_MAX = "4.7"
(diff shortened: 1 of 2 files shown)


jakirkham:
Yeah I think we also migrated to using conda-build's API to mitigate the impact that changes in conda might have. IOW this seems like a good idea to me. Thanks for submitting this fix!

jakirkham:
Proposing we release with this change in issue ( https://github.com/conda-forge/conda-smithy/issues/1092 ).

PR #1093, opened 2019-06-26
build-locally.py: Use Python 3 if multiple versions are available

henryiii:
This requires Python 3.6+, so it should list python3 as the runner.

Note that [PEP 394](https://www.python.org/dev/peps/pep-0394/) still recommends distributions provide `python -> python2`, and distributions like RedHat 8 are not providing an unversioned `python` command at all.


scopatz:
Can you please provide a news entry: https://regro.github.io/rever-docs/news.html

henryiii:
@scopatz I added an entry.

news/henryiii-build-locally-python3.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* `build-locally.py` now uses `python3` even if `python` is `python2` (Python 3.6+ was already required)
```
scopatz: This is rat, not me, so you need to use double backticks :(

henryiii: Ah, I see the file extension. Will correct tomorrow.

scopatz: Stupid autocorrect keyboard 😡

scopatz:
Thanks!

PR #1094, opened 2019-06-27
Use `shyaml -h` for `shyaml` detection to support newer versions

xhochy:
Checklist
* [x] Added a ``news`` entry


xhochy:
Is this worthy a news entry?

scopatz:
Yeah, probably since it does change how CI is running.  Also it is a bit strange on its own since there is no justification in the code (and shouldn't be)

scopatz:
Thanks @xhochy!

PR #1142, opened 2019-09-06
fix render_README crash when azure api returns 404

ap--:
Checklist
* [x] Added a ``news`` entry

If `user_or_org` is set and no azure account is set up for the specified name, the request returns a 404 response, which leads to an uncaught `JSONDecodeError`.

This fix raises an `HTTPError` which is caught by the except clause's `IOError`.


scopatz:
Thanks!

PR #1208, opened 2019-12-07
Add notes pertaining to branding in the README.md

hmaarrfk:
Checklist
* [ ] Added a ``news`` entry


scopatz:
Thanks! It also seems like these should be in the docs too

PR #1222, opened 2020-01-27
Allow conda-forge.yml to be empty

saraedum:
Otherwise, you'd get a `IndexError: list index out of range`.


scopatz:
Thanks!

scopatz:
Can you please include a test for this case?

saraedum:
It's not obvious to me how to do this. Are there any tests that take an entire repository tree as input?

scopatz:
I thought there were, but you could also just test that function

saraedum:
@scopatz, I added tests now. (Took a bit longer than I thought…)

PR #1225, opened 2020-01-29
Unset Azure Boost_ROOT

jschueller:
Azure felt it was a good idea to set Boost_ROOT env var to their preinstalled image, which can interfere with our cmake builds.
https://developercommunity.visualstudio.com/idea/556787/stop-setting-boost-root-as-predefined-variable.html

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/templates/azure-pipelines-win.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-win.yml.tmpl
@@ -65,8 +65,10 @@ jobs:
         updateConda: false
       displayName: Install conda-build and activate environment
 
-    - script: set PYTHONUNBUFFERED=1
+    - script: set "Boost_ROOT="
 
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
You need to use some of the weirder ways to set environment variables.

See here for how to https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-variables-in-scripts

jschueller:
why ? it works for PYTHONUNBUFFERED too doesnt it ?

scopatz:
Also, please add a news entry

jschueller:
@scopatz ok, done

mariusvniekerk:
The PYTHONUNBUFFERED there actually does nothing. The export statement does not persist past running that step.

jakirkham:
Where should this go then?

PR #1228, opened 2020-02-04
minor typo in rerender error message

minrk:
message said "not rerendering inside the recipe" when rerender *was* called from inside the recipe.


scopatz:
Thanks @minrk!

PR #1247, opened 2020-03-09
ENH CFEP-13 implementation #1

beckermr:
This is the first of ~3 PRs for CFEP-13. This PR has the code to hash and register feedstock tokens in the right spots. Future PRs will have the output validation/registration code and CLI utils to rotate the binstar tokens.

Checklist
* [x] Added a ``news`` entry
* [x] add CLI tool

### DIFF
--- a/conda_smithy/feedstock_tokens.py
+++ b/conda_smithy/feedstock_tokens.py
@@ -267,7 +267,6 @@ def register_feedstock_token(user, project, token_repo):
 def register_feedstock_token_with_proviers(
     user,
     project,
-    feedstock_directory,
     drone=True,
     circle=True,
     travis=True,
--- a/conda_smithy/feedstock_tokens.py
+++ b/conda_smithy/feedstock_tokens.py
@@ -136,31 +136,28 @@ def feedstock_token_exists(user, project, token_repo):
     exists = False
     failed = False
     err_msg = None
-    with tempfile.TemporaryDirectory() as tmpdir:
-        with open(os.devnull, "w") as fp, redirect_stdout(fp), redirect_stderr(
-            fp
-        ):
-            try:
-                # clone the repo
-                _token_repo = (
-                    token_repo.replace("$GITHUB_TOKEN", github_token)
-                    .replace("${GITHUB_TOKEN}", github_token)
-                    .replace("$GH_TOKEN", github_token)
-                    .replace("${GH_TOKEN}", github_token)
-                )
-                git.Repo.clone_from(_token_repo, tmpdir, depth=1)
-                token_file = os.path.join(
-                    tmpdir,
-                    "tokens",
-                    project.replace("-feedstock", "") + ".json",
-                )
+    with tempfile.TemporaryDirectory() as tmpdir, open(
+        os.devnull, "w"
+    ) as fp, redirect_stdout(fp), redirect_stderr(fp):
+        try:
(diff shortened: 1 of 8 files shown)


beckermr:
@conda-forge/core This one is ready for review!

conda_smithy/feedstock_tokens.py
```diff
@@ -0,0 +1,569 @@
+import tempfile
```
scopatz: Please add a docstring to this module

conda_smithy/feedstock_tokens.py
```diff
@@ -0,0 +1,569 @@
...
+        ~/.conda-smithy/{user or org}_{repo w/o '-feedstock'}_feedstock.token
+
+    This function will fail if the token file already exists.
+    """
+    # capture stdout, stderr and suppress all exceptions so we don't
+    # spill tokens
+    failed = False
+    err_msg = None
+    with open(os.devnull, "w") as fp:
+        with redirect_stdout(fp), redirect_stderr(fp):
```
scopatz: This can be one level of indentation: `with open(os.devnull, "w") as fp, redirect_stdout(fp), redirect_stderr(fp)` I think

conda_smithy/feedstock_tokens.py
beckermr: done!

conda_smithy/feedstock_tokens.py
beckermr: Indeed it can!

beckermr: done

conda_smithy/feedstock_tokens.py
```diff
@@ -0,0 +1,586 @@
...
+    your environment before calling this function.
+    """
+    from .github import gh_token
+
+    github_token = gh_token()
+
+    exists = False
+    failed = False
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with open(os.devnull, "w") as fp, redirect_stdout(fp), redirect_stderr(
```
scopatz: Actually this can be combined with the above as well too, if we want to reduce the indentation even more

beckermr: indeed it can. I'll push a change later

scopatz:
Thanks @beckermr - can you please add some tests for feedstock_tokens?

beckermr:
Those are all API calls so tests are really hard. What did you have in mind?

scopatz:
Well, it doesn't have be realistic, but you could mock the API calls, or otherwise test that files get written in some approximation of correct

beckermr:
Those kinds of tests are a tautology. If you really want to see them I happy to oblige, but here they don't actually add much. There are cases where it can, especially if you have a lot of logic around the results, but that is not the case here.

beckermr:
I will add tests for the files for sure.

scopatz:
I think that they do help test the logic, so I am not sure that they are a tautology, even if they aren't live. I also don't think, for maingtainability, we should add a module without any tests. Just my opinion :wink:

beckermr:
OK @scopatz I added a pretty complete set of unit tests for most of the functions that hits all of the high-level logic. Ready to merge?

scopatz:
Looks great!

scopatz:
Thanks!

PR #1290, opened 2020-04-22
Add check for existing feedstocks with underscore instead of dash and vv

npavlovikj:
Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -679,7 +679,11 @@ def run_conda_forge_specific(meta, recipe_dir, lints, hints):
     if is_staged_recipes and recipe_name:
         cf = gh.get_user(os.getenv("GH_ORG", "conda-forge"))
         try:
-            cf.get_repo("{}-feedstock".format(recipe_name))
+            cf.get_repo("{}-feedstock".format(recipe_name)) or cf.get_repo(
+                "{}-feedstock".format(recipe_name.replace("-", "_"))
+            ) or cf.get_repo(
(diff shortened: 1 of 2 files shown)


npavlovikj:
@beckermr , I hope it is okay that I am tagging you here since you were part of the conversation on Gitter from few days ago where the package `custom_inherit` was merged although `custom-inherit` already exists in CF. This PR checks for existing feedstocks with the same name, as well as feedstocks that may differ in "-"/"_". I would appreciate any feedback and review on this PR!

conda_smithy/lint_recipe.py
```diff
@@ -679,13 +679,28 @@ def run_conda_forge_specific(meta, recipe_dir, lints, hints):
...
+            else:
+                feedstock_exists = False
         except github.UnknownObjectException as e:
             feedstock_exists = False
 
         if feedstock_exists:
-            lints.append("Feedstock with the same name exists in conda-forge")
+            lints.append(
+                'Feedstock with the name "{}" exists in conda-forge'.format(
+                    existing_recipe_name
```
beckermr, line 701: This is better, but can we be more specific. If the name doesn't match, the message should say something to the effect of "We found a feedstock with the name {}. It doesn't match yours but we suspect it may be the same package." In this case it should be a linting hint.

If the name matches, we can keep the current message and keep it as a lint.

beckermr:
@conda-forge/core Do you all have any thoughts here?

npavlovikj:
@beckermr and @isuruf , thank you so much for your help and suggestions with this PR!

PR #1303, opened 2020-04-29
ENH use read-write permissions for docker mounts

beckermr:
This fixes a pip install issue @isuruf found when making a change to the conda-forge-ci-setup feedstock. 

https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/87/files

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/templates/run_docker_build.sh.tmpl
+++ b/conda_smithy/templates/run_docker_build.sh.tmpl
@@ -60,7 +60,7 @@ fi
 
 export UPLOAD_PACKAGES="${UPLOAD_PACKAGES:-True}"
 {{ docker.executable }} run ${DOCKER_RUN_ARGS} \
-           -v "${RECIPE_ROOT}":/home/conda/recipe_root:ro,z \
+           -v "${RECIPE_ROOT}":/home/conda/recipe_root:rw,z \
            -v "${FEEDSTOCK_ROOT}":/home/conda/feedstock_root:rw,z \
            -e CONFIG \
            -e BINSTAR_TOKEN \
--- a/news/docker_rw.rst
+++ b/news/docker_rw.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Changed the docker mount to the recipe directory to have read-write permissions instead
+  of read-only.
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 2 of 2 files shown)


beckermr:
Testing this here: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/88

beckermr:
Test appears to be ok.

jakirkham:
I believe this was added in PR ( https://github.com/conda-forge/conda-smithy/pull/888 ). @pkgw @bollwyvl, would this work for the SELinux use case or would this cause you issues?

beckermr:
Thanks for tracking this down!

Looks like the text in the pr says “shouldn’t need to edit it”. It turns out we are now pip installing a package in the recipe dir so the container does need write permissions. The main thing in that pr appears to be the z flag which we have kept.

bollwyvl:
Alas, my fedora box where this was an issue has passed on, and I do
basically all of my docker stuff in VMs these days, so can't easily
confirm... I'll see if I can put together a more locked-down test...

pkgw:
Yeah, @beckermr is right — for SELinux the `z` flag is the important bit. No worries about changing to read-write if that's what the build needs.

(To editorialize, I think it would be better if builds could work with the recipe directory mounted read-only, but if there's a problem that needs fixing, that's more important than purism.)

jakirkham:
Thanks all! Let's ship it then 😄

conda_smithy/templates/run_docker_build.sh.tmpl
```diff
@@ -60,7 +60,7 @@ fi
 
 export UPLOAD_PACKAGES="${UPLOAD_PACKAGES:-True}"
 {{ docker.executable }} run ${DOCKER_RUN_ARGS} \
-           -v "${RECIPE_ROOT}":/home/conda/recipe_root:ro,z \
+           -v "${RECIPE_ROOT}":/home/conda/recipe_root:rw,z \
```
isuruf, line 63: If you are changing this, can you get rid of `RECIPE_ROOT` as well? I don't see any reason to have two mounts here if the same permission is used.

beckermr: This variable is used a lot in the underlying code so changing it will take some work. Let's make an issue and come back.

PR #1319, opened 2020-05-17
BUG github team names are case-insensitive

beckermr:
So github team names do not appear to be case sensitive. I have also seen errors when the team name has uppercase values when we are making codeowner files. See @xhochy's PR on the pinnings repo. Thus we always make team names lower case.

This link

https://github.com/orgs/conda-forge/teams/Core

and this one 

https://github.com/orgs/conda-forge/teams/core

go to the same spot for example.

Checklist
* [x] Added a ``news`` entry


beckermr:
cc @conda-forge/core this one is ready for review

PR #1326, opened 2020-05-25
Use the anaconda API for version checks

xhochy:
Closes #1325 

Checklist
* [x] Added a ``news`` entry


xhochy:
Same performance improvement as #1325 , also trimming the runtime by 20s.

Issue #1333, opened 2020-05-27
Upload with `--private` on anaconda


hadim:
The idea is to be able to specify to upload using a `--private` flag.

Here is an associated issue https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/97 and the associated PR: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/98

On the `conda-smithy` side, we need to somehow propagate a `private_upload` key in `./ci_support/*.yaml`.

Two questions:

- are you guys ok with this feature?
- does that make sense to add it to `conda-forge.yml`? if it does with what key? and how can I propagate that to `./ci_support/*.yaml`?

As soon as I have your green light I'll make a PR.

beckermr:
Feature seems fine. There is a large dictionary in the configure_feedstock.py file. Add a default of False for the flag for your feature there. Then edit the ci script templates to add the flag if that key is set. 

Note that right now you are doing this by putting the flag in. ci support scripts. It should not be in there.

hadim:
Let's continue there: https://github.com/conda-forge/conda-smithy/pull/1334

hadim:
Fixed by #1334

PR #1334, opened 2020-05-27
Private upload

hadim:
Following https://github.com/conda-forge/conda-smithy/issues/1333

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1454,6 +1454,7 @@ def _load_forge_config(forge_dir, exclusive_config_file):
         "skip_render": [],
         "bot": {"automerge": False},
         "conda_forge_output_validation": False,
+        "private_upload": False
     }
 
     forge_yml = os.path.join(forge_dir, "conda-forge.yml")
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -376,6 +376,11 @@ def finalize_config(config, platform, forge_config):
             config["docker_image"] = [config["docker_image"][0]]
         else:
             config["docker_image"] = [forge_config["docker"]["fallback_image"]]
+
+    if "private_upload" in config:
+        config["private_upload"] = config["private_upload"]
+    elif "private_upload" in forge_config:
+        config["private_upload"] = forge_config["private_upload"]
     return config
 
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -376,11 +376,6 @@ def finalize_config(config, platform, forge_config):
             config["docker_image"] = [config["docker_image"][0]]
         else:
             config["docker_image"] = [forge_config["docker"]["fallback_image"]]
-
-    if "private_upload" in config:
-        config["private_upload"] = config["private_upload"]
-    elif "private_upload" in forge_config:
-        config["private_upload"] = forge_config["private_upload"]
     return config
 
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1454,7 +1454,7 @@ def _load_forge_config(forge_dir, exclusive_config_file):
         "skip_render": [],
         "bot": {"automerge": False},
         "conda_forge_output_validation": False,
-        "private_upload": False
+        "private_upload": False,
     }
 
     forge_yml = os.path.join(forge_dir, "conda-forge.yml")
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -69,4 +69,4 @@ deploy_script:
 {%- if upload_on_branch %}
     - set "UPLOAD_ON_BRANCH={{ upload_on_branch }}"
 {%- endif %}
-    - cmd: upload_package {% if conda_forge_output_validation %}--validate --feedstock-name="{{ feedstock_name }}"{% endif %} .\ .\{{ recipe_dir }} .ci_support\%CONFIG%.yaml
+    - cmd: upload_package {% if conda_forge_output_validation %}--validate {% if conda_forge_private_upload %}--private --feedstock-name="{{ feedstock_name }}"{% endif %} .\ .\{{ recipe_dir }} .ci_support\%CONFIG%.yaml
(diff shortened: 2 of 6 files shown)


hadim:
@beckermr you suggest tp modify ci templates but I don't see any template for `ci_support/` files. That's why I did it directly in the code.

hadim:
This PR also depends on this PR: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/98

hadim:
ping @beckermr @jakirkham

conda_smithy/configure_feedstock.py
```diff
@@ -376,6 +376,11 @@ def finalize_config(config, platform, forge_config):
             config["docker_image"] = [config["docker_image"][0]]
         else:
             config["docker_image"] = [forge_config["docker"]["fallback_image"]]
+
+    if "private_upload" in config:
+        config["private_upload"] = config["private_upload"]
+    elif "private_upload" in forge_config:
+        config["private_upload"] = forge_config["private_upload"]
```
beckermr, line 383: I am confused by this. Why are we stashing private upload in the ci support scripts?

We should leave this in the conda-forge,yml and use the value from there.

hadim: I was trying to do the same as for `docker_image` but I got your point.

beckermr: Hmmmm. I did not add the docker_image stuff. @conda-forge/core why is docker image in the ci support data? Is that because it varies per platform?

hadim: Yes according to the `cuda` flag I think.

hadim: It should be solved now.

beckermr: Ok. My point here si that we don't need to set it in the config at all. The conda-forge-ci-support package could/should look directly at the conda-forge.yml. The ci support files are for stuff that ends up being used by conda build for building packages.

hadim: Oh ok, I get your point. Then I need some guidance for how to do that in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/98/files

hadim: I tried to find the logic in `conda-forge-ci-support` that gets and uses the `docker_image` flag but cannot find it. The only place where the `docker_image` is used is the `conda-smithy` file.

beckermr: Right. Here is a variation that is slightly better than what I just said.

1. put the flag in the conda-forge.yml
2. use that flag value to render recipes with a `--private` flag passed to the ci support scripts or w/e that flag should be
3. handle the `--private` flag in the ci support scripts.

hadim: Sorry but it looks very similar to what I am doing right now... `config["private_upload"]` is used to populate `ci_support` script. Then the PR in ` conda-forge-ci-support` handles the flag if present.

Could you explain in more detail what you want me to do?

beckermr: I might be wrong, but I think you need to populate the CI templates in the templates directory.

hadim: But there is no template for the `ci_support` scripts.

hadim: Unless I am missing something...

beckermr: I the actual azure/travis/etc templates etc here: https://github.com/conda-forge/conda-smithy/tree/master/conda_smithy/templates

hadim: Ok so I need to message to https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/templates/build_steps.sh.tmpl instead of populating the `.ci_support/` scripts. Will do.

beckermr: Right except that only does linux. You need to handle windows and osx too.

hadim: The current solution is much less verbose than what you propose. Are you sure about that? I don't understand the rationale behind it. I also don't the codebase very well so sorry if this is a naive question.

beckermr: My general feeling is that putting keys in ci support that are not generated by packages or our pinning files could lead to conflicts or errors. Those files are for conda build to use not for smithy.

beckermr: I’m happy to defer to other members of @conda-forge/core on this though.

hadim: Ok, I understand. Thanks for the explanation.

hadim:
@beckermr Let me know if this is fine.

I used ` grep -rn upload_package conda_smithy/templates/` to spot templates that needed the flag.

hadim:
I forgot the `endif` everywhere :-) Will ping you again when it's ready for review!

beckermr:
Ok. Please don’t add this in the middle of the output validation if block. Otherwise one option will effect the other. It needs to be separate.

hadim:
@beckermr LGTM

conda_smithy/templates/appveyor.yml.tmpl
```diff
@@ -69,4 +69,4 @@ deploy_script:
 {%- if upload_on_branch %}
     - set "UPLOAD_ON_BRANCH={{ upload_on_branch }}"
 {%- endif %}
-    - cmd: upload_package {% if conda_forge_output_validation %}--validate --feedstock-name="{{ feedstock_name }}"{% endif %} .\ .\{{ recipe_dir }} .ci_support\%CONFIG%.yaml
+    - cmd: upload_package {% if conda_forge_output_validation %}--validate --feedstock-name="{{ feedstock_name }}"{% endif %} {% if conda_forge_private_upload %}--private {% endif %} .\ .\{{ recipe_dir }} .ci_support\%CONFIG%.yaml
```
beckermr, line 72: This line will introduce an extra space in the rendered configs which will cause extra commits to feedstocks. There is a space after `--private` and one between the two if blocks. Let's remove those.

hadim: Done. Are you sure about removing the space between the two blocks?

```
{{ feedstock_name }}"{% endif %}{% if conda_forge_private_upload %}
```

beckermr: Right. Add an extra space before `--private` too

hadim:
Thanks, @beckermr for the patience!

PR #1338, opened 2020-05-29
Fix wrong private upload key in templates

hadim:
Checklist
* [ ] Added a ``news`` entry


hadim:
Fix #731 

I was using `conda_forge_output_validation` instead of `private_upload` in the templates.

@beckermr

PR #1344, opened 2020-06-24
Support hexadecimal Github user names

henrykironde:
Checklist
* [ ] Added a ``news`` entry

Use format string to convert  names to strings.
This enables hexadecimal usernames
https://github.com/conda-forge/staged-recipes/issues/11987


conda_smithy/lint_recipe.py
```diff
@@ -723,7 +723,7 @@ def run_conda_forge_specific(meta, recipe_dir, lints, hints):
             # It's a team. Checking for existence is expensive. Skip for now
             continue
         try:
-            gh.get_user(maintainer)
+            gh.get_user({0}.format(maintainer))
```
marcelotrevisani, line 726: maybe just?
```suggestion
            gh.get_user(str(maintainer))
```

Can you create a test for it please? is it possible?

marcelotrevisani: Well, you can monkeypatch the git process which return the user right? monkey patch it to return an hexadecimal
In that way you will reproduce the problem

PR #1354, opened 2020-07-01
Clone all return

thomashopkins32:
Checklist
* [x] Added a ``news`` entry

Changed return type of `feedstocks.clone_all()` from `None` to the list of repositories cloned.
I also changed `feedstocks_clone_all_handle_args` so it does not print the list of Repository objects when using the entry point.


scopatz:
Thanks!

PR #1357, opened 2020-07-09
ci_register.py: Guard against KeyError from empty travis_get_repo_info

tom--pollard:
Snippet modified from deprecated add_project_to_travis func


tom--pollard:
@beckermr Done!

PR #1369, opened 2020-07-31
Follow `requests` logic by loading simplejson if available. Required …

hadim:
…to properly catch JSONDecodeError

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -12,7 +12,7 @@ import hashlib
 import requests
 
 # The `requests` lib uses `simplejson` instead of `json` when available.
-# In consequence the same JSON library must be used of the `JSONDecodeError`
+# In consequence the same JSON library must be used or the `JSONDecodeError`
 # used when catching an exception won't be the same as the one raised
 # by `requests`.
 try:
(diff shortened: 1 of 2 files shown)


scopatz:
Doesn't JSONDecodeError come from the standard library?

hadim:
The one from `json` yes. But `requests` when available uses `simplejson`. In consequence the `JSONDecodeError` error can come from both `json` or `simplejson`. Even if the errors have the same name, they are different objects (since they come from two different libs).

hadim:
So without this PR an error will be raised during rendering if your env contains `simplejson`.

conda_smithy/configure_feedstock.py
```diff
@@ -10,7 +10,11 @@
 import copy
 import hashlib
 import requests
-import json
+
+try:
```
scopatz, line 14: Can you please add a comment as to why this is needed, because it it pretty weird that this is triggered by requests

hadim: I agree this error is weird and boring. I wasn't aware of it but after some googling, it seems to be annoying a lot of people (https://github.com/psf/requests/issues/4842#issuecomment-665502871 among others).

hadim: Done.

scopatz:
Thanks!

PR #1381, opened 2020-08-11
Add lint for deleting staged_recipes example meta.yaml

synapticarbors:
This adds a lint for staged recipes to warn when `recipes/example/meta.yaml` has been removed. I see this not infrequently when reviewing PRs there.

I'm not super familiar with the linting code, so I'm not 100% sure that I've done this correctly. Feedback appreciated.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -742,6 +742,13 @@ def run_conda_forge_specific(meta, recipe_dir, lints, hints):
             "into its own dir."
         )
 
+    # 4: Do not delete example recipe
+    if is_staged_recipes and not os.path.exists("recipes/example/meta.yaml"):
+        lints.append(
+            "Please do not delete the example recipe found in "
+            "(recipes/example/meta.yaml)"
+        )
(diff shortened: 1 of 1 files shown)


jakirkham:
Awesome! Thanks Joshua! 😄

synapticarbors:
This is not working as desired. See https://github.com/conda-forge/staged-recipes/pull/12364#issuecomment-672339382 where the example meta.yaml was not deleted, but the recipe is not in `recipes`.

beckermr:
Those are two different lints, right?

synapticarbors:
The recipe not in `recipes` is not a lint, but an actual CI error. I think this was the first recipe submitted since the new version of `conda-smithy` was released that includes this lint.

beckermr:
Ah. The lint misfired. Indeed.

synapticarbors:
I'm trying to figure out what at what level in the directory hierarchy the lint is being run from, that the path `recipes/example/meta.yaml` is not correct. I'm still working through the code to try to understand what's going on.

PR #1396, opened 2020-10-05
Add link to the feedstock license file

jwillemsen:
Add link to spdx.org for feedstock license

### DIFF
--- a/conda_smithy/templates/README.md.tmpl
+++ b/conda_smithy/templates/README.md.tmpl
@@ -10,7 +10,7 @@ Home: {{ package.meta.about.home }}
 
 Package license: {{ package.meta.about.license }}
 
-Feedstock license: BSD-3-Clause
+Feedstock license: [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause)
 
 Summary: {{ package.meta.about.summary }}
 
--- a/conda_smithy/templates/README.md.tmpl
+++ b/conda_smithy/templates/README.md.tmpl
@@ -11,7 +11,7 @@ Home: {{ package.meta.about.home }}
 
 Package license: {% if license_url %}[{% endif %}{{ package.meta.about.license }}{% if license_url %}]({{ license_url }}){% endif %}
 
-Feedstock license: [BSD-3-Clause](LICENSE.txt)
+Feedstock license: [BSD-3-Clause](https://github.com/{{ github.user_or_org }}/{{ github.repo_name }}/LICENSE.txt)
 
 Summary: {{ package.meta.about.summary }}
 
--- a/conda_smithy/templates/README.md.tmpl
+++ b/conda_smithy/templates/README.md.tmpl
@@ -10,7 +10,7 @@ Home: {{ package.meta.about.home }}
 
 Package license: {{ package.meta.about.license }}
 
-Feedstock license: [BSD-3-Clause](https://spdx.org/licenses/BSD-3-Clause)
+Feedstock license: [BSD-3-Clause](LICENSE.txt)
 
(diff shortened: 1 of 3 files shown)


jakirkham:
As the license file is visible in the top-level directory of the repository, I'm curious why we would want to link to SPDX instead or why such link would be needed at all. Thoughts?

jwillemsen:
If the license is always part of the repo, why not generate a link to the license file?

jakirkham:
Yeah I just assume people can find this easily. That said, no objection to link to it.

jakirkham:
Could you please include a new entry as well?

conda_smithy/templates/README.md.tmpl
```diff
@@ -10,7 +10,7 @@ Home: {{ package.meta.about.home }}
 
 Package license: {{ package.meta.about.license }}
 
-Feedstock license: BSD-3-Clause
+Feedstock license: [BSD-3-Clause](LICENSE.txt)
```
scopatz, line 13: Doesn't this need to be a full URL?

jakirkham: No it's a relative URL based on the feedstock's root directory.

jakirkham (approved):

conda_smithy/templates/README.md.tmpl
scopatz: That seems too github specific, to be honest. It wouldn't be helpful if someone opened a file locally

jwillemsen: With markdown a relative local link also workels, see https://stackoverflow.com/questions/32563078/how-link-to-any-local-file-with-markdown-syntax/58317580#58317580

jakirkham: How about this? Verified to work with [MacDown]( https://macdown.uranusjr.com ) locally.

```suggestion
Feedstock license: [BSD-3-Clause](./LICENSE.txt)
```

jakirkham: Actually it works without the `./` locally. So I don't know that it is needed actually.

jwillemsen: Ok, so when it works without `./` just let it be as it is

scopatz: I'd rather this be a full URL, personally

jakirkham: Sounds good. Will wait to see if Anthony has any other thoughts here :)

jwillemsen: Updated the PR to use a full link

jakirkham: Well that can be tricky if the repo name doesn't match the package name, which does happen when we rename things periodically. Am a little worried about us generating broken links.

jwillemsen: The link generation uses `github.repo_name` which is used more in the README.md, renaming the packaging would keep the same link, only when you change the repo name the links are broken, but that are at that moment all links to github

scopatz: Yeah, this looks good to me

jakirkham: Could you please test this change on this feedstock ( https://github.com/conda-forge/python-feedstock )?

jwillemsen: Made one fix, attached generated file

jakirkham (approved):

jakirkham:
Thanks Johnny! 😄

scopatz:
Thanks!

PR #1397, opened 2020-10-05
Make use of license_url when it has been set in the feedstock meta.yaml

jwillemsen:


jakirkham:
This makes sense to me. I wonder if we can set a local variable with `license_url` and use that in this line. May improve readability a bit.

scopatz:
Please add a news entry: https://regro.github.io/rever-docs/news.html

scopatz:
Thanks!

PR #1421, opened 2020-11-02
ENH move ci to github actions

beckermr:
Checklist
* [x] Added a ``news`` entry


beckermr:
@conda-forge/core this is ready for review!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,40 @@
...
+  push: null
+
+jobs:
+  tests:
+    name: tests
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: goanpeca/setup-miniconda@v1
```
chrisburr, line 13: Should this use https://github.com/conda-incubator/setup-miniconda instead?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          conda install --file requirements.txt black coverage coveralls mock pytest pytest-cov pip ${CONDA_PKGS} --yes --quiet
+          python -m pip install -e .
+          git config --global user.email "smithy@smithy.smithy"
+          git config --global user.name "smithy"
+
+      - name: run tests
+        shell: bash -l {0}
+        run: |
+          py.test tests --cov conda_smithy
```
ocefpaf, line 31: Minor things but `py.test` is the old way to call `pytest`. Not sure if they'll deprecate it and make it go away.

beckermr, line 31:
```suggestion
          pytest tests --cov conda_smithy
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,40 @@
...
+          black --check --verbose --diff --config=pyproject.toml .
```
ocefpaf, line 32: Out of the scope of this PR but should we add pre-commits here? It may be a bit late for that and it could disrupt more than it could help.

.github/workflows/tests.yml
beckermr, line 13:
```suggestion
      - uses: conda-incubator/setup-miniconda@v1
```

ocefpaf (approved):

.github/workflows/tests.yml
beckermr: Yeah, I'd say we punt on that. We might need a big diff to get it to pass.

PR #1434, opened 2020-12-29
BUG fix keys zipped with docker image

beckermr:
Checklist
* [x] Added a ``news`` entry

closes #1409


beckermr:
@jakirkham @isuruf @conda-forge/core This one is ready to go. It should ensure proper rendering for any key zipped with `docker_image`, including `cdt_name` so that we can select `cos7` for cuda 11.

PR #1463, opened 2021-03-15
Full repo checkout for github action to support using tags in recipe.

ryanvolz:
So many thanks to @isuruf for implementing the Github actions provider! I gave it a spin and think that the template would benefit from a simple upgrade:

A common use case (indeed, the default using ci-skeleton) will use `git describe` to set the version in the conda recipe. Without a full depth checkout, tags that would make the git describe output usable are not included. This changes the github-actions template so that it does a full depth checkout so `git describe` will return what the user expects.

Skipping a news entry since the feature hasn't been released yet.

Checklist
* [ ] Added a ``news`` entry


CJ-Wright:
Please add a news entry, that is collated into a changelog when the smithy release is cut

beckermr:
This git checkout there is for the feedstock code, not the source code right. Why do we need tags in the feedstock code?

beckermr:
If we are going to ship this, we should make it optional.

beckermr (changes requested):
See my comments above.

CJ-Wright:
In the `ci-skeleton` workflow the recipe is in the code dir. https://conda-forge.org/docs/user/ci-skeleton.html#getting-started

I agree that this should be optional, most likely with a config in the `conda-forge.yml` which the `ci-skeleton` cmd sets for you.

ryanvolz:
> This git checkout there is for the feedstock code, not the source code right. Why do we need tags in the feedstock code?

It's for the "Using conda-forge to manage your CI" case where the source code *is* the feedstock.

> If we are going to ship this, we should make it optional.

Yeah, I can see that. I wasn't sure if a full-depth checkout was enough overhead to warrant a configuration option. But if that's the consensus, I can add that and make up a news entry.

beckermr:
Ahhh that thing. Yeah we should make it optional and adjust the other provider files to use this option.

beckermr (changes requested):
Does a depth of zero really mean the same thing on all platforms?

ryanvolz:
> Does a depth of zero really mean the same thing on all platforms?

No, but it's not passing the `clone_depth` setting directly for all platforms either. On the ones that I could find that 0 meant "full", it passes that. Otherwise, it just leaves the default (which is full clone) or sets another parameter that *does* indicate full depth.

ryanvolz:
I'll add that I changed the ci-skeleton default in the second commit because I think it makes a better default, but I can drop that.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
```diff
@@ -6,6 +6,10 @@ jobs:
 - job: linux
   {{ azure_yaml|indent(2) }}
   steps:
+{%- if clone_depth is not none %}
+  - checkout: self
+    fetchDepth: {{ clone_depth }}
```
beckermr, line 11: What does a fetch depth of zero mean for azure?

beckermr:
Also, something is odd here. Where is the original checkout command for azure?

ryanvolz:
> Also, something is odd here. Where is the original checkout command for azure?

As far as I can tell, it doesn't exist. When it's not specified, it defaults to a "self" checkout.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
ryanvolz: I thought I saw some documentation of this somewhere, but now I can't find it. I'll just test a value of zero and report back.

ryanvolz: I can confirm that `fetchDepth: 0` works on Azure to set no limit. The resulting git fetch command omits the --depth argument.

conda_smithy/templates/drone.yml.tmpl
```diff
@@ -10,6 +10,11 @@ platform:
   os: linux
   arch: {{ arch_map[data.build_platform.split("-")[1]] }}
 
+{%- if clone_depth is not none %}
+clone:
+  depth: {{ clone_depth }}
```
beckermr, line 15: Zero works on drone too?

beckermr: docs are ambiguous: https://docs.drone.io/pipeline/docker/syntax/cloning/

conda_smithy/templates/circle.yml.tmpl
```diff
@@ -31,7 +31,11 @@ jobs:
         DOCKER_IMAGE: "{{ data.config["docker_image"][-1] }}"
 {%- endif %}
     steps:
+{%- if clone_depth %}
+      - run: git clone --depth {{ clone_depth }} "$CIRCLE_REPOSITORY_URL" --branch "$CIRCLE_BRANCH"
+{%- else %}
       - checkout
+{%- endif %}
```
beckermr, line 38: Is this going to break for repos with special access permissions needed?

conda_smithy/templates/drone.yml.tmpl
ryanvolz: The place where I saw this has turned out to be an archived drone plugin, so I don't know anymore. Is there a good feedstock we can test this on?

conda_smithy/templates/circle.yml.tmpl
ryanvolz: Oh quite possibly. This one was just pulled from some docs and I have no idea if it's worth implementing. The thing I definitely found out is that Circle doesn't let you customize the checkout step. You just have to implement it yourself if you want to do anything custom.

conda_smithy/templates/drone.yml.tmpl
beckermr: you can try this one https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock

beckermr: make a PR turning on drone and add this to the ci scripts after rerendering

conda_smithy/templates/circle.yml.tmpl
ryanvolz: Looks like it breaks, period: https://app.circleci.com/pipelines/github/conda-forge/cf-autotick-bot-test-package-feedstock/1300/workflows/28077417-7245-4308-86c7-8d8ff74ab4ef/jobs/1158

I suppose I could try to dig into this and figure out a clone command that works, or I could just leave it out. I'm inclined to the latter, and I doubt people will miss having the ability to set a finite clone depth on Circle.

conda_smithy/templates/drone.yml.tmpl
ryanvolz: Zero does work on Drone, confirmed: https://cloud.drone.io/conda-forge/cf-autotick-bot-test-package-feedstock/414

ryanvolz:
I tested `clone_depth: 0` and `clone_depth: 1` on all of the providers except github_actions over here: https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/220. I can verify outside of those that github_actions works as well. For Appveyor I ran off of my fork, [here](https://ci.appveyor.com/project/ryanvolz/cf-autotick-bot-test-package-feedstock/builds/38272816) and [here](https://ci.appveyor.com/project/ryanvolz/cf-autotick-bot-test-package-feedstock/builds/38273099).

Aside from Circle with `clone_depth: 1`, they all worked as expected. I even had to tweak Circle to upgrade from xcode 9.0 to xcode 11.0 to get its macOS build running, so clearly no one has done that in awhile. The issue with Circle is that we'll have to write the entire git clone procedure ourselves to be able to set the depth, and while I probably could figure that out with some bashing, it's probably not worth it. If everyone agrees, I can just remove the setting's effect on Circle completely and then this should be good to go.

conda_smithy/templates/circle.yml.tmpl
beckermr: I'm ok removing the circle code. FWIW, it appears the branch circle gives in the variable is not the one in the repo.

isuruf (approved):

beckermr (changes requested):
We need to remove the broken circle command before merging.

conda_smithy/templates/drone.yml.tmpl
beckermr: awesome!

beckermr (approved):

PR #1464, opened 2021-03-17
Log groups: temporarily disable set -x if it was set

jaimergp:
Checklist
* [X] Added a ``news`` entry

Trying to partially address https://github.com/conda-forge/conda-smithy/pull/1436#issuecomment-791040414, aka this confusing log style:

image

### DIFF
--- a/conda_smithy/templates/run_osx_build.sh.tmpl
+++ b/conda_smithy/templates/run_osx_build.sh.tmpl
@@ -4,14 +4,17 @@ source .scripts/logging_utils.sh
 
 set -x
 
-startgroup "Installing a fresh version of Miniforge"
+{ set +x; startgroup "Installing a fresh version of Miniforge" } 2> /dev/null
+
 MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/latest/download"
 MINIFORGE_FILE="Miniforge3-MacOSX-x86_64.sh"
 curl -L -O "${MINIFORGE_URL}/${MINIFORGE_FILE}"
 bash $MINIFORGE_FILE -b
-endgroup "Installing a fresh version of Miniforge"
 
-startgroup "Configuring conda"
+{ set +x; endgroup "Installing a fresh version of Miniforge" } 2> /dev/null
+
+{ set +x; startgroup "Configuring conda" } 2> /dev/null
+
 {% if build_with_mambabuild -%}
 GET_BOA=boa
 BUILD_CMD=mambabuild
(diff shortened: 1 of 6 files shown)


jaimergp:
I need to test this live, but haven't had the time yet. Will report later!

jaimergp:
Ok, I think this should work now... This is how it looks on GitHub Actions:


image


There are some orphan lines but I think those come from changing scripts. It might work correctly on other platforms. At any rate it looks way better than before.

Getting there involves subshells and stderr redirections, so it's not exactly pretty... Let me know what you think @conda-forge/core!

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -42,11 +45,13 @@ echo -e "\n\nMangling homebrew in the CI to avoid conflicts."
...
 {{ build_setup }}
 {%- endif %}
-endgroup "Configuring conda"
+
+{ set +x; endgroup "Configuring conda" } 2> /dev/null
 
 set -e
 
-startgroup "Running conda $BUILD_CMD"
+{ set +x; startgroup "Running conda $BUILD_CMD" } 2> /dev/null
```
isuruf, line 53: Can we remove this group and make conda-build command not be folded?

jaimergp: Done. Check how it looks here:
https://github.com/jaimergp-feedstocks/test-feedstock/runs/2134595185?check_suite_focus=true

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -4,14 +4,17 @@ source .scripts/logging_utils.sh
 
 set -x
 
-startgroup "Installing a fresh version of Miniforge"
+{ set +x; startgroup "Installing a fresh version of Miniforge" } 2> /dev/null
```
mbargull, line 7: If you don't run this in a subshell, you alter the state of `x` option for all following code. Also misses a semicolon before `}`.
(But I guess you just missed changing this file since the others look correct :).)

----
But since you want to (and do) use subshells for which you throw away stderr, you can also simplify these to, e.g.,
```sh
( startgroup "Installing a fresh version of Miniforge" ) 2> /dev/null
```

jaimergp: Oh, nice! I updated all the instances to the simplified version, thanks!

jaimergp:
Suggestions applied. Check results here: https://github.com/jaimergp-feedstocks/test-feedstock/runs/2173923806?check_suite_focus=true

jaimergp:
Anything else needed here @conda-forge/core?

beckermr (approved):

beckermr:
waiting on one of @isuruf or @mbargull to merge in case they have final comments!

thank you!

mbargull (approved):
LGTM!

I wasn't even aware that `function-definition 2>/dev/null` was a thing :). Nice touch!

(And just a general by the by: `echo "$1"` is fine since `-E` is the default. Other shells, e.g., Zsh, might use `-e` by default, so if you want to be really sure, `printf '%s\n' "${1}"`/`printf '::group::%s\n' "${1}"` should always work. But we are guaranteed to be in Bash here, so everything is fine with this PR :).)

PR #1469, opened 2021-04-02
Add option to use different path for feedstock config yaml

ryanvolz:
This makes it possible to specify the feedstock configuration in a file that is named/located at something other than "conda-forge.yml". Feedstock init and ci-skeleton still use the default name, but if a user decides to manually move/rename the config file, there is now a "--feedstock_config" option for appropriate smithy subcommands so that they can point to the alternate path. Clearly this is not useful for conda-forge itself, but it could be used for personal feedstocks and I have a CI application (below).

One test has been added for configure_feedstock to ensure that the configuration can be loaded from an alternative path.

The reason I'd like to see this get merged (or something that accomplishes the same) is that I'm trying to set up some CI that uses conda-smithy (i.e. the ci-skeleton route) and getting a little pushback from the project developers about needing to have a "conda-forge.yml" file at the top level in the repository. Making it be a hidden file like ".conda-forge.yml" might be enough for them, but even better would be the ability to put it in a separate directory. It's already possible to move the recipe directory around, and all of the generated files (at least for ci-skeleton's purposes, sans README and whatnot) live in hidden directories, so the feedstock configuration file is the only thing remaining that sticks out.

I tried to be thorough and go for minimum impact, but I'm happy to make changes if at least the basic idea is acceptable.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/anaconda_token_rotation.py
+++ b/conda_smithy/anaconda_token_rotation.py
@@ -32,7 +32,7 @@ def _get_anaconda_token():
 def rotate_anaconda_token(
     user,
     project,
-    feedstock_directory,
+    feedstock_config_path,
     drone=True,
     circle=True,
     travis=True,
@@ -107,7 +107,7 @@ def rotate_anaconda_token(
                         rotate_token_in_travis(
                             user,
                             project,
-                            feedstock_directory,
+                            feedstock_config_path,
                             anaconda_token,
                             token_name,
                         )
@@ -140,7 +140,7 @@ def rotate_anaconda_token(
                 if appveyor:
                     try:
                         rotate_token_in_appveyor(
-                            feedstock_directory, anaconda_token, token_name
+                            feedstock_config_path, anaconda_token, token_name
                         )
                     except Exception as e:
                         if "DEBUG_ANACONDA_TOKENS" in os.environ:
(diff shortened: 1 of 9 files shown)


ryanvolz:
I'm seeing that I failed to take `conda-forge-ci-setup` into account, and that hard-coded `conda-forge.yml` is a barrier there. I'm currently looking through the code there and trying to figure out how to tackle that.

ryanvolz:
It looks like the only real issue with `conda-forge-ci-setup` is where the `run_conda_forge_build_setup` scripts pull the `channel_priority` from `conda-forge.yml`. For my purposes, this could be "fixed" by simply tweaking the scripts to not fail when `conda-forge.yml` does not exist, and I can put in a PR to that effect. That wouldn't require any changes to this PR, so I'll mark this as ready to review.

To actually make the `channel_priority` work with a config named something other than `conda-forge.yml`, it makes the most sense to me to configure it in the same way as the `channel_sources` and `channel_targets` variables, which are set through `conda_build_config.yaml` (globally through conda-forge-pinning and locally overridden in the recipe dir) and propagated into the configuration files in `.ci_support`. The goal in that case would be to have the `conda-forge-ci-setup` scripts not rely on `conda-forge.yml` at all (except for backward compatibility) so that it is only used for `conda-smithy` functions and not at all post-render.

ryanvolz:
Pinging on this. I've been using it without issue for a little while now to hide the config file as `.conda-forge.yml` in a smithy CI skeleton setup. Are there any fundamental objections to a change like this, or comments on the implementation details? Thanks!

PR #1494, opened 2021-07-19
Limit the amount of parallel jobs running at the same time

jaimergp:
Checklist
* [x] Added a ``news`` entry

As discussed in the last core call, and specially relevant to the CUDA matrix explosion, it would be beneficial to limit the amount of parallel jobs each feedstock adds to the queue (limited to 200 parallel jobs across conda-forge). For example, OpenMM adds 78 Azure jobs for every push 😬 

In this PR, I implemented a workaround based on the [`job.strategy.maxParallel` configuration key](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#define-a-single-job) (see the full example a bit later in that section). Smithy will add this key to the config if the number of configs exceeds the established threshold (50 by default), with a value proportional to the amount of configs found for that OS (Linux, MacOS, Windows).

This is the split for OpenMM:

```
linux = 40 / 78 = 0.51 -> 26
win = 24 / 78 = 0.31 -> 15
osx = 14 / 78 = 0.18 -> 9
```

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1262,7 +1262,6 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
     azure_settings.setdefault("strategy", {})
     azure_settings["strategy"].setdefault("matrix", {})
 
-
     # Limit the amount of parallel jobs running at the same time
     # weighted by platform population
     max_parallel = forge_config["azure"]["max_parallel"]
@@ -1275,9 +1274,9 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
             ]
         )
         ratio = platform_counts[platform.split("-")[0]] / n_configs
-        azure_settings["strategy"]["maxParallel"] = \
-            max(1, round(max_parallel * ratio))
-
+        azure_settings["strategy"]["maxParallel"] = max(
+            1, round(max_parallel * ratio)
+        )
 
     for data in forge_config["configs"]:
         if not data["build_platform"].startswith(platform):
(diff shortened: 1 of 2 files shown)


jaimergp:
I am assuming this will allow other feedstocks to run their jobs while a job-hungry feedstocks has already queued all their jobs... Any idea how to test this?

jakirkham:
Would pick a feedstock with a lot of builds, try re-rendering with this change, and push a PR

jaimergp:
I have put a dummy PR here: https://github.com/conda-forge/openmm-feedstock/pull/59

The Azure UI: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=354355&view=results

After 5 minutes running, the amount of running jobs stays at 50 concurrent, as expected! The rest are queued, but I don't know if other feedstocks would be able to run their tasks before those come in. I guess it would be a reasonable expectation, though.

jaimergp:
Some other feedstocks are running their jobs after that PR's, but I am not sure yet whether they can use the "released" slots (79-50=29) or they are just running on the remaining capacity of the pipeline (up to 200).

Screenshot 2021-07-21 at 12 04 32

beckermr:
I don't think we'll be able to tell if azure is holding workers when the # of jobs on a platform is less than maxParallel. I hope they aren't, but if they are, there is not much we can do about it except to decrease the upper limit. 

Thus I think we should go ahead here.

beckermr (approved):

beckermr:
@conda-forge/core any other comments here?

jakirkham:
> I don't think we'll be able to tell if azure is holding workers when the # of jobs on a platform is less than maxParallel. I hope they aren't, but if they are, there is not much we can do about it except to decrease the upper limit.

We can also reach out to them if we discover issues

PR #1512, opened 2021-10-18
Lint py27, py35 and friends

### LINKED ISSUE #1509: Teach linter to error on non-functional `# [py3x]` selectors
I've been maintaining packages for a while, but was unaware that that conda-build does not support `# [py37]` and higher. It's [mentioned](https://conda-forge.org/docs/maintainer/maintainer_faq.html?highlight=py37#mfaq-py37-selector) in the CF-docs, but that doesn't keep people from using them - since it's such a natural extension from `# [py36]` (which is working) -, and not getting what they're expecting.

Aside from recipes that are probably not working as intended, it also caused me to waste a fair amount of time chasing down a [problem](https://github.com/conda/conda/issues/10975) that was solely due to this detail of selector syntax (which is unfortunately compounded by conda not warning on unknown selectors: https://github.com/conda/conda-build/issues/4298).

In any case, I think the linter should error on `# [py3x]` (possibly only warn on `# [py36]`), so that people do not get caught out like that. Xref https://github.com/conda-forge/conda-smithy/issues/1026, which @jaimergp pointed out to me in https://github.com/conda-forge/conda-forge.github.io/issues/1526, but hasn't been followed up upon in 2.5 years.

jaimergp:
Checklist
* [x] Added a ``news`` entry

Closes #1509 

* This will warn if `py{27,34,35,36}` is used
* It will error out if other numbers are used; e.g. 38, 310, 32, since they are not supported by conda-build and instead silently ignored.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -245,8 +245,9 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
         if pyXY_selectors_lint:
             lints.append(
                 "Old-style Python selectors (py27, py35, etc) are only available "
-                "for Python 2.7, 3.4, 3.5, and 3.6. Please use the int ``py``. For "
-                "example: ``# [py>=37]``. See lines {}".format(pyXY_lines_lint)
+                "for Python 2.7, 3.4, 3.5, and 3.6. Please use explicit comparisons "
+                "with the integer ``py``, e.g. ``# [py==37]`` or ``# [py>=37]``. "
+                "See lines {}".format(pyXY_lines_lint)
             )
 
     # 7: The build section should have a build number.


conda_smithy/lint_recipe.py
```diff
@@ -208,21 +208,46 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
...
+            hints.append(
+                "Old-style Python selectors (py27, py34, py35, py36) are "
+                "deprecated. Instead, consider using the int ``py``. For "
+                "example: ``# [py>=36]``. See lines {}".format(pyXY_lines_hint)
+            )
+        if pyXY_selectors_lint:
+            lints.append(
+                "Old-style Python selectors (py27, py35, etc) are only available "
+                "for Python 2.7, 3.4, 3.5, and 3.6. Please use the int ``py``. For "
+                "example: ``# [py>=37]``. See lines {}".format(pyXY_lines_lint)
```
h-vetinari, line 249:
```suggestion
                "Old-style Python selectors (py27, py35, etc) are only available "
                "for Python 2.7, 3.4, 3.5, and 3.6. Please use explicit comparisons "
                "with the integer ``py``, e.g. ``# [py==37]`` or ``# [py>=37]``. "
                "See lines {}".format(pyXY_lines_lint)
```

PR #1513, opened 2021-10-19
Add ccache caching to Azure Pipelines

IvanYashchuk:
This PR adds the "Cache" task to Azure Pipelines steps as the first step that caches the "ccache" folder. This would allow to significantly decrease the build time for C++ projects.

When using Docker with `run_docker_build.sh` CCACHE_DIR is mounted to `/home/conda/.cache/ccache` which is default inside the container.
Whoever uses ccache would need to set `CCACHE_BASEDIR` in the build script for example to the home directory because each conda build happens in different directories.

It's not possible to add steps correctly using `conda-forge.yml` to do it on a per-recipe basis. For builds that do not use ccache this change is harmless.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
@@ -5,7 +5,16 @@
 jobs:
 - job: linux
   {{ azure_yaml|indent(2) }}
+  variables:
+    CCACHE_DIR: $(Pipeline.Workspace)/ccache
   steps:
+  - task: Cache@2
+    inputs:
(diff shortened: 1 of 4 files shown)


IvanYashchuk:
> > It's not possible to add steps correctly using conda-forge.yml to do it on a per-recipe basis.
> 
> Why not?

Because currently settings from conda-forge.yml are added in the beginning with `{{ azure_yaml|indent(2) }}`.
See this message https://gitter.im/conda-forge/conda-forge.github.io?at=60d5d6dcbcb3e8060ee37e67

PR #1515, opened 2021-10-21
slightly modify Comic at end of README

sgbaird:
The current version is a bit off-color. Would it be possible to modify the image slightly?

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,9 +129,8 @@ $ rever 0.1.2
 Conda-smithy in a nutshell
 --------------------------
 
-#### xkcd 1629: Tools
-
-[![xkcd 1629: Tools](https://imgs.xkcd.com/comics/tools.png)](https://xkcd.com/1629/)
+#### xkcd 1629: Tools (modified)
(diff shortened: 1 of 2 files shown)


sgbaird:
@BastianZim, apologies for not being clearer. I edited the comic in PowerPoint to remove the reference to pornography at the very end. That's what I meant by off-color.

BastianZim:
Ahh sorry now I have to clarify myself. I would keep it since it as is  since captures the spirit of the XKCD and thought that the wording is artistic freedom that we should tolerate. :)

wolfv:
I did get that :) just wanted to point out that removing the link & just modifying the image might not be OK with the license. 
Personally, I don't mind either option -- keeping, changing or removing.

sgbaird:
I added the links to the original and the xkcd license page. I figured a <sub> tag underneath the image worked well.

timsnyder:
Think this can be closed as superseded by #1802

PR #1537, opened 2021-11-03
Use https for NumFOCUS

mwtoews:
Prefer to use secure https links for https://numfocus.org <s>and https://www.w3.org/2000/svg</s>

Not modified in this PR are files in `tests/recipes/click-test-feedstock`, as it's part of the test suite. Maybe someone wants to follow-up with that?


mwtoews:
Removed change to SVG file, as apparently this is an issue with some browsers.

jakirkham:
Thanks Mike! 😄

PR #1501, opened 2021-09-15
ENH support master or main for tooling

beckermr:
Checklist
* [x] Added a ``news`` entry

This PR adds support for master or main as the default branch for the tooling repos.

PR #1553, opened 2021-12-10
Switch GitHub Actions to main

BastianZim:
Checklist
* [x] Added a ``news`` entry

Following up on https://github.com/conda-forge/conda-smithy/pull/1501 this PR switches the GitHub actions to `main` (from `master`) as that's what they're using now:
* https://github.com/conda-forge/automerge-action/branches
* https://github.com/conda-forge/webservices-dispatch-action/branches

From what I can see in the [search](https://github.com/conda-forge/conda-smithy/search?q=tooling_branch_name) the key is not used anywhere else so this should not affect anything else.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1724,7 +1724,7 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
             "user_or_org": "conda-forge",
             "repo_name": "",
             "branch_name": "master",
-            "tooling_branch_name": "master",
+            "tooling_branch_name": "main",
         },
         "github_actions": {
             "self_hosted": False,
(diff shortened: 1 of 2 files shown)


BastianZim:
@beckermr Pinging you here as you're overseeing the whole migration.

beckermr (changes requested):
Thank you for the PR but we cannot merge this. Our automated rerendering will break if we do due to permissions on the github token it uses.

BastianZim:
Ahh ok makes sense. Just out of personal curiosity, wouldn't it make more sense then to switch the actions back to master?

beckermr:
I am keeping tags called "master" on the actions to make sure everything works. The plan is to eventually move all of conda-forge to main. We're just in the middle of that process now.

BastianZim:
Ahh, that's smart. Ok, thanks for the explanation. I'll close this for now then.

PR #1554, opened 2021-12-13
ENH add new token to GHA template

beckermr:
Checklist
* [x] Added a ``news`` entry

This PR adds a new token to the GHA template for rendering and the one for automerge. Eventually, the admin bot will add this token to the repos secrets as needed. For now, I have tested things and it does no harm being here. I have also used code from Isuru to adjust the rendering bot to not change girhub action workflow files unless the token it has allows it. Thus we can merge and release changes here without an issue.


beckermr:
@conda-forge/core any comments here?

beckermr:
This change has gone live via admin-requests FYI.

beckermr:
OK merging. LMK if you have comments etc.

PR #1560, opened 2022-01-02
Update license year

BastianZim:
Checklist
* [x] Added a ``news`` entry


BastianZim:
Ping @conda-forge/core

SylvainCorlay:
I guess we could remove the year in the copyright header altogether.

BastianZim:
Would make sense especially for the feedstocks. Since they don't get rerendered regularly, I've seen some still stuck in 2018. So if the license should always be correct, the feedstocks either have to get rerendered more often (difficult) or the year removed (much easier) entirely.

PR #1576, opened 2022-02-11
Update versioneer and exclude it from pre-commit

dbast:


dbast:
Uh, "pep440-branch-based" requires https://github.com/python-versioneer/python-versioneer/pull/174 ...

PR #1583, opened 2022-02-16
add handling for boa recipes

wolfv:
This is a preliminary PR to add handling for `boa` recipes. It didn't require many changes since we are duck-typing the relevant `MetaData` class in `boa` anyways. Variant handling is slightly different but we can certainly patch that up on the `boa`-side.
Also you can completely ignore the changes in the templates for now, since those were only made to see if I can get the xtensor-feedstock to work.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -21,6 +21,13 @@ try:
 except ImportError:
     import json
 
+try:
+    import boa.api
+
+    boa_available = True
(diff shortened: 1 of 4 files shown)


wolfv:
Additional changes that were necessary in `boa` are found here: https://github.com/mamba-org/boa/pull/240/files

baszalmstra:
I believe we should close this. rattler-build has superseded boa.

PR #1596, opened 2022-02-21
Fix uninstallation of conda-forge-ci-setup

hmaarrfk:
This attempts to help the bootstrap process uninstall any pre-installed version of `conda-forge-ci-setup`.

I've added comments to the effect where I knew how to do it in the relevant syntax.

The purpose of specifying extra packages was specifically designed in to avoid needless churn with the `tensorflow` package which depends on `lief_dev` which has not had a new release in about 8 months.

I'm really trying to contain scope creep here and not balloon the options provided in `conda-forge.yml`

References: 
Lief releases: https://github.com/lief-project/LIEF/releases

Request to use lief dev: https://github.com/conda-forge/tensorflow-feedstock/pull/189#issuecomment-1029924154

Interested parties: @beckermr @isuruf @xhochy

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/templates/appveyor.yml.tmpl
+++ b/conda_smithy/templates/appveyor.yml.tmpl
@@ -55,7 +55,9 @@ install:
     # Tell conda we want an updated version of conda-build and {{ " ".join(remote_ci_setup) }}
     - cmd: conda.exe install -n root -c conda-forge --quiet --yes conda-build pip {{ " ".join(remote_ci_setup) }}
     {%- if local_ci_setup %}
-    - cmd: conda.exe uninstall --quiet --yes --force {{ " ".join(remote_ci_setup) }}
+    # Uninstall the conda-forge-ci-setup package but leave its dependencies
+    # so that we may bootstrap the installation process
+    - cmd: conda.exe uninstall --quiet --yes --force conda-forge-ci-setup
     - cmd: pip install --no-deps .\{{ recipe_dir }}\.
     {%- endif %}
     - cmd: setup_conda_rc .\ .\{{ recipe_dir }} .\.ci_support\%CONFIG%.yaml
--- a/conda_smithy/templates/azure-pipelines-win.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-win.yml.tmpl
@@ -68,7 +68,7 @@ jobs:
     - script: |
         call activate base
         {%- if local_ci_setup %}
-        conda.exe uninstall --quiet --yes --force {{ " ".join(remote_ci_setup) }}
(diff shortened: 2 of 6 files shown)


conda_smithy/templates/appveyor.yml.tmpl
```diff
@@ -55,7 +55,9 @@ install:
     # Tell conda we want an updated version of conda-build and {{ " ".join(remote_ci_setup) }}
     - cmd: conda.exe install -n root -c conda-forge --quiet --yes conda-build pip {{ " ".join(remote_ci_setup) }}
     {%- if local_ci_setup %}
-    - cmd: conda.exe uninstall --quiet --yes --force {{ " ".join(remote_ci_setup) }}
+    # Uninstall the conda-forge-ci-setup package but leave its dependencies
+    # so that we may bootstrap the installation process
+    - cmd: conda.exe uninstall --quiet --yes --force conda-forge-ci-setup
```
beckermr, line 60: This was an option so the package could be renamed. We should preserve that.

hmaarrfk: The 3rd option is to leave it as is.

The current "state" works for tensorflow, and works for `conda-forge-ci-setup`. This is "good enough" for me.

beckermr: We actually have the version as part of the option: https://github.com/conda-forge/conda-smithy/blob/c299172b8e86f16b94d5f02234e89df35c5440db/conda_smithy/configure_feedstock.py#L1759

So smithy has been using a mix of things. 

I don't have strong feelings here, but it seems like if folks want to add extra ci setup packages, then we should have added a new optional originally called "extra_ci_setup" or similar. 

In any case, most things are half-broken most of the time. I'll defer to the rest of core here. 

cc @conda-forge/core

hmaarrfk: I had originally proposed something like `extra`, it was rejected. I think we hadn't considered the bootstrap procedure. Selfishly speaking, i don't have many reasons to push this PR forward

beckermr: That is totally fine. Let's close this one and leave things as they are. We can go back and fix later by making the conda-forge specific stuff a separate option since it is special anyways. Thank you for all of the hard work here!

beckermr: xref: https://github.com/conda-forge/conda-smithy/issues/1597

beckermr:
closing as stated above!

PR #1613, opened 2022-03-27
Skip render tuneup

timsnyder:
Checklist
* [x] Added a ``news`` entry

[skip_render can skip github_actions_services](https://github.com/conda-forge/conda-smithy/commit/6fa380ce60f0fa388dc9e0866bb82df25dc888f3) 

Also enhanced it so that you can skip rendering anything
below a directory by specifying the directory name (without
For example, if '.github' is in the skip_render list, all
files inside .github/ will be skipped.

Logging now tells you when rendering was actually skipped
as opposed to just what was read from the config.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -59,9 +59,10 @@ def package_key(config, used_loop_vars, subdir):
     )
     return key.replace("*", "_").replace(" ", "_")
 
+
 def _ignore_match(ignore, rel):
     """Return true if rel or any of it's PurePath().parents are in ignore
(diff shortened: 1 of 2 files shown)


timsnyder:
If we decide that this is an okay addition, I'll update https://conda-forge.org/docs/maintainer/conda_forge_yml.html#skip-render

timsnyder:
@conda-forge/conda-smithy could someone review this please?  We reuse the CI configuring abilities in sparcians/map but would like to be able to configure smithy to not render the webservices and automerge workflows.  Thanks!

Issue #1616, opened 2022-04-05
Lint `conda-forge.yml`


jakirkham:
Sometimes issues in `conda-forge.yml` can crop up. It would be useful in these cases to have a way to lint `conda-forge.yml` and share this feedback with users in PRs before that content makes into `main`.

bollwyvl:
Driving it with no-fooling JSON schema, and encouraging referring to it directly in the file via `$schema`, would be pretty excellent, as people's editors would then give them nice autocompletion and diagnostics from e.g. https://github.com/redhat-developer/yaml-language-server

jaimergp:
Closed by #1756

PR #1619, opened 2022-04-29
[wip] add conda-forge.yml linting and schema validation

bollwyvl:
Checklist
* [ ] Added a ``news`` entry
- [x] sorts deps
- [x] adds `yamllint` and `jsonschema`
- [x] adds `--quiet` for all of the lint subcommands
- [x] adds a schema
  - [x] harmonize with [conda_forge_yml.rst](https://github.com/conda-forge/conda-forge.github.io/blob/main/src/maintainer/conda_forge_yml.rst)
    - if done cleverly, that page could be generated from the schema with e.g. `sphinx-jsonschema`
    - the `.rst` is presently injected into a `.yml` version: this is strictly _less_ useful than a `.json` file, but
      - it would be reasnonable enough to treat the `.yml` version as the input, and generate/ship/reference the json one
- [x] move annotated defaults to yaml file 
- [x] adds `lint-config` CLI
  - [x] runs `yamllint` on the file-on-disk, ignoring line length, indentation
  - [x] checks immediately _after_ being merged with the defaults, but _before_ any normalization occurs
- [x] adds `lint-feedstock` CLI (includes `lint-recipe`)
- [x] ~~update `black` (could be a separate PR)~~ #1629
- [ ] tests

References:
- starts #1616

Notes:
- this runs with one finding against ~200 feedstocks i happen to have sitting around
  - only thing i found is a case of `True` which `yamllint` would like to be `true`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,6 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/psf/black
-    rev: 21.7b0
+    rev: 22.3.0
     hooks:
     -   id: black
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include requirements.txt
 
 recursive-include conda_smithy/feedstock_content *
 recursive-include conda_smithy/templates *.tmpl
+recursive-include conda_smithy/schema *.json *.yml *.yaml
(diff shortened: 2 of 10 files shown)


jaimergp:
Thanks @bollwyvl for all the work you did here, plus all the other comments you provided in the reviews and discussions. It helped shape the final result that ended up in the codebase. Greatly appreciated! I'll close this now.

PR #1620, opened 2022-05-04
update pre-commit to black stable

bollwyvl:
Checklist
* [ ] Added a ``news`` entry

This updates `black` to a stable version that doesn't fail to install.

References

- from #1619


bollwyvl:
Thanks!

PR #1642, opened 2022-06-30
fix index for pin_compatible lint

minrk:
in #1624, `pin[-1]` was incorrect if build pattern is specified, picking up the build string instead of the package name. The message has the right index, but the test didn't match what the message was claiming. Fixing the test to match the message allows lint to pass.

Incorrect lint example: https://github.com/conda-forge/petsc-feedstock/pull/146#issuecomment-1169836525

Checklist

* [x] Added a ``news`` entry


carterbox (approved):
I wrote this linting rule, and this patch makes sense to me.

carterbox:
Closes #1670

Issue #1677, opened 2022-09-27
Unable to install conda-smithy with conda 22.9.0 installed


mcflugen:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

I am unable to install *conda-smithy* with the most recent version of *conda* (22.9.0). This, I believe, is a result of *conda* switching from *SemVer* to *CalVer* but *conda-smithy* still having `CONDA_VERSION_MAX = "5.0"`.

```bash
$ conda smithy rerender
You appear to be using conda 22.9.0, but conda-smithy 3.21.1
is currently only compatible with conda versions < 5.0.
```

### Installed packages

```shell
# Name                    Version                   Build  Channel
anyio                     3.6.1              pyhd8ed1ab_1    conda-forge
appnope                   0.1.3              pyhd8ed1ab_0    conda-forge
argon2-cffi               21.3.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0           py39h63b48b0_2    conda-forge
asttokens                 2.0.8              pyhd8ed1ab_0    conda-forge
attrs                     22.1.0             pyh71513ae_1    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
bat                       0.20.0               hbbacdb1_0    conda-forge
beautifulsoup4            4.11.1             pyha770c72_0    conda-forge
bleach                    5.0.1              pyhd8ed1ab_0    conda-forge
blinker                   1.4                        py_1    conda-forge
boa                       0.11.0             pyha770c72_3    conda-forge
boolean.py                3.7                        py_0    conda-forge
brotlipy                  0.7.0           py39h63b48b0_1004    conda-forge
bzip2                     1.0.8                h0d85af4_4    conda-forge
c-ares                    1.18.1               h0d85af4_0    conda-forge
ca-certificates           2022.9.24            h033912b_0    conda-forge
certifi                   2022.9.24          pyhd8ed1ab_0    conda-forge
cffi                      1.15.1           py39hc55c11b_0
chardet                   5.0.0            py39h6e9494a_0    conda-forge
charset-normalizer        2.1.1              pyhd8ed1ab_0    conda-forge
click                     8.1.3            py39h6e9494a_0    conda-forge
[...]

PR #1679, opened 2022-09-27
Debug CI problems with tags and versioneer

jaimergp:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,8 @@ jobs:
         shell: bash -l {0}
         run: |
           mamba install --file requirements.txt pre-commit coverage coveralls mock pytest pytest-cov pip ${CONDA_PKGS} --yes --quiet
-          python -m pip install -e .
+          mv pyproject.toml pyproject.toml.bak
+          python -m pip install -e -vv .
           git config --global user.email "smithy@smithy.smithy"
           git config --global user.name "smithy"
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,8 +26,11 @@ jobs:
         shell: bash -l {0}
         run: |
           mamba install --file requirements.txt pre-commit coverage coveralls mock pytest pytest-cov pip ${CONDA_PKGS} --yes --quiet
+          # get pyproject.toml out of the way so setuptools doesn't complain
           mv pyproject.toml pyproject.toml.bak
           python -m pip install -vv -e .
+          # restore so pre-commit's black get its config 
+          mv pyproject.toml.bak pyproject.toml
           git config --global user.email "smithy@smithy.smithy"
           git config --global user.name "smithy"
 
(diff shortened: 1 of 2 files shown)


jaimergp:
This fails, as expected, due to #1677, but the tests do run now.

@conda-forge/core, ready for review :)

.github/workflows/tests.yml
```diff
@@ -26,7 +26,11 @@ jobs:
         shell: bash -l {0}
         run: |
           mamba install --file requirements.txt pre-commit coverage coveralls mock pytest pytest-cov pip ${CONDA_PKGS} --yes --quiet
-          python -m pip install -e .
+          # get pyproject.toml out of the way so setuptools doesn't complain
+          mv pyproject.toml pyproject.toml.bak
```
jakirkham, line 30: What is setuptools complaining about?

.github/workflows/tests.yml
```diff
@@ -26,7 +26,11 @@ jobs:
...
+          python -m pip install -vv -e .
+          # restore so pre-commit's black get its config
+          mv pyproject.toml.bak pyproject.toml
```
jakirkham, line 33: Given this just has a `black` setting in it, wonder if we should just move this into a `black` specific config file to avoid issues with other build tools

jakirkham (commented):
Thanks Jaime! 🙏

Had a couple questions below

.github/workflows/tests.yml
jaimergp: Black _only_ supports pyproject.toml 😬 

https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file

.github/workflows/tests.yml
jaimergp: I guess we could also pin setuptools to an earlier (no clue which one) version, but same point as above. I just wanted a quick fix to unblock people affected by #1677

jakirkham: Ok let's raise an issue to address this later (likely by migrating to `pyproject.toml`)

.github/workflows/tests.yml
jakirkham: `black` does take [a `--line-length` CLI argument]( https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html?highlight=--line-length#line-length ). Maybe we can just pass this to `black` directly?

.github/workflows/tests.yml
hmaarrfk: Is this a versionneer issue? If so i e been moving my projects to miniver https://github.com/jbweston/miniver

jaimergp: I thought so at first (hence the branch name), but it's unrelated. The warning we get right on top of the error just happens to be there.

.github/workflows/tests.yml
jaimergp: We could in CI, but people developing locally won't have that config setting and will reformat everything with the defaults.

beckermr:
I very much appreciate the deliberative approach here but these semver checks are breaking the bot. Shall we proceed with all haste?

beckermr:
I'm going to merge this to main and then merge main into the semver PR.

PR #1702, opened 2022-12-07
`{{ blas_impl}}` in requirements.host lints cleanly.

timsnyder:
Checklist
* [X] Added a ``news`` entry

In conda-forge/dakota-feedstock#31, when building `dakota` using the netlib reference BLAS packages and then testing with openblas, we get a segfault as detailed at https://github.com/conda-forge/dakota-feedstock/pull/31#issuecomment-1335803404.
If we specifically depend on `{{ blas_impl }}` [as described in the conda-forge mantainer knowledge base](https://conda-forge.org/docs/maintainer/knowledge_base.html#blas:~:text=If%20a%20package%20needs%20a%20specific%20implementation%E2%80%99s%20internal%20API%20for%20more%20control%20you%20can%20have) `dakota` builds and tests cleanly.  However, `conda smithy recipe-lint` doesn't handle `{{ blas_impl }}` in `requirements.host`.  The current code hits an `AttributeError` at https://github.com/conda-forge/conda-smithy/blob/2bd12c86814b358ab548a3b6581212d4a0fd2832/conda_smithy/lint_recipe.py#L478 because `blas_impl` is not defined in the `jinja2.Environment` used by `conda_smithy.utils.render_meta_yaml`.

This PR defines `blas_impl` so that it will lint.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -475,6 +475,8 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
     # 22: Single space in pinned requirements
     for section, requirements in requirements_section.items():
         for requirement in requirements or []:
+            # rather than adding to utils.render_meta_yaml() we could
+            # handle requirement is None by ignoring it
             req, _, _ = requirement.partition("#")
             if "{{" in req:
                 continue
--- a/conda_smithy/utils.py
+++ b/conda_smithy/utils.py
@@ -82,6 +82,7 @@ def render_meta_yaml(text):
             time=time,
             target_platform="linux-64",
             mpi="mpi",
+            blas_impl="openblas",
         )
     )
     mockos = MockOS()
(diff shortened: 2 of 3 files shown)


conda_smithy/lint_recipe.py
```diff
@@ -475,6 +475,8 @@ def lintify(meta, recipe_dir=None, conda_forge=False):
     # 22: Single space in pinned requirements
     for section, requirements in requirements_section.items():
         for requirement in requirements or []:
+            # rather than adding to utils.render_meta_yaml() we could
+            # handle requirement is None by ignoring it
```
timsnyder, line 479: I don't completely understand how `{{ blas_impl }}` gets rendered as `None` given the implementation of `conda_smithy.utils.NullUndefined`.  However, thinking about this more, it probably makes sense to continue whitelisting variables in `utils.render_meta_yaml` rather than ignoring `requirement is None` here.

timsnyder:
@conda-forge/conda-smithy, could I get some feedback on this and possibly get it merged so that we can move forward in conda-forge/dakota-feedstock#31?

isuruf (changes requested):
It's not scalable to add all the variables that we have in our main config and whatever a user may define in conda_build_config.yaml.

Please use a trick like https://github.com/conda-forge/gfortran_impl_osx-64-feedstock/blob/main/recipe/meta.yaml#L5-L8

timsnyder:
Sounds good.  With the trick @isuruf pointed to, we don't need to change the smithy code.  Closing this PR

PR #1704, opened 2022-12-22
Rework error catching on Windows (WIP)

jaimergp:
Checklist
* [ ] Added a ``news`` entry

I found a case where [conda-build crashes with error level -1](https://github.com/napari/packaging/issues/50#issuecomment-1362008150), which the current constructs were not catching (only >=1). I found more info in this [SO answer](https://stackoverflow.com/questions/34936240/batch-goto-loses-errorlevel/34937706#34937706) and this might worth a more thorough audit of `conda-build-setup-ci` and its usage (e.g. using `CALL`).

Leaving this here while I evaluate other error catching options on Windows (like using `||`).


jaimergp:
Superseded by https://github.com/conda-forge/conda-smithy/pull/1763

PR #1706, opened 2023-01-07
BUG always check team membership even when making the team

beckermr:
The person on the token used to make the team gets added by default. This means our bots appear as team members. This PR fixes that so that the bots are not on the teams. This is a potential security risk since it escalates permissions. 

Checklist
* [x] Added a ``news`` entry


beckermr:

jakirkham (approved):

PR #1713, opened 2023-01-17
REF move to setuptools_scm and only pyproject.toml

beckermr:
Checklist
* [x] Added a ``news`` entry


beckermr:
@ocefpaf care to review?

ocefpaf:
> @ocefpaf care to review?

Here are a few tests to ensure we are writing the version:

1. `python setup.py --version` from the git source tree.
```
3.22.2.dev41+ge9ab3eb
```

2. `python -m build --sdist . --outdir dist`
```
conda-smithy-0.0.0.tar.gz
```

3. `python setup.py sdist`
```
conda-smithy-3.22.2.dev41+ge9ab3eb.tar.gz
```

4. pip installing from the sdist (2 and 3) and checking the version
```
python -c "import conda_smithy; print(conda_smithy.__version__)"
0.0.0
```

I usually use the write to file option b/c it is simpler and avoids that error. Here is the pattern I use: https://github.com/ioos/ioos_qc/blob/121620a2ac8955885db0c7b0a06e8b6b95a41ac3/setup.py#L5-L9

jakirkham:
Wonder if moving build requirements to `pyproject.toml` would help the `python -m build` case

beckermr:
Yeah build is broken even if I use the explicit file.

beckermr:
Wait NVM. The build and non-build sdists work fine for me locally.

beckermr:
The sdists have 0.0.0 in the name but they all give the correct version on install.

ocefpaf:
> The sdists have 0.0.0 in the name but they all give the correct version on install.

Not sure how you are testing it but here they do not work:

```
❯ pip install conda-smithy-3.22.2.dev41+ge9ab3eb.tar.gz
Processing ./conda-smithy-3.22.2.dev41+ge9ab3eb.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: conda-smithy
  Building wheel for conda-smithy (pyproject.toml) ... done
  Created wheel for conda-smithy: filename=conda_smithy-0.0.0-py3-none-any.whl size=109593 
Installing collected packages: conda-smithy
Successfully installed conda-smithy-0.0.0 
conda-smithy/dist on  setup-refactor via 🅒 TEST took 2s 
❯ python -c "import conda_smithy; print(conda_smithy.__version__)"
0.0.0
```

beckermr:
```
          rm -rf dist/*
          python setup.py sdist
          pip install dist/*.tar.gz
          python -c "import conda_smithy; assert conda_smithy.__version__ != '0.0.0'"
          pip uninstall conda-smithy --yes
          
          rm -rf dist/*
          python -m build --sdist . --outdir dist
          pip install dist/*.tar.gz
          python -c "import conda_smithy; assert conda_smithy.__version__ != '0.0.0'"
          pip uninstall conda-smithy --yes
```

^ code to test

ocefpaf:
Looks like you are in the source tree directory, no?

beckermr:
That didn't matter locally. Pushed a change to make sure. Let's see what happens.

.github/workflows/tests.yml
```diff
@@ -51,13 +51,17 @@ jobs:
           rm -rf dist/*
           python setup.py sdist
           pip install dist/*.tar.gz
+          pushd ..
           python -c "import conda_smithy; assert conda_smithy.__version__ != '0.0.0'"
```
ocefpaf, line 55: https://twitter.com/ocefpaf/status/1562400506437373953

ocefpaf:
> That didn't matter locally. Pushed a change to make sure. Let's see what happens.

Really? That is odd b/c in the root tree you have the git info and doesn't need the file. However, with the _version.py now it works like a charm.

.github/workflows/tests.yml
beckermr: bahahahahahahahahahah

I am so confused right now. What is all of this stuff for? Why does the sdist have 0.0.0 in the name? Where do I put anything?

ocefpaf (approved):

beckermr:
I have to say I am nervous to merge this one. I'm gonna hack on it more to try and figure out stuff.

.github/workflows/tests.yml
ocefpaf: The sdist with build has build isolation, so no setuptools_scm is installed and no version is recorded in the name. The direct call to setup.py uses whatever you have install so, if you have setuptools_scm you will get a version in the name. That version is derived from the git metadata and, as long as you are in the git tree you have it in the package too.

If you create an sdist that does not ship enough git metadata then you don't get a correct version. Some folks use the .git_archive.txt (or something similar) but I prefer the write to you that you implemented here.

TL;DR Ideally we need setuptools_scm in the pyproject.toml too to cover all cases.

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 100
+extend-ignore = E203
```
beckermr, line 3:
```suggestion
extend-ignore = E203

```

.github/workflows/tests.yml
beckermr: This is fixed now. I moved the entire thing to pyproject.toml

beckermr:
@ocefpaf @jakirkham I fixed this up one more time. We are now entirely on pyproject.toml, using the beta setuptools stuff. Everything works as expected. Whew. Please look again!

setup.py
```diff
@@ -1,26 +1,3 @@
 from setuptools import setup
 
-setup(
-    name="conda-smithy",
```
ocefpaf: There is a long standing bug in GitHub that only reads the name metadata of Python packages from setup.py. Not that we need it but removing this line from here will mess up the info in https://github.com/conda-forge/conda-smithy/network/dependents.

PS: really not a big deal for conda-smithy but something to consider for other packages.

news/prod.rst
```diff
@@ -0,0 +1,3 @@
+**Changed:**
+
+* Changed python packaging to use setuptools-scm instead of versioneer.
```
beckermr, line 3:
```suggestion
* Changed python packaging to use setuptools-scm instead of versioneer.
* Moved build system to only use ``pyproject.toml``.
```

pyproject.toml
```diff
@@ -1,3 +1,31 @@
+[build-system]
+requires = ["setuptools>=45", "setuptools_scm[toml]>=7"]
```
jakirkham, line 2: Would suggest breaking this over multiple lines (makes diffs that change this easier to read)

Also calling out `tomli` explicitly (and when we can drop it as the standard library includes it) may be helpful

```suggestion
requires = [
    "setuptools>=45",
    "setuptools_scm>=7",
    "tomli>=1.0.0; python_version < '3.11'",
]
```

pyproject.toml
```diff
@@ -1,3 +1,31 @@
...
+[project.urls]
+home = "https://github.com/conda-forge/conda-smithy"
+
+[project.scripts]
+feedstocks = "conda_smithy.feedstocks:main"
+conda-smithy = "conda_smithy.cli:main"
+
+[tool.setuptools]
+packages = ["conda_smithy"]
+zip-safe = true
```
jakirkham, line 22: This was `False` in `setup.py`. Curious why it can be `true` now?

beckermr: ugh - meant to be False! arghhhh

setup.py
beckermr: Ah ok. I propose we let this bug be a bug and remove this.

pyproject.toml
```diff
@@ -1,2 +1,47 @@
+[build-system]
+requires = ["setuptools>=45", "setuptools_scm[toml]>=7"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "conda-smithy"
+authors = [
+    {name = "Phil Elson", email = "pelson.pub@gmail.com"},
+]
```
jakirkham, line 13: Wonder if we should move to [`maintainers`]( https://peps.python.org/pep-0621/#authors-maintainers ) with a new email. Admittedly not critical right now

pyproject.toml
```diff
@@ -1,2 +1,51 @@
+[build-system]
```
hmaarrfk, line 1: How do you disable build isolation with pyproject.tom? I was always confused and avoided it with python code that was "conda first"

beckermr: I think it just works with a flag?

https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#disabling-build-isolation

hmaarrfk: I see. so it is something that users have to do, not something you can specify as part of the pyproject.toml file.

beckermr: Right. We'll have to add it to recipes or use env vars as needed.

hmaarrfk: I think we already setup the pip configuration in our standard build machinery for feedstocks. More just on "general development" instructions.

I guess maybe I need to get on board with this pyproject thing.

beckermr: I feel you man. I am skeptical but with the setuptools integration in beta, at least everything (but flake8) can go in the file. For complicated builds I still don't see a way around a setup.py but who knows.

pyproject.toml
jakirkham: Raised as issue ( https://github.com/conda-forge/conda-smithy/issues/1714 )

PR #1724, opened 2023-03-09
Fix syntax error when preparing conda artifacts on Azure + Windows

### LINKED ISSUE #1723: Prepare conda build artifacts step fails on Azure + Windows with "The syntax of the command is incorrect."

### Issue

Since #1701, and version 3.23.0, enabling storage of the conda build artifacts causes a failure on Azure + Windows during the "Prepare conda build artifacts" step, producing the error message "`The syntax of the command is incorrect.`" See [example](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=670151&view=logs&jobId=a70f640f-cc53-5cd3-6cdc-236a1aa90802&j=a70f640f-cc53-5cd3-6cdc-236a1aa90802&t=2f510c44-5236-5714-95fa-3351828195a7).

The syntax error comes from the fact that `%GITHUB_OUTPUT%` is not defined on Azure (it is for GitHub Actions), and despite the fact that that variable is only used behind an `if` statement that checks for GHA CI, the script fails to parse since the variable expands to the empty string.

Switching to delayed expansion by accessing that variable as `!GITHUB_OUTPUT!` eliminates the syntax error. PR incoming.

### Installed packages

```shell
N/A
```

### Environment info

```shell
N/A
```

ryanvolz:
Checklist
* [x] Added a ``news`` entry

Fixes #1723.

Forgot to add: I tested this over at https://github.com/conda-forge/adwaita-icon-theme-feedstock/pull/18, and it does indeed fix the syntax error.

### DIFF
--- a/conda_smithy/templates/create_conda_build_artifacts.bat.tmpl
+++ b/conda_smithy/templates/create_conda_build_artifacts.bat.tmpl
@@ -54,8 +54,8 @@ if defined BLD_ARTIFACT_PREFIX (
         echo ##vso[task.setVariable variable=BLD_ARTIFACT_PATH]!BLD_ARTIFACT_PATH!
     )
     if "%CI%" == "github_actions" (
-        echo BLD_ARTIFACT_NAME=!BLD_ARTIFACT_NAME!>> %GITHUB_OUTPUT%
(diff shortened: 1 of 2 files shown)


hmaarrfk:
This bug was in fact created in https://github.com/conda-forge/conda-smithy/pull/1701 
and one can see from inspection that this restores the old syntax.

PR #1727, opened 2023-03-24
Add __str__ for NullUndefined

### LINKED ISSUE #1726: Python 3 regression: Undefined Jinja2 variables get rendered as empty string in linting

### Issue

For linting, undefined Jinja2 variables get rendered by `conda_smithy.utils.NullUndefined`. That class contains a `__unicode__` method that returns the name of the variable. This is useful to put a clear placeholder where variables will be filled by variants from `conda_build_config.yaml` during the actual build. However, `NullUndefined` doesn't overwrite the `__str__` method of Jinja's own `Undefined`, which returns an empty string.

In effect, linting in a Python 2 environment renders, e.g. `- {{ libjpeg }}` as `- libjpeg`, but in a Python 3 environment, we get `- ` which becomes `None` in the `requirements_section` dictionary.

### Installed packages

```shell
-
```

### Environment info

```shell
-
```

zklaus:
Checklist
* [x] Added a ``news`` entry

Fixes #1726

### DIFF
--- a/conda_smithy/utils.py
+++ b/conda_smithy/utils.py
@@ -64,6 +64,8 @@ class NullUndefined(jinja2.Undefined):
     def __unicode__(self):
         return self._undefined_name
 
+    __str__ = __unicode__
+
     def __getattr__(self, name):
         return "{}.{}".format(self, name)
 
--- a/conda_smithy/utils.py
+++ b/conda_smithy/utils.py
@@ -61,10 +61,8 @@ def tmp_directory():
 
 
 class NullUndefined(jinja2.Undefined):
-    def __unicode__(self):
+    def __str__(self):
         return self._undefined_name
-
-    __str__ = __unicode__
 
     def __getattr__(self, name):
         return "{}.{}".format(self, name)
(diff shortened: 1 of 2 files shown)


conda_smithy/utils.py
```diff
@@ -64,6 +64,8 @@ class NullUndefined(jinja2.Undefined):
     def __unicode__(self):
         return self._undefined_name
 
+    __str__ = __unicode__
```
jaimergp, line 67: I'd argue we invert this and use `__str__` as the method name, with the `__unicode__` alias if we want to keep backwards compatibility, but afaik we are not supporting py2k anymore so I'd just rename `__unicode__` to `__str__` and call it a day.

jaimergp:
Thanks @zklaus. I read the issue and it looks sensible to me. See my comment above too; oh, and we need a `news` item for this. Thanks again and sorry for the delay!

conda_smithy/utils.py
zklaus: Thanks, that makes sense to me. I agree that maintaining Python2 compatibility is not necessary and have adopted your suggestion.

jaimergp (approved):

zklaus:
Thanks, @jaimergp!

PR #1744, opened 2023-07-11
Provide GHA option to set maximum parallel jobs

jakirkham:
Provide a `conda-forge.yml` option to configure the number of GHA jobs that can be used in parallel. Default is to leave this unspecified. When set to a non-zero value, this is added to the appropriate field in the GHA workflow YAML.

<hr>

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1806,6 +1806,8 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
         },
         "github_actions": {
             "self_hosted": False,
+            # Set maximum parallel jobs
+            "max_parallel": None,
             # Toggle creating artifacts for conda build_artifacts dir
             "store_build_artifacts": False,
             "artifact_retention_days": 14,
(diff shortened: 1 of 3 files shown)


jakirkham:
@conda-forge/core, could you please review?

jaimergp (approved):

jaimergp:
LGTM. Only thing to notice is that this should be documented in `conda-forge.github.io` once available.

hmaarrfk:
is GHA = Azure? i don't think so right. if not, then where would this flag be used?

jakirkham:
Could you please clarify that question, Mark? Am a little unclear where Azure is coming up in the PR

The GHA logic is used by folks doing builds outside conda-forge on GHA

It's also possible we might use GHA more in the future in conda-forge (where this could be useful). Wouldn't be the first time we used a different CI provider 😉

hmaarrfk:
Got it. I understand the usecase.

jakirkham:
Thank you both! 🙏

PR #1754, opened 2023-07-29
Add option to cleanup GHA images

jakirkham:
Provide a similar option for GHA to cleanup images before builds as was done on Azure. As the same image is used in both cases, the same approach works in both cases.

Checklist
* [x] Added a ``news`` entry


jakirkham:
Please let me know if anything else is needed

cc @conda-forge/core

PR #1751, opened 2023-07-25
Add switch to make a swap file

mikemhenry:
I found that while working on https://github.com/conda-forge/staged-recipes/pull/22691 I needed to create a swap file for the build to complete since it exhausted the host memory.  

Checklist
* [ ] Added a ``news`` entry

PR #1756, opened 2023-08-01
Add schema validation and configuration for conda-forge config yaml

viniciusdc:
Checklist
* [x] Added a ``news`` entry

In this pull request, we propose a comprehensive refactor of the `configure_feedstock::_load_forge_config `function and its related dependencies. The primary objective is to replace the current reliance on a hardcoded dictionary with a more robust approach using Pydantic Models for the forge YAML configuration.

Key Changes:

- I've included a new schema file in the codebase, housing Pydantic models that represent various aspects of the ConfigModel, which is the core schema for the forge YAML. These models also include schemas for nested fields like bot, azure, and more.

- Created a separate function called `_read_forge_config` responsible for loading the YAML file itself and initializing the models with the respective configurations as well as handling the deep_merge dictionary section.
- Moved legacy checks to their dedicated function, `_legacy_compatibility_checks`.
- Ive added test cases to ensure the correctness and reliability of the new schema and code changes.
**Edited**
- Given the nature of  having `pydantic`as a runtime dependency, we adopted a mixed mode of `jsonschema` and the model to generate both a valid json schema mirroring the model, and versioned, `conda-forge.v2.json` and a default value that is dynamically generated by the model itself, `conda-forge.v2.yml`. This is used for type validation inside the new `_read_forge_config` function and to help future users who would like to access the schema by an API (which we can now do, by hosting the new json schema file)

### Why this now?

The rationale behind this refactor is to address issues that have surfaced over the years with the YAML content in forge configurations across numerous hosted feedstocks. These issues have resulted in minor exceptions and inconsistencies. Some problems were mitigated through hardcoded conditionals, while others slipped through, causing disruptions in the ecosystem.

The primary objectives of this refactor are as follows:

- By employing a Pydantic schema, we aim to significantly reduce unforeseen configuration errors from syntax errors or incorrect attributes in the forge YAML.

- This refactor will make it easier for future maintainers to introduce new settings, such as improving the list of available platforms or other heavily coupled configurations across the build. These changes can be seamlessly integrated into the configuration fields as validators, simplifying the management of previously scattered checks throughout the codebase. As well as deprecation warnings and migration checks can also be performed by the validators.

In summary, this pull request represents a substantial step towards enhancing the reliability and maintainability of the configuration handling in conda-smithy.

## What is not included in this PR?
- To keep the diff minimal in the first iteration, this PR only integrates the new schema (pydantic) and its artifacts (`*.json` and `.yml` files). In a subsequent PR, we can re-arrange some internal validations to be part of the generated schema, reducing some hard-coded logic from within the code.


viniciusdc:
Hi @jaimergp, this will be on draft this week as I fix some issues with the `Providers` enum logic (to accept and validate aliases). In the meantime, if you want to have a look at what is currently available, free free to do so :smile:

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+
+class CondaBuildConfig(BaseModel):
+    pkg_format: Union[int, str, None] = Field(
+        description="The package format for conda build. This can be one of 1, 2, or tar. The default is 2.",
+        default="2",
+    )
+
+    zstd_compression_level: int = Field(
+        default=16,
+        description="The compression level for the zstd compression algorithm for .conda artifacts. conda-forge uses a default value of 16 since its artifacts can be large.",
```
jaimergp, line 101:
```suggestion
        description="The compression level for the zstd compression algorithm for .conda artifacts. conda-forge uses a default value of 16 for a good compromise of performance and compression.",
```

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    def pkg_format_validator(cls, v):
+        if not v:
+            return v
+
+        valid_values = ["2", None]
+        if v not in valid_values:
+            raise ValueError(
+                f"Invalid value {v} for pkg_format. Valid values are {valid_values}"
+            )
+        return v
```
jaimergp, line 114: See the `Literal` alternative above.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+        description="The fallback image for Docker",
+        default="quay.io/condaforge/linux-anvil-comp7",
+    )
+
+    command: str = Field(
+        description="The command to run in Docker", default="bash"
+    )
+
+    interactive: bool = Field(
+        description="Whether to run Docker in interactive mode", default=None
```
jaimergp, line 134: The default should be a bool; I guess `false`?

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+        valid_values = ["2", None]
+        if v not in valid_values:
+            raise ValueError(
+                f"Invalid value {v} for pkg_format. Valid values are {valid_values}"
+            )
+        return v
+
+
+class CondaForgeDocker(BaseModel):
+    model_config = ConfigDict(extra="allow")
```
jaimergp, line 118: Do we want to allow extra keys? Why?

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+        description="AppVeyor CI settings. This is usually read-only and should not normally be manually modified. Tools like conda-smithy may modify this, as needed.",
+    )
+
+    azure: AzureConfig = AzureConfig()
+
+    bot: BotConfig = BotConfig()
+
+    build_platform: Annotated[
+        Union[Dict[Platforms, Platforms], None],
+        Field(
```
jaimergp, line 159: Why use this as `Annotated` instead of `Field` in value (as done elsewhere)?

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    build_with_mambabuild: Annotated[
+        Union[bool, None],
+        Field(
+            description="configures the conda-forge CI to run a debug build using the mamba solver. More information can be in the [mamba docs](https://conda-forge.org/docs/maintainer/maintainer_faq.html#mfaq-mamba-local)."
+        ),
+    ] = True
+
+    channel_priority: Annotated[
+        Union[ChannelPriorityConfig, None],
+        Field(
```
jaimergp, line 173: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    ] = None
+
+    channels: CondaForgeChannels = CondaForgeChannels()
+
+    choco: Annotated[
+        Union[List[str], None],
+        Field(
+            description="This parameter allows for conda-smithy to run chocoloatey installs on Windows when additional system packages are needed. This is a list of strings that represent package names and any additional parameters."
+        ),
+    ] = []
```
jaimergp, line 185: Mutable defaults should use `Field` with `default_factory`. Use `Field` in value, not Annotated.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+
+    circle: Dict[str, str] = Field(
+        default={},
```
jaimergp, line 188:
```suggestion
        default_factory=dict,
```

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+        description="Circle CI settings. This is usually read-only and should not normally be manually modified. Tools like conda-smithy may modify this, as needed.",
+    )
+
+    conda_build: Annotated[
+        Union[CondaBuildConfig, None],
+        Field(
```
jaimergp, line 194: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    conda_forge_output_validation: Annotated[
+        Union[bool, None],
+        Field(
+            description="This field must be set to True for feedstocks in the conda-forge GitHub organization. It enables the required feedstock artifact validation as described in [Output Validation and Feedstock Tokens](https://conda-forge.org/docs/maintainer/infrastructure.html#output-validation)."
+        ),
+    ] = False
+
+    docker: Annotated[
+        Union[CondaForgeDocker, None],
+        Field(
```
jaimergp, line 208: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+            description="This is a mapping for Docker-specific configuration options."
+        ),
+    ] = CondaForgeDocker()
+
+    github: Annotated[
+        Union[GithubConfig, None],
+        Field(
```
jaimergp, line 215: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    ] = GithubConfig(
+        user_or_org="conda-forge",
+        repo_name="",
+        branch_name="main",
+        tooling_branch_name="main",
+    )
+
+    idle_timeout_minutes: Annotated[
+        Union[int, None],
+        Field(
```
jaimergp, line 227: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    win: Annotated[
+        Union[WinConfig, None],
+        Field(
+            description="Windows-specific configuration options. This is largely an internal setting and should not normally be manually modified."
+        ),
+    ] = None
+
+    osx: Annotated[
+        Union[MacOsxConfig, None],
+        Field(
```
jaimergp, line 241: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    linux_aarch64: Annotated[
+        Union[Aarch64Config, None],
+        Field(
+            description="ARM-specific configuration options. This is largely an internal setting and should not normally be manually modified."
+        ),
+    ] = None
+
+    linux_ppc64le: Annotated[
+        Union[Ppc64leConfig, None],
+        Field(
```
jaimergp, line 262: Same. `Field` as value.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+            description="PPC-specific configuration options. This is largely an internal setting and should not normally be manually modified."
+        ),
+    ] = None
+
+    noarch_platforms: Annotated[
+        Union[List[Platforms], None],
+        Field(
```
jaimergp, line 269: Same. `Field` as value. And same elsewhere. You get the point :D

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+            description="Platforms on which to build noarch packages. The preferred default is a single build on linux_64."
+        ),
+    ] = [Platforms.linux_64.value]
+
+    os_version: Annotated[
+        Union[Dict[Platforms, str], None],
+        Field(
+            description="This key is used to set the OS versions for linux_* platforms. Valid entries map a linux platform and arch to either cos6 or cos7. Currently cos6 is the default for linux-64. All other linux architectures use CentOS 7."
+        ),
+    ] = {p.value: None for p in Platforms if p.value.startswith("linux")}
```
jaimergp, line 279: We should use `default_factory` I think.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+
+    provider: Annotated[
+        Union[
+            Dict[Platforms, Union[List[CIservices], CIservices, bool, None]],
+            None,
+        ],
+        Field(
+            description="The provider field is a mapping from build platform (not target platform) to CI service. It determines which service handles each build platform. If a desired build platform is not available with a selected provider (either natively or with emulation), the build will be disabled. Use the build_platform field to manually specify cross-compilation when no providers offer a desired build platform."
+        ),
+    ] = {
```
jaimergp, line 289: `default_factory` preferred.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+        description="The relative path to the recipe directory",
+        default="recipe",
+    )
+
+    remote_ci_setup: Annotated[
+        Union[List[str], None],
+        Field(
+            description="This option can be used to override the default conda-forge-ci-setup package. Can be given with ${url or channel_alias}::package_name, defaults to conda-forge channel_alias if no prefix is given."
+        ),
+    ] = ["conda-forge-ci-setup=3"]
```
jaimergp, line 314: `default_factory`

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+            description="Shell scripts used for builds or activation scripts can be linted with shellcheck. This option can be used to enable shellcheck and configure its behavior."
+        ),
+    ] = None
+
+    skip_render: Annotated[
+        Union[List[str], None],
+        Field(
+            description="This option specifies a list of files which conda smithy will skip rendering. The possible values can be a subset of .gitignore, .gitattributes, README.md, LICENSE.txt. The default value is an empty list [ ]"
+        ),
+    ] = []
```
jaimergp, line 328: `default_factory`

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+            "README.md",
+            "LICENSE.txt",
+            ".github/workflows",
+        ]
+        for i in v:
+            if i not in valid_values:
+                raise ValueError(
+                    f"Invalid value {i} for skip_render. Valid values are {valid_values}"
+                )
+        return v
```
jaimergp, line 347:
```suggestion
    def skip_render_validator(cls, values):
        if not values:
            return values

        valid_values = [
            ".gitignore",
            ".gitattributes",
            "README.md",
            "LICENSE.txt",
            ".github/workflows",
        ]
        for value in values:
            if value not in valid_values:
                raise ValueError(
                    f"Invalid value '{value}' for skip_render. Valid values are {valid_values}"
                )
        return values
```

Although I think this can be more easily computed with a set difference or something like that.

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+
+    test: Annotated[
+        Union[str, None],
+        Field(
+            description="This is used to configure on which platforms a recipe is tested. Default is all. Valid values are all, native_only, native_and_emulated, and emulated_only."
+        ),
+    ] = None
+
+    @field_validator("test")
+    def test_validator(cls, v):
```
jaimergp, line 371: Same as before. That said, this validator is not needed if the typing uses `Literal[<values allowed>]`.

conda_smithy/schema/platforms.py
jaimergp, line 1: Why is this file separate from `models`?

tests/test_condaforge_config_schema.py
```diff
@@ -0,0 +1,179 @@
+import pytest
+from pydantic import ValidationError
+from conda_smithy.schema.models import ConfigModel
+
+
+@pytest.fixture
+def default_config_data():
```
jaimergp, line 7: Are maintainers expected to keep this synced with the code or is this only for the transition period?

jaimergp (commented):
Thanks Vini! As you can see I have added a few comments, but most of them are repeated and then I stopped the spam. It all boils down to these two items:

- Default values that are mutable (e.g. lists or dicts) should use `default_factory` to return a fresh instance anytime. Otherwise all instances of the model share the same instance (defined at class definition time when the module was imported), which is then a nightmare. I don't think we would have concurrent use of more than one instance in the code, but you never know and it's good practice anyway.
- You are mixing usage of `Field` both as a value and `Annotated`. I am not familiar with `Annotated`; I think it might have performance benefits, but I do find it trickier to read. I prefer it as the value of the attribute, instead of in-annotation. If there are strong reasons for in-annotation, I am willing to listen, but what we can't do is to mix two styles because it's confusing :D

Let's fix these items and then we can keep iterating.

jaimergp:
Ah, and:

- we need to merge back the changes added in #1751 to resolve the conflicts. Please update accordingly 🙏 
- fix the CI to install pydantic as it's required now

conda_smithy/schema/models.py
```diff
@@ -0,0 +1,476 @@
...
+    )
+
+    targets: List[List[str]] = Field(
+        default=[["conda-forge", "main"]],
+        description="targets is a list of 2-lists, where the first element is the channel to push to and the second element is the label on that channel",
+    )
+
+
+class CondaBuildConfig(BaseModel):
+    pkg_format: Union[int, str, None] = Field(
```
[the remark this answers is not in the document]
viniciusdc: Thanks, I didn't find many docs on that right now under conda-forge, so I was not sure if I could assume if 1 was valid as well (even if that was the must possible scenario).

conda_smithy/schema/models.py
viniciusdc: The reason for it being none is that under the original method dictionary https://github.com/conda-forge/conda-smithy/blob/main/conda_smithy/configure_feedstock.py#L1716, those fields do not exist, and to have the same behavior you need to set those as none and under the dump method make sure its set to not render none values.

In any case those are quick fixes in both sides. tks

conda_smithy/schema/models.py
viniciusdc: Oh, my bad. Sometimes I like to use Anotated everywhere, and sometimes I use Field only. Though in this case, I started with Fields, then moved everything to Annotaded (I guess the reason was to be more compact, and found it more elegant); though as I was testing and changing the model, for quickness, I used Fields to test faster

conda_smithy/schema/models.py
viniciusdc: thanks, I guess I didn't fully understand what it was

conda_smithy/schema/platforms.py
viniciusdc: I thought it would be better to organize than a giant file with all child models and the main one, for readability basically

tests/test_condaforge_config_schema.py
viniciusdc: Just for now, to keep this as a test suit, I wanted to ensure the overall execution of the tests and methods would work. This can be removed as per the other tests we have with the builds -- though we should have some specific ones just for the models, not sure we would go with that though

conda_smithy/schema/models.py
viniciusdc: Update all to Fields :smile:

conda_smithy/schema/models.py
viniciusdc: removed and moved all fields to optional values, in case one of the supported config values is passed validation will occur, in case of extra variables in the config, right now the behavior is to ignore it (will not show up when dumping to dict/json)

tests/test_condaforge_config_schema.py
viniciusdc: removed

viniciusdc:
Hi @jaimergp, I updated the contents of this PR with the recent changes on `main` as well as committed some updates to the model following the presets we discussed.  I tested with the following feedstocks without any "visible" errors:
```
2dfatmic-feedstock         r-rebus-feedstock      tornado_sqlalchemy_login-feedstock
aws-cdk.aws-ssm-feedstock  subsurface-feedstock   types-futures-feedstock
paraprobe-feedstock        tar-feedstock          window_ops-feedstock
parsley-feedstock          tensortrade-feedstock  xproperty-feedstock
regro-neabari-feedstock    torchvision-feedstock
```
 though I would like to ask you if you could test with another feedstock to double check that the artifacts generated by `conda-smithy` are expected (this is, everything is normally running )

viniciusdc:
also, I am not sure what these errors are about: 
```
FAILED tests/test_configure_feedstock.py::test_upload_on_branch_azure - assert 'UPLOAD_ON_BRANCH="foo-branch"' in 'docker run --rm --privileged multiarch/qemu-user-static:register --reset --credential yes\nls /proc/sys/fs/binfmt_misc/\n'
FAILED tests/test_lint_recipe.py::Test_linter::test_maintainer_exists - AssertionError: 'A conda package with same name (fitsio) already exists.' not found in []
```
I don't think this related to this changes, so any extra insight would be great :smile:

viniciusdc:
Now that I noticed, some files in this PR are mixed with the recent additions. I will fix this at another date, for now just disregard it

bollwyvl:
I'm fully in favor of adding declarative structure to the various conda-forge files... if these result in checked in, at-rest, and published-on-the-web `.schema.json` files, with an `$id`, such that they can be consumed by downstream tools like language servers and non-python static checkers. This could be as simple as a test/build step that verifies that the thing that comes out matches what is checked in, with a step to generate it, but being able to look at these files over time is very helpful.

In the specific case of `pydantic`: this requires some additional effort in definitions, as it's easy to reach for python-specific validation approaches in "fat" models, but if a `pattern` or `format` can do the work, or at least provide a first pass, that helps a lot.

Somewhat off-topic, I'll say that I feel adding `pydantic` as a runtime dependency has been a net loss for some other tooling packages: its complexity is increasing and the version divide is fairly pronounced on `1.x`  vs `2.x`.

jaimergp:
Thanks @bollwyvl. Agreed, we'll try to make the most out of this PR. Publishing schemas is indeed planned. We still need to decide where though. Is the repo enough? We re currently rendering them as part of the source, I think.

Wrt the runtime dep, are you suggesting we only "render" the JSON schemas at build time / contribution time, and resort to validation using only the schema via `jsonschema` or similar? That could work, but I don't know if we can do all the validation like that. Maybe we do need to separate custom validation from the pydantic model for leaner runtime.

bollwyvl:
> Is the repo enough?

Well, schemas _change_ over time, but the content of a _schema version_ shouldn't.

So `rawgithubusercontent` does _work_ for a lot of cases:

```
https://github.com/conda-forge/conda-smithy/raw/v3.26.3/conda_smithy/static/schema/conda-forge-yaml.schema.json
```

A versioned ReadTheDocs would be relatively clean, and would also allow for consuming this content _without_ ReadTheCode, and would also provide a `latest`.

```
https://conda-smithy.readthedocs.org/en/3.26.3/schema/conda-forge-yaml.schema.json
```

While maybe the ideal would be:

```
https://conda-forge.org/schema/conda-forge-yaml/3.26.3/schema.json
```

> Wrt the runtime dep

Just pointing out that supporting complex schema with `pydantic 1.x` and `2.x` is challenging. Folks that invested heavily in `1.x` and its 150k footprint are partially left out to dry while some features are being filled out so they have to pin to `<2`, while `2.x` and its 2mb of `rust`-powered deps creates its own issues. Having to do _everything_ through custom validator hooks, instead of the declarative stuff `pydantic` provides... sorta defeats the purpose.

`jsonschema` _itself_ had a similar big breaking change, on a point release no less, and messed with its `[extras]` selector to get all the extra string format dependencies.

At any rate, as long as the schema is _available_ (and tested for the case of things it _can_ validate), _using_ the schema may be orthogonal to how it is defined.

viniciusdc:
Hello @bollwyvl, thank you for providing such valuable feedback! I want to address your point about having pydantic as a runtime dependency and the reasons behind it. Initially, we chose this approach for two main reasons: first, it ensured a short learning curve for future maintainers, and second, it allowed us to leverage integrations like pydantic_autodocs.

Regarding the transition from version 1.x to 2.x, I agree entirely that it was a significant change that caught many by surprise. To mitigate such challenges in the future, I'm considering pinning the dependency to >=2.x, which would enable us to use the latest pydantic code. This way, if any necessary breaking changes occur, the transition should be smoother, given our commitment to the 2.x range.

As for publishing the `json.schemas`, my initial thought was to include them within the conda-forge.io documentation, specifically within the class schema component. However, your example above provided an excellent illustration of how these schemas could be beneficial for general API consumption. Therefore, I propose incorporating them into the conda-smithy documentation, which would then be rendered by our CI pipeline.

I haven't introduced any new information beyond your suggestion, but I wanted to ensure we're on the same page. Do you have any other ideas or suggestions you believe would be interesting to explore?

Additionally, there's an alternative possibility to consider: using protobuf for the schema and then converting it to pydantic during the CI process. This approach could help us partially avoid the issues associated with runtime dependencies. However, it does appear to create a somewhat circular workflow... and break the original goal for easier maintainability.

bollwyvl:
> pinning the dependency to >=2.x,

See above: this would now make this useful package uninstallable next to [many packages](https://github.com/search?q=org%3Aconda-forge+%2Fpydantic+.*%3C2%2F+language%3AYAML+path%3A%2F%5Erecipe%5C%2F%2F&type=code).

> `pydantic_autodocs`

There is also [sphinx-jsonschema](https://pypi.org/project/sphinx-jsonschema). This approach is [taken](https://boa-build.readthedocs.io/en/latest/jsonschema_spec.html) by [boa](https://github.com/mamba-org/boa/pull/219) where pydantic is a build-time tool, while actual _validation_ is done by using [`jsonschema`](https://github.com/mamba-org/boa/blob/81dc74a9974ecd02494b300fb97a0a7b2f186afc/boa/core/validation.py#L21) directly.

> using protobuf for the schema and then converting it to pydantic

With no `protobuf` in the build stack at this time, I'd recommend keeping it that way, as it would ratchet the complexity up even further. It seems more likely for JSON schema knowledge to be broadly distributed than protobuf.

"Dumb" `pydantic` models (e.g. without methods) can also be generated from [JSON schema](https://github.com/koxudaxi/datamodel-code-generator/), though this process can lose specificity.

viniciusdc:
Hi @jaimergp, I've resolved the direct conflict that arose due to recent changes. However, I plan to conduct a thorough review of [conda_smithy/configure_feedstock.py](https://github.com/conda-forge/conda-smithy/pull/1756/files#diff-dc66a1980eec62725cfb8b6faea5b3c90dc7d1009d1229f5f55eecfb267a6677) to ensure I haven't overlooked any details during the conflict resolution process.

I'll also update the default values to eliminate the need for the extra `schema/defaults.yml` file, as suggested. I also intend to follow the advice provided by @bollwyvl to shift validation into the JSON schemas. The reasoning suggested above seems a plausible argument for this change. 

Let me know if there are any specific aspects you'd like me to focus on or if you have any other feedback or recommendations.

viniciusdc:
@jaimergp here's a sittuation where I would be in doubt, see the yaml specs are: 
```yaml
azure:
  ...
  settings_linux:
    pool:
      vmImage: ubuntu-latest
    swapfile_size: 0GiB
    timeoutInMinutes: 360
  settings_osx:
    pool:
      vmImage: macOS-11
    timeoutInMinutes: 360
```

Currnetly this is defined as the following:
```python
    settings_linux: AzureSelfHostedRunnerSettings = Field(
        default_factory=lambda: AzureSelfHostedRunnerSettings(),
        description="Linux-specific settings for self-hosted runners",
    )

    settings_osx: AzureSelfHostedRunnerSettings = Field(
        default_factory=lambda: AzureSelfHostedRunnerSettings(),
        description="OSX-specific settings for self-hosted runners",
    )
```
where:
```python
class AzureSelfHostedRunnerSettings(BaseModel):
    """This is the settings for self-hosted runners."""

    pool: Optional[Dict[str, str]] = Field(
        default_factory=lambda: {"vmImage": "ubuntu-latest"},
        description="The pool of self-hosted runners, e.g. 'vmImage': 'ubuntu-latest'",
    )

    swapfile_size: Optional[str] = Field(
        default="0GiB", description="Swapfile size in GiB"
    )

    timeoutInMinutes: Optional[int] = Field(
        default=360, description="Timeout in minutes for the job"
    )

    variables: Optional[Dict[str, str]] = Field(
        default_factory=dict, description="Variables"
    )
```

However, if I move forward, we will then have the other platforms receiving the swap memory entry, would this be acceptable or should the swap memory limit be something only available for linux?

conda_smithy/configure_feedstock.py
```diff
@@ -1872,45 +1753,109 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
...
+        config["conda_build_tool"] = (
+            "mambabuild" if config["build_with_mambabuild"] else "conda-build"
+        )
+    if file_config.get("conda_build_tool_deps"):
+        raise ValueError(
+            "Cannot set 'conda_build_tool_deps' directly. "
+            "Use 'conda_build_tool' instead."
+        )
+
+    # Validate the merged configuration against a JSON schema
```
jaimergp, line 1802: I feel we should validate with the json schema _before_ we run the custom validations.

conda_smithy/configure_feedstock.py
```diff
@@ -1938,22 +1883,6 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
...
-        )
-        config["conda_build_tool"] = (
-            "mambabuild" if config["build_with_mambabuild"] else "conda-build"
-        )
-    if file_config.get("conda_build_tool_deps"):
-        raise ValueError(
-            "Cannot set 'conda_build_tool_deps' directly. "
-            "Use 'conda_build_tool' instead."
-        )
     valid_build_tools = (
```
jaimergp, line 1886: It feels like this block (1886-1908) should be covered by the schema, not Python code. Can you double check?

conda_smithy/schema.py
```diff
@@ -0,0 +1,1310 @@
...
+    def __init__(
+        self,
+        enum_class: Union[Type[Enum], Type[EnumMeta]],
+        value_model: BaseModel,
+        model_name: str,
+    ):
+        self.enum_class = enum_class
+        self.value_model = value_model
+        self.model_name = model_name
+        self.generated_model = self._generate_model()
```
jaimergp, line 64: Instead of storing it, should we just call this method with `__call__`? Or even rename `_generate_model` to  `__call__`.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1310 @@
...
+
+#############################################
+######## Choices (Enum) definitions #########
+#############################################
+
+
+class CondaBuildTools(str, Enum):
+    conda_build = "conda-build"
+    conda_build_classic = "conda-build+classic"
+    conda_build_mamba = "conda-build+conda-libmamba-solver"
```
jaimergp, line 100:
```suggestion
    conda_build_libmamba = "conda-build+conda-libmamba-solver"
```

conda_smithy/schema.py
```diff
@@ -0,0 +1,1310 @@
...
+    mambabuild = "mambabuild"
+
+
+class CIservices(str, Enum):
+    azure = "azure"
+    circle = "circle"
+    travis = "travis"
+    appveyor = "appveyor"
+    default = "default"
```
jaimergp, line 109: It's a pity that we can't use `StrEnum` til 3.11. Otherwise we could use `auto()` for all the values.

conda_smithy/configure_feedstock.py
viniciusdc: Indeed, I was glancing over that so many times that I forgot, but our own values might obfuscate issues with the forge yaml present in the feedstock due to the merge of values. Thanks for pointing out

conda_smithy/schema.py
viniciusdc: Fixed, thanks 😄

conda_smithy/configure_feedstock.py
viniciusdc: done 😄

viniciusdc:
@jaimergp this is ready for a second review 😄

conda_smithy/schema.py
```diff
@@ -0,0 +1,1328 @@
...
+
+    #########################################
+    ##### Self-hosted runners settings ######
+    #########################################
+    settings_linux: AzureSelfHostedRunnerSettings = Field(
+        default_factory=lambda: AzureSelfHostedRunnerSettings(
+            swapfile_size="0GiB"
+        ),
+        description="Linux-specific settings for self-hosted runners",
+    )
```
isuruf, line 218: This is not specific to self-hosted runners. It supports all options in https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job?view=azure-pipelines which includes self-hosted runner implementation details.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1328 @@
...
+
+    store_build_artifacts: Optional[bool] = Field(
+        description="Whether to store build artifacts",
+        default=False,
+    )
+
+
+class PlatformUniqueConfig(BaseModel):
+    enabled: bool = Field(
```
jaimergp, line 302: I'm not sure whether we are using these keys anymore? e.g. [like this](https://github.com/conda-forge/libsimpleitk-feedstock/blob/983292ea3c2b439d30fbbbf617332cf0d43b21df/conda-forge.yml#L4-L7):

```yaml
linux_aarch64:
  enabled: false
linux_ppc64le:
  enabled: false
```

I thought that these days we just add them to the `build_platform` dict, and their presence in the dict is enough to enable them (the value would determine whether we emulate or cross-compile?).

Do you have any feedback @isuruf

isuruf: No, those are not user facing values. They are used internally in rendering.

isuruf:
I have couple of questions.

1. When adding a new field what are the steps to add them? I currently see lots of places
   - conda_smithy/data/conda-forge.v2.yml
   - conda_smithy/data/conda-forge.v2.json
   - conda_smithy/schema.py

2. For fields like build_platforms, we need to be able to add unknown platforms.
For eg: https://github.com/conda-forge/bzip2-feedstock/blob/main/conda-forge.yml#L5 is valid

jaimergp:
> When adding a new field what are the steps to add them?

The idea is `conda_smithy/schema.py`, and then the other outputs are regenerated via that script (JSON=schema, YAML=defaults). The YAML file is not necessary if there's a clearer way to initialize the dict.

> For fields like build_platforms, we need to be able to add unknown platforms.

Can't we just add them to the schema as needed? My concern is that typos leak in (e.g. `win_arn64`) unnoticed. At least a warning if the platform is not recognized?

conda_smithy/schema.py
```diff
@@ -0,0 +1,1172 @@
...
+    )
+
+    timeout_minutes: Optional[int] = Field(
+        default=360,
+        description="The maximum amount of time (in minutes) that a \
+            job can run before it is automatically canceled",
+    )
+
+    triggers: Optional[list] = Field(
+        default=[],
```
jaimergp, line 313:
```suggestion
        default_factory=list,
```

I think this is the canonical way of defining default mutables. Otherwise we get the same `function(value=[])` issue.

isuruf: No, see https://stackoverflow.com/a/63808835. Your suggestion results in `triggers: !!python/name:builtins.list ''` in `data/conda-forge.v2.yml`

jaimergp: Oh nice, they have finally[ documented it](https://github.com/pydantic/pydantic/pull/6122/files). It always felt like an implementation detail so it felt weird to rely on it. Glad to see it's actually official.

isuruf:
I ran this on all conda-forge.yml files from feedstocks and after the fixes, the remaining errors are genuine errors.

environment.yml
```diff
@@ -32,3 +32,6 @@ dependencies:
   - license-expression
   - libarchive
   - cirun
+  - pydantic >=1.8.1,<2
+  - jsonschema
+  - backports.strenum
```
isuruf, line 37: Going to use the newly forged package. :smile:

isuruf:
This is ready for some reviews.

conda_smithy/cli.py
```diff
@@ -596,13 +596,22 @@ def __call__(self, args):
                 all_good = False
                 print(
                     "{} has some lint:\n  {}".format(
-                        recipe, "\n  ".join(lints)
+                        recipe,
+                        "\n  ".join(
```
bollwyvl, line 600: could be:
```py
from textwrap import indent
# ...
indent("\n".join(lints), " " * 4)
```
et.c

conda_smithy/configure_feedstock.py
```diff
@@ -1814,138 +1826,13 @@ def _update_dict_within_dict(items, config):
...
-        "clone_depth": None,
-        # Specific channel for package can be given with
-        #     ${url or channel_alias}::package_name
-        # defaults to conda-forge channel_alias
-        "remote_ci_setup": ["conda-forge-ci-setup=4"],
-    }
+def _read_forge_config(forge_dir, forge_yml=None):
+    # Load default values from the conda-forge.yml file
+    yaml_defaults_file = (
+        Path(__file__).resolve().parent / "data" / "conda-forge.v2.yml"
```
bollwyvl, line 1832: perhaps hoist to semi-useful global:

```py
DEFAULT_FILE = Path(__file__).resolve().parent / "data/conda-forge.v2.yml"
```

conda_smithy/configure_feedstock.py
```diff
@@ -1958,76 +1845,42 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
...
-            raise ValueError(
-                "Cannot rerender with matrix in conda-forge.yml."
-                " Please migrate matrix to conda_build_config.yaml and try again."
-                " See https://github.com/conda-forge/conda-smithy/wiki/Release-Notes-3.0.0.rc1"
-                " for more info."
-            )
+    # Validate loaded configuration against a JSON schema.
+    validate_lints, validate_hints = validate_json_schema(file_config)
+    for err in validate_lints:
+        raise err
```
bollwyvl, line 1856: perhaps
```py
import sys
if sys.version_info < (3,11):
    from exceptiongroup import ExceptionGroup

# ...

if validate_lints:
    raise ExceptionGroup("lints", [*map(ValueError, validate_lints)])
```

conda_smithy/configure_feedstock.py
```diff
@@ -1958,76 +1845,42 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
...
-                "Setting docker image in conda-forge.yml is removed now."
-                " Use conda_build_config.yaml instead"
-            )
-        if file_config.get("channels"):
-            raise ValueError(
-                "Setting channels in conda-forge.yml is removed now."
-                " Use conda_build_config.yaml instead"
-            )
+    for hint in validate_hints:
+        print(hint.message)
```
bollwyvl, line 1859: `prints` might be generally more useful if wired up with:

```py
import logging
log = logging.getLogger(__name__)

# ...

    log.info(hint.message)

```

conda_smithy/data/conda-forge.v2.yml
```diff
@@ -0,0 +1,130 @@
+appveyor:
+  image: Visual Studio 2017
+azure:
+  force: false
+  max_parallel: 50
+  project_id: 84710dde-1620-425b-80d0-4cf5baca359d
```
bollwyvl, line 6: is this really a viable default?

conda_smithy/schema.py
```diff
@@ -0,0 +1,1217 @@
...
+    )
+
+
+_VALIDATOR_CLASS = get_validator_class()
+
+
+def validate_json_schema(config, schema_file: str = None):
+    # Validate the merged configuration against a JSON schema
+    json_schema_file = (
+        Path(__file__).resolve().parent / "data" / "conda-forge.v2.json"
```
bollwyvl, line 48: similarly to above, hoisting this to a global would be useful

conda_smithy/schema.py
```diff
@@ -0,0 +1,1217 @@
+# This model is also used for generating and automatic documentation for the conda-forge.yml file. The documentation is generated using sphinx and "pydantic-autodoc" extension. For an upstream preview of the documentation, see https://conda-forge.org/docs/maintainer/conda_forge_yml.html.
+
+import json
+from enum import Enum, EnumMeta
+from pathlib import Path
+from typing import Any, Dict, List, Literal, Optional, Type, Union
+
+import jsonschema
+import yaml
+from jsonschema import Draft7Validator, validators
```
bollwyvl, line 10: if this is `pydantic 2`, this should be `Draft202012Validator`.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1217 @@
...
+class ConfigModel(BaseModel):
+    """
+    This model describes in detail the top-level fields in  ``conda-forge.yml``.
+    General configuration options are described below within the ``Fields``
+    specifications. Additional examples are provided as part of the object
+    description. Values and options are subject to change, and will be
+    flagged as Deprecated as appropriate.
+
+    """
+
```
bollwyvl, line 504: Recommend adding some metadata which will make it out to the `.json`.

```py
class ConfigModel(BaseModel):
    class Config:
        schema_extra = {
            "$schema": "https://json-schema.org/draft/2020-12/schema#",
            "$id": "https://conda-forge.org/conda-smithy/conda-forge.v2.schema.json",
        }
```

environment.yml
```diff
@@ -32,3 +32,6 @@ dependencies:
   - license-expression
   - libarchive
   - cirun
+  - pydantic >=1.8.1,<2
```
bollwyvl, line 35: if there isn't a super-compelling reason, getting on the pydantic 2 bus probably makes sense.

tests/test_configure_feedstock.py
```diff
@@ -1,4 +1,6 @@
 import os
+
+from jsonschema import ValidationError
```
bollwyvl, line 3: With a shiny new JSON schema in hand, one can go _ape_ with stuff like https://github.com/python-jsonschema/hypothesis-jsonschema

conda_smithy/cli.py
jaimergp: Or even:

```python
print(f"{recipe} has some lint:", *lints, sep="\n  ")
```

conda_smithy/schema.py
bollwyvl: While, _technically_, `$id` is just an URI (identifier) and doesn't need to be locate-able. However, it is _really good_ if they are, once the time is taken to actually build them. A top-level subdirectory is a fine idea. This could hold other things, too, like:
- `repodata-patches` minigrammar
- `county-fair` content descriptions

Indeed, if the schema permitted `$schema`, and a user has a [YAML language server](https://github.com/redhat-developer/yaml-language-server) loaded up in their editor of choice, and something like:

```yaml
$schema: https://conda-forge.org/conda-smithy/conda-forge.v2.schema.json
bo # press <tab>
```

They might well see `bot`, and some helpful documentation. I haven't _yet_ wired all of these up into a statically-hostable docs "playground" site, but have been slowly making progress towards it.

conda_smithy/cli.py
isuruf: None of the above suggestions does what I do in this PR. i.e. all lint need two space indent and each multi-line lint need 4 spaces for the second line and after.

conda_smithy/data/conda-forge.v2.yml
isuruf: Yes, that's for conda-forge.

conda_smithy/schema.py
isuruf: Let's add that when there is a PR to the website.

isuruf:
Yes, it is.

conda_smithy/validate_schema.py
```diff
@@ -0,0 +1,55 @@
+import json
+from pathlib import Path
+
+from jsonschema import Draft202012Validator, validators
+from jsonschema.exceptions import ValidationError
+
+CONDA_FORGE_YAML_DEFAULTS_FILE = (
+    Path(__file__).resolve().parent / "data" / f"conda-forge.yml"
```
jaimergp, line 8:
```suggestion
    Path(__file__).resolve().parent / "data" / "conda-forge.yml"
```

conda_smithy/validate_schema.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+CONDA_FORGE_YAML_SCHEMA_FILE = (
+    Path(__file__).resolve().parent / "data" / f"conda-forge.json"
```
jaimergp, line 12:
```suggestion
    Path(__file__).resolve().parent / "data" / "conda-forge.json"
```

conda_smithy/validate_schema.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class DeprecatedFieldWarning(ValidationError):
+    pass
+
+
+def deprecated_validator(validator, value, instance, schema):
+    if value and instance is not None:
+        yield DeprecatedFieldWarning(
+            f"'{schema['title']}' is deprecated.\n" + schema["description"]
```
jaimergp, line 23:
```suggestion
            f"'{schema['title']}' is deprecated.\n{schema['description']}"
```

news/1756-schema.rst
```diff
@@ -0,0 +1,25 @@
+**Added:**
+
+* New conda-forge pydantic schema for ``conda-forge.yaml``, includes a generate default forge yaml and json schema dynamically from the pydantic model
```
jaimergp, line 3:
```suggestion
* New JSON schema for ``conda-forge.yaml``. A Pydantic model is used to dynamically generate both a YAML document with the default values and the JSON schema itself.
```

news/1756-schema.rst
```diff
@@ -0,0 +1,25 @@
...
+* Included ``jsonschema``and ``pydantic`` as dependencies into the ``environment.yml``
```
jaimergp, line 4:
```suggestion
* Included ``jsonschema`` and ``pydantic`` as dependencies into the ``environment.yml``. (#1756)
```

news/1756-schema.rst
```diff
@@ -0,0 +1,25 @@
...
+
+**Changed:**
+
+* Included extra ``jsonschema`` validation for forge yaml, under ``configure_feedstock``
```
jaimergp, line 8:
```suggestion
* Included extra ``jsonschema`` validation for conda-forge.yaml, under ``configure_feedstock``. (#1756)
```

news/1756-schema.rst
```diff
@@ -0,0 +1,25 @@
...
+* Moved legacy checks of old_file and providers into a new  auxiliary ``_legacy_compatibility_checks`` function
```
jaimergp, line 9:
```suggestion
* Moved legacy checks of old_file and providers into a new auxiliary ``_legacy_compatibility_checks`` function. (#1756)
```

tests/test_configure_feedstock.py
```diff
@@ -899,12 +900,12 @@ def test_conda_build_tools(config_yaml):
         fp.write(unmodified)
         fp.write("conda_build_tool: does-not-exist")
 
-    with pytest.raises(AssertionError):
+    with pytest.raises(configure_feedstock.ExceptionGroup):
```
jaimergp, line 903: I wonder if we should raise a subclass of the ExceptionGroup for better specificity. If not, I prefer we import from the package directly here.

tests/test_configure_feedstock.py
```diff
@@ -899,12 +900,12 @@ def test_conda_build_tools(config_yaml):
...
         assert load_forge_config()
 
 
 def test_remote_ci_setup(config_yaml):
-    load_forge_config = lambda: cnfgr_fdstk._load_forge_config(  # noqa
+    load_forge_config = lambda: configure_feedstock._load_forge_config(  # noqa
```
jaimergp, line 908: Thanks for doing this :D

tests/test_condaforge_config_schema.py
```diff
@@ -0,0 +1,82 @@
...
+        "github": {
+            "branch_name": "main",
+            "tooling_branch_name": "main",
+        },
+        "conda_forge_output_validation": True,
+        "conda_build": {
+            "pkg_format": 2,
+        },
+    }
+    # Extra value should be ignored
```
jaimergp, line 79: Why ignored? If there's a typo in a field name (e.g. `conda_bild`), we should emit a warning. Is that feasible? Or are we simply not testing that here.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1165 @@
...
+    # will run vanilla conda-build, with the classic solver
+    "conda-build+classic",
+    # will run vanilla conda-build, with libmamba solver
+    "conda-build+conda-libmamba-solver",
+    # will run 'conda mambabuild', as provided by boa
+    "mambabuild",
+]
+
+
+class CIservices(str, Enum):
```
jaimergp, line 119: We are importing `StrEnum` above, but then using `str, Enum` here? I guess we should use here as well (and in the other occurrences).

conda_smithy/schema.py
```diff
@@ -0,0 +1,1165 @@
...
+    )
+
+    variables: Optional[Dict[str, str]] = Field(
+        default_factory=dict, description="Variables"
+    )
+
+
+class AzureConfig(BaseModel):
+    """
+    This dictates the behavior of the Azure Pipelines CI service. It is a sub-mapping for Azure-specific configuration options. For more information and some variables specifications, see the [Azure Pipelines schema reference documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/?view=azure-pipelines).
```
jaimergp, line 184: Some line wrapping would be nice here.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1165 @@
...
+
+    automerge: Optional[Union[bool, BotConfigAutoMergeChoice]] = Field(
+        False,
+        description="Automatically merge PRs if possible",
+    )
+
+    check_solvable: Optional[bool] = Field(
+        None,
+        description="Open PRs only if resulting environment is solvable.",
+        exclude=True,  # Will not be rendered in the model dump
```
jaimergp, line 340: Why do we want to exclude these from the model? 🤔

conda_smithy/schema.py
```diff
@@ -0,0 +1,1165 @@
...
+    interactive: Optional[Union[bool, Nullable]] = Field(
+        description="Whether to run Docker in interactive mode",
+        default=None,
+        exclude=True,  # Will not be rendered in the model dump
+    )
+
+    #########################################
+    #### Deprecated Docker configuration ####
+    #########################################
+    image: Optional[Union[str, Nullable]] = Field(
```
jaimergp, line 409: How does the deprecation hint validation work here? I don't see any metadata here that annotates this as deprecated? Or is the list maintained elsewhere?

jaimergp: Ah, I saw in `build_with_mambabuild` that we have a `deprecated=True` kwarg in Field? That seems to be pydantic v1 only, and apparently we need `json_schema_extra={'deprecated': True}` now.

jaimergp: More info https://github.com/pydantic/pydantic/issues/2255

conda_smithy/schema.py
isuruf, line 184:
```suggestion
    This dictates the behavior of the Azure Pipelines CI service. It is a sub-mapping for
    Azure-specific configuration options. For more information and some variables
    specifications, see the [Azure Pipelines schema reference documentation](
    https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/?view=azure-pipelines).
```

tests/test_configure_feedstock.py
isuruf: > If not, I prefer we import from the package directly here.

I didn't do that because then we'd have to duplicate the logic of where ExceptionGroup comes from. i.e. check if it is a builtin or use the backport.

conda_smithy/schema.py
isuruf: Marked with `deprecated=True` kwarg in Field. We have our own custom validator for deprecated fields.

conda_smithy/schema.py
isuruf: Done in [2cf8f73](https://github.com/conda-forge/conda-smithy/pull/1756/commits/2cf8f7378b25ec6da3f6a03964bbcf1551932b38)

tests/test_condaforge_config_schema.py
isuruf: The PR here only checked the values of known keys. I'm not sure how to check for unknown keys.

jaimergp: Via `extra=forbid` ([see docs](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra)) and catch, maybe?

conda_smithy/schema.py
jaimergp: Will this interfere with the JSON schema output? Maybe it's worth either switching to `json_schema_extra={'deprecated': True}` as the marker or having the schema dumper do that from `deprecated=True`, since pydantic v2 won't do that for us. This way the IDE tools can use the [`deprecated`](https://json-schema.org/draft/2020-12/json-schema-validation#name-deprecated) marker to style keys and throw warnings. That said, it's not yet ready on VS Code, so feel free to ignore. I'm just concern we forget about this courtesy detail in the future :)

isuruf: No, the schema dumper sets `deprecated=True` correctly.
For eg:
```json
"compiler_stack": {
      "anyOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ],
      "default": "comp7",
      "deprecated": true,
      "description": "\n        Compiler stack environment variable. This is used to specify the compiler\n        stack to use for builds. Deprecated.\n\n        .. code-block:: yaml\n\n            compiler_stack: comp7\n        ",
      "title": "Compiler Stack"
    },
```

conda_smithy/schema.py
```diff
@@ -0,0 +1,1208 @@
...
+if __name__ == "__main__":
+    from pathlib import Path
+
+    # This is used to generate the model dump for conda-smithy internal use
+    # and for documentation purposes.
+
+    model = ConfigModel()
+
+    with CONDA_FORGE_YAML_SCHEMA_FILE.open(mode="w+") as f:
+        f.write(model.schema_json(indent=2))
```
bollwyvl, line 1204:
```
The method "schema_json" in class "BaseModel" is deprecated
  The `schema_json` method is deprecated; use `model_json_schema` and json.dumps (
```

conda_smithy/schema.py
```diff
@@ -0,0 +1,1208 @@
...
+        f.write("\n")
+
+    with CONDA_FORGE_YAML_DEFAULTS_FILE.open(mode="w+") as f:
+        f.write(yaml.dump(model.dict(), indent=2))
```
bollwyvl, line 1208:
```
The method "dict" in class "BaseModel" is deprecated
  The `dict` method is deprecated; use `model_dump` instead.
```

isuruf: Fixed in c4fda5b

conda_smithy/schema.py
isuruf: Fixed in c4fda5b

isuruf:
@bollwyvl, I've removed the custom enum code here. Have a look now.

jaimergp:
I've merged and fixed the conflicts. @isuruf, assuming @bollwyvl has no further comments, is there anything missing here? I'd like to move this forward.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1141 @@
...
+    build_with_mambabuild: Optional[bool] = Field(
+        default=True,
+        exclude=True,
+        deprecated=True,
+        description="""
+        build_with_mambabuild is deprecated, use conda_build_tool instead. Configures the conda-forge CI to run a debug build using the ``mamba`` solver.
+        More information can be found in the
+        `mamba docs <https://conda-forge.org/docs/maintainer/maintainer_faq.html#mfaq-mamba-local>`__.
+
+        .. code-block:: yaml
```
jaimergp, line 1107: Hm, now that the site uses Markdown, we might need to update these RST blobs. I can do that, just noting it here for reference.

isuruf:
pre-commit.ci autofix

isuruf:
No, this should be ready to go.

jaimergp:
I took a look at the current docs and the rendered docs (using this PR schema) and there are some differences:

image

Mostly, the top-level OS keys (linux, linux_*, osx, win) and `channels` are gone. Do we want to add them to the schema as deprecated keys?

isuruf:
No, they are internal variables and is not affected by any user input.

jaimergp:
@conda-forge/core, please take a look in this PR before we merge. This introduces a JSON schema (built with Pydantic) for conda-forge.yml, allowing us to run linting checks and auto-generate documentation (see https://github.com/conda-forge/conda-forge.github.io/pull/2095).

I'd like to have this merged sometime this week, so if there are no comments, I'll do it on Thursday.

beckermr:
pre-commit.ci autofix

beckermr (approved):
LGTM!

We should make sure to clarify and/or update the docs here over time.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1240 @@
...
+    ROS_DISTRO = "rosdistro"
+    # CRAN = "CRAN"
+    # GITHUB = "Github"
+    # INCREMENT_ALPHA_RAW_URL = "IncrementAlphaRawURL"
+    # LIBRARIES_IO = "LibrariesIO"
+    # NPM = "NPM"
+    # NVIDIA = "NVIDIA"
+    # PYPI = "PyPI"
+    # RAW_URL = "RawURL"
+    # ROS_DISTRO = "ROSDistro"
```
ytausch, line 103: What about these comments? I think currently the bot only supports the lowercase version

conda_smithy/schema.py
```diff
@@ -0,0 +1,1240 @@
...
+
+
+class CIservices(StrEnum):
+    azure = "azure"
+    circle = "circle"
+    travis = "travis"
+    appveyor = "appveyor"
+    github_actions = "github_actions"
+    drone = "drone"
+    woodpecker = "woodpecker"
```
ytausch, line 55: I want to point out that these two services are currently [undocumented](https://conda-forge.org/docs/maintainer/conda_forge_yml/#provider) and also not used by any feedstock, as far as I can tell.

conda_smithy/schema.py
```diff
@@ -0,0 +1,1240 @@
...
+    """
+
+    model_config: ConfigDict = ConfigDict(extra="forbid")
+
+    random_fraction_to_keep: Optional[float] = Field(
+        None,
+        description="Fraction of versions to keep for frequently updated packages",
+    )
+
+    exclude: Optional[List[Union[str, float]]] = Field(
```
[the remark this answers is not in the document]
jaimergp: That list is awesome, thanks!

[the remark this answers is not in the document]
jaimergp: @isuruf, any thoughts here? Since it's only three feedstocks, I think it's ok to be strict here and only allow `str` (same with `abi_migration_branches`).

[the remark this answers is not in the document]
isuruf: The errors are a bit cryptic here if I remember correctly. That's why I added these. I don't mind changing it. Let's send PRs to those 3 feedstocks.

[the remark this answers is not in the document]
jaimergp: Added:

- https://github.com/conda-forge/cuda-python-feedstock/pull/77
- https://github.com/conda-forge/wgpu-native-feedstock/pull/18
- https://github.com/conda-forge/aws-c-common-feedstock/pull/94

conda_smithy/schema.py
```diff
@@ -0,0 +1,1240 @@
...
+    circle = "circle"
+    travis = "travis"
+    appveyor = "appveyor"
+    github_actions = "github_actions"
+    drone = "drone"
+    woodpecker = "woodpecker"
+    default = "default"
+    emulated = "emulated"
+    native = "native"
+    disable = "None"
```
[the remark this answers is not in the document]
jaimergp: See:

- https://github.com/conda-forge/vector-classes-feedstock/issues/9
- https://github.com/conda-forge/root-feedstock/issues/238
- https://github.com/conda-forge/lbapcommon-feedstock/issues/64

The other mentioned repos don't show that in `main`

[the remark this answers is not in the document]
henryiii: Chris builds and uploads those locally on an M1 machine from what I remember.

jaimergp (approved):
I think this is ready to go!

Thanks everyone involved, this has been one of the good ones :P 

I'll merge later today or tomorrow in case there are any last minute comments.

jakirkham:
This was neat to see: https://github.com/conda-forge/coverage-feedstock/pull/112#issuecomment-1998610696

Very descriptive 🤩

bollwyvl:
Yes, have started getting those, looking great.

Some notes:

- it doesn't tell _which_ file this is all about
  - it could include a link to that exact file in the PR branch, such that a maintainer could click to edit
- as the output is for _humans_, it seems like rendering `title`, `description`, etc. as Markdown would be valuable.
  - however, as it was written (or dumped to) JSON Schema, it's would possible to generate a canonical link to the generated documentation for that schema item, once available
  - similarly, the raw repr that gets dumped could be a table
    - or, indeed, the whole thing could be a table
  - it may be worth putting these inside `details` tags

Now all we gotta do is schematize the other user-edited `.yml` :)

jaimergp:
Thanks @bollwyvl, I'll take your comment and open a new issue for tracking! Thanks!

0xbe7a:
https://github.com/pydantic/pydantic/pull/8237 has been merged and will probably be released as part of `pydantic` 2.7 and will make the custom deprecation validator obsolete.

conda_smithy/schema.py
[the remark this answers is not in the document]
chrisburr: I only build `root` locally and use CFEP-03. The use of that key must have been copied from somewhere at some point, or maybe I just made it up 😆

PR #1757, opened 2023-08-07
ENH add service feedstocks

beckermr:
Checklist
* [x] Added a ``news`` entry

This needs some testing live to make sure it works.


beckermr:
Tested and this works fine. Note that our webservices won't be able to rerender feedstocks that use this option without us supplying a special github token. We have the hooks to do this.

beckermr:

ocefpaf (approved):
Makes sense. All of those should have priority.

PR #1760, opened 2023-09-01
Use `python-build` to create `sdist`

jakirkham:
Checklist
* [x] Added a ``news`` entry

Comes from https://github.com/conda-forge/conda-smithy-feedstock/pull/259#discussion_r1312426400

### DIFF
--- a/news/use_python-build_sdist.rst
+++ b/news/use_python-build_sdist.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Use ``python-build`` to create ``sdist`` #1760
+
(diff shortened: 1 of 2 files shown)


jaimergp:
Do we need to add `build` (or `python-build`) to the runtime requirements metadata?

xhochy:
> Do we need to add `build` (or `python-build`) to the runtime requirements metadata?

No, this is only a build-time requirement. If we would have an `environment.yml`, we would need to add it there, but currently we don't have anything in this regard. I can make a follow-up PR to add this.

jakirkham:
Thank you both! 🙏

PR #1761, opened 2023-09-06
Unify Windows batch logic

jaimergp:
Checklist
* [x] Added a ``news`` entry

We had duplicate logic for Azure and Github Actions, which was resulting in duplicated efforts in PRs like #1732, and divergence of features/bugfixes (GHA didn't have `python=3.10`, `UPLOAD_ON_BRANCH` or the `%TEMP%` fixes for `anaconda upload ...`).

In this PR I am hoping to unify both in a single script, which should also get us closer to having `build-locally.py` features for Windows. The downside is that the Azure build steps will be now a single one, which is not terrible and can be worked with logging groups like we do in Unix if needed.

One thing to pay special attention to is the error handling on CMD. It can be terribly confusing depending on too many edge cases (see #1704, I will probably restart work on that after this one lands).

Tests:

- For Azure: https://github.com/conda-forge/dav1d-feedstock/pull/10
- For GHA: https://github.com/jaimergp-feedstocks/dav1d-feedstock/pull/1


conda_smithy/configure_feedstock.py
```diff
@@ -941,7 +941,7 @@ def _get_build_setup_line(forge_dir, platform, forge_config):
         if platform == "win":
             build_setup += textwrap.dedent(
                 """\
-                run_conda_forge_build_setup
+                CALL run_conda_forge_build_setup
```
jaimergp, line 944: Without using `CALL`, this batch script coming from `conda-forge-ci-setup` might (will?) exit early with errorlevel = 0, misreporting success without even running the recipe.

jaimergp:
@conda-forge/core - ready for review!

beckermr (approved):
I know nothing about windows but it it works, LGTM!

PR #1763, opened 2023-09-08
Windows: Catch all non-zero exit codes (negative values too)

jaimergp:
Checklist
* [x] Added a ``news`` entry

Supersedes #1704. The formula `if errorlevel 1` actually means "if the exit code is >=1", so it doesn't catch `-1` and other negative exit codes. This can happen in conda-build when it segfaults during the linkage analysis steps, for example.

Exit codes in CMD are tricky and there are many gotchas. Since we can afford to enable delayed expansion, this is relatively robust. See [this SO answer](https://stackoverflow.com/questions/34936240/batch-goto-loses-errorlevel/34937706#34937706) for (way ) more info if you are curious.

### DIFF
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -20,7 +20,7 @@ call activate base
 
 :: Provision the necessary dependencies to build the recipe later
 echo Installing dependencies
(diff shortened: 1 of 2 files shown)


xhochy:
We should have this as a linter message for `bld.bat`, too. We use the previous `if errorlevel` statements in nearly all windows builds.

Issue #1767, opened 2023-09-19
Prepare release 3.26


jaimergp:
### Tasks

- [x] Ensure required PRs have been merged
- [x] Test `main` on some feedstocks:
  - [x] `dav1d`: https://github.com/conda-forge/dav1d-feedstock/pull/10
    - [X] [Default conda-forge.yml (install with mamba, build with mambabuild)](https://github.com/conda-forge/dav1d-feedstock/pull/10/commits/65f90444f622681ee343148089c186f2c3288ed6)
    - [X] [Install and build use libmamba](https://github.com/conda-forge/dav1d-feedstock/pull/10/commits/c966b1b425838e87915f5af86b9809a7e0114278) + [fix](https://github.com/conda-forge/dav1d-feedstock/pull/10/commits/e96bcc00f4f56695cc5919a2b72b02fb6e2504ae) via https://github.com/conda-forge/conda-smithy/pull/1768. This doesn't solve well on macOS. We'll probably need conda-libmamba-solver 23.9.0 for this. Not a blocker!
- [X] Run `rever -c x.y.z` and check everything works
- [x] Run `rever` and [push changelog](https://github.com/conda-forge/conda-smithy/commit/ff4f6cae87bebbf00a54e5fb4c4fbf73490d5bc2) and [release](https://github.com/conda-forge/conda-smithy/releases/tag/v3.26.0)
- [x] Release package on [feedstock](https://github.com/conda-forge/conda-smithy-feedstock) - https://github.com/conda-forge/conda-smithy-feedstock/pull/260

cc @xhochy

xhochy:
@jaimergp The current workflow is that `rever` will directly do the release directly and I don't see an option to do it as PR.

Running the local check for me works:

```
% rever -c 3.26.0
All CLI and import requirements met for authors activity
activity-check:authors:Checked activity authors
No checks needed for changelog activity
No checks needed for tag activity
All CLI and import requirements met for push_tag activity
Initialized empty Git repository in /Users/uwe/Development/conda-smithy/rever/git-have-push-perm/.git/
Switched to a new branch '__rever__'
`.pre-commit-config.yaml` config file not found. Skipping `pre-commit`.
[__rever__ (root-commit) b47c598] Checking rever permissions
Enumerating objects: 2, done.
Counting objects: 100% (2/2), done.
Writing objects: 100% (2/2), 565 bytes | 565.00 KiB/s, done.
[...]

PR #1768, opened 2023-09-19
Follow up for `conda_install_tool` and `conda_solver` settings

jaimergp:
Checklist
* [x] Added a ``news`` entry

Round of fixes found while tesrting for: #1767

- `conda update` is receiving specs with version constraints (`conda update conda-forge-ci-setup=3` fails). `update` and `uninstall` expect name only specs.
- Checks for `conda_solver` didn't check for value, only presence
- Configure `solver` here instead of `conda-forge-ci-setup` so the first `conda` calls respect the configured value (conda-forge-ci-setup arrives "too late" for those initial calls). In Docker we use `conda config ...` because it's a Docker instance and it's disposable. In macOS and Windows we set an env var so the changes are only visible to that script and do not leak into local configurations when run outside CI.
- Ensure strict channel priority on Linux and macOS to prevent packages from `defaults` leaking in the CI.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -29,6 +29,7 @@ import conda_build.utils
 import conda_build.variants
 import conda_build.conda_interface
 import conda_build.render
+from conda.models.match_spec import MatchSpec
 
 from copy import deepcopy
 
@@ -1956,6 +1957,7 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
         f"Invalid conda_build_tool: {config['conda_build_tool']}. "
         f"Valid values are: {valid_build_tools}."
     )
+    # NOTE: Currently assuming these dependencies are name-only (no version constraints)
     if config["conda_build_tool"] == "mambabuild":
         config["conda_build_tool_deps"] = "conda-build boa"
     elif config["conda_build_tool"] == "conda-build+conda-libmamba-solver":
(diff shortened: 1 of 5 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -2029,9 +2032,8 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
     if config["provider"]["linux_s390x"] in {"default", "native"}:
         config["provider"]["linux_s390x"] = ["travis"]
 
-    config["remote_ci_setup"] = _santize_remote_ci_setup(
-        config["remote_ci_setup"]
-    )
+    config["remote_ci_setup"] = _santize_remote_ci_setup(config["remote_ci_setup"])
+    config["remote_ci_setup_names"] = [MatchSpec(pkg).name for pkg in config["remote_ci_setup"]]
```
0xbe7a:
```suggestion
    config["remote_ci_setup"] = _santize_remote_ci_setup(
        config["remote_ci_setup"]
    )
    config["remote_ci_setup_names"] = [
        MatchSpec(pkg).name for pkg in config["remote_ci_setup"]
    ]
```

0xbe7a:
https://www.githubstatus.com/ "Investigating - We are investigating reports of degraded performance for Pull Requests"

jaimergp:
@xhochy - ready for review. This will _not_ enable libmamba as default since I have identified a couple of frictions in `dav1d-feedstock` (we'll need conda-libmamba-solver 23.9.0, coming out soon). I don't want to block you any longer so this is enough to make things work. My suggestion is to leave these new keys undocumented and after I've done some more testing then we can roll it out.

xhochy (approved):

xhochy:
Thanks!

PR #1770, opened 2023-09-20
fix how `FEEDSTOCK_NAME` is set on Windows

jaimergp:
Checklist
* [x] Added a ``news`` entry

A bug introduced in #1761, where `FEEDSTOCK_NAME` on Azure was being set to the whole thing (e.g. `conda-forge/package-feedstock`) instead of just `package-feedstock).

This is causing failing uploads on Windows.


xhochy (approved):

jaimergp:
Yes, sorry about the downtime, it was a regression in the Windows scripts 😅. You can usually check conda-forge/status (which powers conda-forge.org/status) in addition to gitter/Element for extra info (sometimes). 🙏

PR #1772, opened 2023-09-21
take linter hints from pinning repo

### LINKED ISSUE #1740: ENH: Define package hints in a place that doesn't need a new linter version for updates
In #1734, I added a dictionary `specific_hints`, which built upon existing nudges (e.g. about matplotlib-base), and is meant to be both performant and extensible:

> As a [side note](https://github.com/conda-forge/conda-smithy/pull/1734#discussion_r1186575235), I made this a dictionary (rather than a chain of cases) because I expect this list to grow substantially and wanted to keep performance snappy (checking dict membership is fast in python).
> 
> The anticipated growth is not mainly due to renamings, but because I think it would be useful for all kinds of output-hygiene (e.g. packages from archived/abandoned/superseded feedstocks, like all the ancient "homegrown" cuda feedstocks).

 @beckermr noted:

> It'd be nice to store this metadata somewhere else so that we can add to it without making a smithy release. This can happen in another PR though.

I agree that this would be good. Opening this issue to discuss where we should keep that information, and how the linter would interface with it.

h-vetinari:
Closes #1740

I hope our runner is on 3.11 already? 🤔 
Otherwise I'll need to add a lib to read toml.


h-vetinari:
> I hope our runner is on 3.11 already?

haha:
```
============================= test session starts ==============================
platform linux -- Python 3.8.17, pytest-7.4.2, pluggy-1.3.0
```

Aaaaanyway.

beckermr (approved):
Lgtm! We may want to update the recipe in the repo for the new rep if needed.

h-vetinari:
> Lgtm! We may want to update the recipe in the repo for the new dep[?] if needed.

I specifically chose tomli because it's already part of the deps here already:
https://github.com/conda-forge/conda-smithy/blob/8ffa0d8f73e69dbcfe19c97c92b05e77187cad6a/environment.yml#L12

Though I guess I just now see there's a separate in-repo [recipe](https://github.com/conda-forge/conda-smithy/tree/main/conda_smithy.recipe). Is that actually still in use? At least it's not in sync with https://github.com/conda-forge/conda-smithy-feedstock/blob/main/recipe/meta.yaml...

beckermr:
The feedstock is the correct one. We should sync them or delete this one.

beckermr (approved):

h-vetinari:
@beckermr just saw that tomli is already only specified as a conditionally dependency:
https://github.com/conda-forge/conda-smithy/blob/8ffa0d8f73e69dbcfe19c97c92b05e77187cad6a/pyproject.toml#L5

I think it makes sense to fall back to tomli only if we're on py<3.11. Will push a commit shortly, please let me know if you don't think that's a good idea.

beckermr (approved):

jaimergp (commented):
I wonder if we should put this logic in `regro/conda-forge-metadata` too to avoid hardcoding URLs directly. And then call `conda_forge_metadata.linter.get_hints()` here or something like that.

h-vetinari:
> I wonder if we should put this logic in `regro/conda-forge-metadata` to avoid hardcoding URLs directly. And then call `conda_forge_metadata.linter.get_hints()` here or something like that.

Sounds fine by me. I haven't worked with that repo yet, but I tried my hand at https://github.com/regro/conda-forge-metadata/pull/17

beckermr:
Yeah I'd rather not link bot-specific stuff to the main org stuff just yet. I don't want to have to maintain the api deps

beckermr:
I do think we should do this eventually, just not right now. Or at least we need to move the package to conda-forge on GitHub.

h-vetinari:
> Yeah I'd rather not link bot-specific stuff to the main org stuff just yet. I don't want to have to maintain the api deps

I'll let you two figure this one out - I don't mind either way (though I can see how the config in the pinning repo falls under "conda-forge-metadata", regardless of whether that's under regro or the conda-forge org).

beckermr:
Yeah let's merge this for now.

h-vetinari:
With Jaime's 👍 on Matthew's comment, it seems we are all agreed, so I'll merge this now.

PR #1781, opened 2023-10-05
ensure cache_dir is a Path when XDG_CACHE_HOME is set

minrk:
Checklist
* [x] Added a ``news`` entry

When XDG_CACHE_HOME was set, a string was returned, but consumers of get_cache_dir assume it returns a Path object.


xhochy (approved):

xhochy:
I'll release this ASAP.

PR #1810, opened 2023-11-28
Add Mambaforge and macos SDK files to feedstock gitignore file

hadim:
Checklist
* [ ] Added a ``news`` entry

I found those two files to be annoying when debugging a feedstock locally on macOS.

### DIFF
--- a/conda_smithy/feedstock_content/.gitignore
+++ b/conda_smithy/feedstock_content/.gitignore
@@ -1,3 +1,6 @@
 *.pyc
 
 build_artifacts
+
+Mambaforge-MacOSX-*.sh
+MacOSX*.sdk.tar.xz


hmaarrfk:
I feel like https://github.com/conda-forge/conda-smithy/pull/1413 will get merged in first :/

hadim:
> I feel like #1413 will get merged in first :/

oh ok - I wasn't aware of that PR. Well, let me know if I should close here or not. I would argue that if you think #1413 will take time before being merged, maybe that one could be a quick fix waiting for it.

hmaarrfk:
you can leave it open. I think it is a good idea and we should merge this if that other PR takes too long.

Its a rather "complicated" `.gitignore` file which is why it has taken so long to get approved.

hmaarrfk (approved):
I would like to merge this if https://github.com/conda-forge/conda-smithy/pull/1413 dargs on to Dec 15.

Issue #1782, opened 2023-10-07
Non-deterministic behavior when rerendering pytensor-suite-feedstock


maresb:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

See https://github.com/conda-forge/pytensor-suite-feedstock/issues/54 for technical details.

In short, since several months, running `conda-smithy rerender` randomly skips `.ci-support` files which shouldn't be skipped. It took me several hours with the debugger to get an idea about what's going on.

My rough understanding is that conda-smithy calls conda-build in order to determine whether to render a group like osx-arm64. Associated to this group is a set of Python versions ranging from 3.8 to 3.11. The recipe contains `skip: true  # [py<39]`. In evaluating whether or not to skip `osx-arm64`, a Python version is selected at random. Whenever 3.8 comes up, it skips `osx-arm64` and causes the corresponding `.ci-support` files to go missing.

### Installed packages

```shell
conda-build 3.25.0, conda-smithy 3.26.2, and conda-forge-pinning 2023.10.07.18.27.00
```

### Environment info

```shell
N/A
```

mbargull:
Hi @maresb,
the `conda-smithy` part of this issue is likely fixed via gh-1815 in a future `conda-smithy` release.
Another part of the issue which makes it much more likely to occur for `pytensor-suite-feedstock` is due to its usage of the `py` selector in the top-level `skip` line of the recipe in combination with `python` being only used for the `outputs`.
I've suggested a change to better handle this case in `conda-build` via https://github.com/conda/conda-build/pull/5139 .
See that pull request's description for more details on why it happens and also how you could work around it for now in the recipe.

maresb:
Thanks @mbargull for helping me solve #54. That was the functional problem I was trying to solve.

Please feel free to close this issue whenever you see fit. Thanks so much for tracking down these really subtle problems!!!

PR #1815, opened 2023-12-20
Fix squished zip key ordering

mbargull:
Checklist
* [x] Added a ``news`` entry

Fixes gh-1459
Fixes the `conda-smithy` part affecting gh-1782 (that issue needs further adjustments to the recipe and/or `conda-build`).

gh-1459 is an old issue such that the initial discussion refers to old code.
A recent analysis regarding current code is at https://github.com/conda-forge/conda-smithy/issues/1459#issuecomment-1858846653 .
This PR adds the suggested change from https://github.com/conda-forge/conda-smithy/issues/1459#issuecomment-1858848004 .

With that change the issue from https://github.com/conda-forge/python-feedstock/pull/656#discussion_r1428500644 has not been reproducible for me (for a selection of `PYTHONHASHSEED` values).
Similarly the issue from https://github.com/conda-forge/pytensor-suite-feedstock/issues/54 reported here at gh-1782 does not reproduce for me (when fixing top-level `py` selector without `{{ python }}` usage in that recipe).

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -391,6 +391,16 @@ def _collapse_subpackage_variants(
 
     for k, v in squished_used_variants.items():
         if k in all_used_vars:
+            input_variant = squished_input_variants.get(k)
+            if input_variant and isinstance(input_variant, (tuple, list)):
+                # Ensure we retain the input order to avoid mismatched entries
+                # with the preserve_top_level_loops overrides below.
+                # NOTE: "pin_run_as_build" (dict of dict) order is not handled.
+                # "tuple" below only used for uniform list/tuple/str comparison.
+                v_set = set(map(tuple, v))
+                v = type(input_variant)(
+                    v_i for v_i in input_variant if tuple(v_i) in v_set
+                )
             used_key_values[k] = v
 
     for k in preserve_top_level_loops:
--- a/news/fix-squished-zip_key-ordering.rst
+++ b/news/fix-squished-zip_key-ordering.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fix randomly mismatched zipped variant keys. (#1459 and #1782 via #1815)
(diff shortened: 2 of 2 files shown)


beckermr (approved):

jakirkham:
Is there a test we could include? Admittedly this may be tricky

beckermr:
@mbargull said on the call just now that the code above is not correct and needs changes

beckermr:
A test here is going to be pretty hard to make.

conda_smithy/configure_feedstock.py
```diff
@@ -391,6 +391,16 @@ def _collapse_subpackage_variants(
...
+            input_variant = squished_input_variants.get(k)
+            if input_variant and isinstance(input_variant, (tuple, list)):
+                # Ensure we retain the input order to avoid mismatched entries
+                # with the preserve_top_level_loops overrides below.
+                # NOTE: "pin_run_as_build" (dict of dict) order is not handled.
+                # "tuple" below only used for uniform list/tuple/str comparison.
+                v_set = set(map(tuple, v))
+                v = type(input_variant)(
+                    v_i for v_i in input_variant if tuple(v_i) in v_set
+                )
```
mbargull, line 403: I opted to no handle `pin_run_as_build` since its order *should* not matter, i.e., we'd had to do special handling for that case here with no further benefit.
I added at least a note so we are aware that it's intentionally omitted.

The variant entries should be of type `str`, so a straight `v_set = set(v)` would be possible.
We do, however, also have `zip_keys` which is a `list` of `list`/`tuple`.
To handle potential `tuple`-to-`list` comparisons (and to get hashable types for `list` entries) this just casts everything to `tuple`s.

Feel free to amend any of this in case you'd want more explicit handling.

mbargull:
> Happy to merge assuming you've tested that this fixes the issue on a feedstock.

Tried this on `python-feedstock` with some `PYTHONHASHSEED` values that make it break without these changes.

> A test here is going to be pretty hard to make.

Yeah, at least we have https://github.com/conda-forge/conda-smithy/blob/v3.30.4/tests/test_cli.py#L90 (which was actually tripped with the first pass I had with this).

mbargull:
> I'll wait a day or so and then merge if things are looking good. Thank you!

Makes sense. Thanks for reviewing!

Issue #1840, opened 2024-02-16
ENH: correctly rerender CI configs with new stdlib jinja-function


h-vetinari:
This is in the context of https://github.com/regro/cf-scripts/pull/2135 (see links therein).

I tried rendering a recipe with `- {{ stdlib("c") }}`, and unsurprisingly it failed. Adding the following patch 
```diff
--- a/conda_smithy/utils.py
+++ b/conda_smithy/utils.py
@@ -93,6 +93,7 @@ def render_meta_yaml(text):
     env.globals.update(
         dict(
             compiler=lambda x: x + "_compiler_stub",
+            stdlib=lambda x: x + "_stdlib_stub",
             pin_subpackage=stub_subpackage_pin,
             pin_compatible=stub_compatible_pin,
             cdt=lambda *args, **kwargs: "cdt_stub",
```
eliminates the failure (edit: this turned into #1846), but while trying to write tests for this, I got stuck with understanding where in smithy the actual content of the CBC gets generated, and more importantly, where the values are taken from.

For example, I cannot find how gcc gets inserted as the c_compiler - it's neither specified in the code, nor injected by the fixture or the test, nor does it read the global pinning AFAICT.

I tried to follow what's happening in 
https://github.com/conda-forge/conda-smithy/blob/main/conda_smithy/configure_feedstock.py and others, and have followed several (apparently) false leads already, so I thought I'd ask if someone can tell me where to look.

To be clear, what I want to ensure is that a line like `- {{ stdlib("c") }}` ends up inserting the correct `c_stdlib{,_version}: ...` into the generated ci-config files (pretty much exactly like we do for the compilers), c.f. also https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5499.

CC @beckermr

beckermr:
I wonder if conda build has a default somewhere.

h-vetinari:
Sure, that's a probable answer where "gcc" comes from. But luckily we shouldn't need conda-build (or changes to it) here - we only need to pick up settings for `c_stdlib{,_version}` from a local (or global) `conda_build_config.yaml`, and populate the CI configs correctly.

So far, even if I extend the `config_yaml` fixture to contain the information from https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5499, it doesn't get inserted into the CI yamls, and I don't know where I need to look to make that happen...

beckermr:
[...]

PR #1844, opened 2024-02-20
Change default build tool to conda-build again

mbargull:
Checklist
* [x] Added a ``news`` entry

I'd like us to get the [fixes from `conda-build=24.1.2` in which also affect `conda-build=3.28.4`](https://github.com/conda/conda-build/blob/24.1.2/CHANGELOG.md#2412-2024-02-15) which we are currently limited to due to https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/657 / https://github.com/mamba-org/boa/issues/392 .
If we change to `conda-build` by default, we can avoid the constraint needed for `boa=0.16.0`.
(We should also put out a `boa=0.16.1` for feedstocks that have `conda-mambabuild` set explicitly, of course.)

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1946,7 +1946,7 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
         "conda_forge_output_validation": False,
         "private_upload": False,
         "secrets": [],
-        "conda_build_tool": "mambabuild",
+        "conda_build_tool": "conda-build",
         "conda_install_tool": "mamba",
         "conda_solver": "libmamba",
         # feedstock checkout git clone depth, None means keep default, 0 means no limit
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1954,7 +1954,7 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
         # Specific channel for package can be given with
         #     ${url or channel_alias}::package_name
         # defaults to conda-forge channel_alias
-        "remote_ci_setup": ["conda-forge-ci-setup=4"],
+        "remote_ci_setup": ["conda-forge-ci-setup=4", "conda-build>=24.1"],
     }
 
     if forge_yml is None:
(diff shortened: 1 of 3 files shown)


beckermr (changes requested):
I am for this, but we need to put a pin somewhere to ensure the latest version is used. There were some bugs which is why re rolled this back last I checked.

mbargull:
> I am for this, but we need to put a pin somewhere to ensure the latest version is used.

I guess shove a `conda-build>=24.1` into `remote_ci_setup` so that the user can override it when needed, right?

> There were some bugs which is why re rolled this back last I checked.

https://github.com/conda-forge/conda-smithy/issues/1805
Yep, there were some oddities with `conda-build` using legacy code paths such that it didn't play well with `conda-libmamba-solver`; we've added some changes to `conda`/`conda-libmamba-solver` in November to fix that.

beckermr:
Yes I think so or we can have it as a constraint on our ci setup package.

h-vetinari:
> If we change to `conda-build` by default, we can avoid the constraint needed for `boa=0.16.0`.

Not against this change, but just wanted to note that we could also (try to) backport the fix for https://github.com/mamba-org/boa/issues/392 to unblock that constraint.

jakirkham:
> Yes I think so or we can have it as a constraint on our ci setup package.

Are we guaranteed to get the latest `conda-forge-ci-setup` when installing?

If not, maybe adding it here is easiest (though no objections to doing both)

beckermr:
Yeah let's add it here. Good idea!

mbargull:
Conceptually, I think `conda-forge-ci-setup` would be a nicer place for it (yes, we'd need to then have `conda-forge-ci-setup=4,>={version with lower bound on conda-build}` in `remote_ci_setup` here then), but for now, I'd also guess it's fine to add it here (done in the latest commit) and would postpone conceptual discussion currently.

beckermr:
I agree. What's nice here is that people can override easily.

beckermr (approved):

xhochy (approved):

jakirkham:
Thanks all! 🙏

Should we make a new release ( https://github.com/conda-forge/conda-smithy/issues/1845 )?

mbargull:
Thanks for the reviews!

h-vetinari:
> Should we make a new release ( #1845 )?

Give me a short time to make a minimal PR for #1840 please (just adding the comprehension for stdlib jinjas), so that - ideally - we won't have to release separately for that.

PR #1849, opened 2024-02-24
check up-to-date with not broken versions only

### LINKED ISSUE #1848: Soften up-to-date check
Currently if `conda-smithy` is not on the latest version during a rerender, it will raise an error

However if the latest `conda-smithy` is marked broken, this behavior persists without taking into account the latest non-broken version is older than the latest version naively checked

This can leave [bot codes that rerender]( https://github.com/search?q=org%3Aconda-forge+conda+smithy+rerender+language%3APython&type=code&l=Python ) in a dysfunctional state. Meaning core needs to hunt down and fix these lines at the same time a broken conda-smithy emerges (the exact same time we may not have bandwidth to fix it for the same reason)

Some options we might consider here:
* Teach the up-to-date check about broken versions
* Soften the up-to-date check to a warning unless a flag is passed to create the error
* ...?

Please feel free to suggest more

jaimergp:
Checklist
* [ ] Added a ``news`` entry

Closes #1848

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2188,17 +2188,17 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
     return config
 
 
-def get_most_recent_version(name):
+def get_most_recent_version(name, only_not_broken=False):
     from conda_build.conda_interface import VersionOrder
 
     request = requests.get(
         "https://api.anaconda.org/package/conda-forge/" + name
     )
     request.raise_for_status()
-
-    pkg = max(
-        request.json()["files"], key=lambda x: VersionOrder(x["version"])
-    )
+    files = request.json()["files"]
+    if only_not_broken:
+        files = [f for f in files if "broken" not in f.get("labels", ())]
+    pkg = max(files, key=lambda x: VersionOrder(x['version']))
 
     PackageRecord = namedtuple("PackageRecord", ["name", "version", "url"])
     return PackageRecord(name, pkg["version"], "https:" + pkg["download_url"])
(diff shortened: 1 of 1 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -2188,17 +2188,17 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
     return config
 
 
-def get_most_recent_version(name):
+def get_most_recent_version(name, only_not_broken=False):
```
xhochy, line 2191: Shouldn't this be the unconditional default, i.e. not add another parameter?

xhochy (changes requested):

conda_smithy/configure_feedstock.py
jakirkham: At a minimum, think the default value should be `True`. Choosing to include broken packages should be opt-in (as it is with installing them)

jakirkham:
Thanks Jaime! 🙏

Agree with others on the line above

Otherwise think this is a welcome improvement. Thanks again for putting it together 🙂

conda_smithy/configure_feedstock.py
jaimergp: Ok, changed the kwarg to `include_broken=False`.

xhochy (approved):

h-vetinari:
The `include_broken` kwarg doesn't seem to be wired up anywhere? It doesn't hurt for now (so fine to release 3.31.1), but I think we should either remove the kwarg completely, or ensure it can be used from CLI.

jakirkham:
Should we move that to an issue so we don't forget?

xhochy:
> Should we move that to an issue so we don't forget?

Done.

PR #1856, opened 2024-03-08
feat: add pixi as a way to install package

nichmor:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/templates/README.md.tmpl
+++ b/conda_smithy/templates/README.md.tmpl
@@ -228,10 +228,23 @@ or with `mamba`:
 mamba install {{ ' '.join(outputs) }}
 ```
 
+or with [`pixi`](https://pixi.sh):
+
+```
+pixi global install --channel {{ channel_with_label }} {{ ' '.join(outputs) }}
+```
+
+
 It is possible to list all of the versions of `{{ outputs[0] }}` available on your platform with `conda`:
 
 ```
 conda search {{ outputs[0] }} --channel {{ channel_with_label }}
+```
+
+or with [`pixi`](https://pixi.sh):
+
+```
+pixi search --channel {{ channel_with_label }} {{ outputs[0] }}
 ```
 
 or with `mamba`:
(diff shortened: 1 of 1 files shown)


conda_smithy/templates/README.md.tmpl
```diff
@@ -228,12 +228,31 @@ or with `mamba`:
 mamba install {{ ' '.join(outputs) }}
 ```
 
+or with [`pixi`](https://pixi.sh):
```
xhochy, line 231: I would like to omit having the global install as the most prominent pixi version. The vast majority of packages on conda-forge are modules where a global install doesn't make sense. 

Personally, I would not mention the global option at all.

xhochy (changes requested):

conda_smithy/templates/README.md.tmpl
nichmor: I agree with you
Thanks for the tip! I've removed it

isuruf (commented):
I don't know if we should keep adding all these tools in every feedstock. I think this would be better to have it as a note in the conda-forge website. (Btw, we should remove mamba too)

baszalmstra:
Lowering the barrier to be able to try a particular piece of software seems very vital to me. We want people to use this! I feel it's very useful to have a shortcut on how to use a package with your favorite package manager in the feedstock itself.

But I do agree it clutters up the readme a little. Perhaps we can add all package managers are foldable sections like this:

### How to use
<details>
  <summary>With Conda</summary>

  ```
  conda env create ..
  ```
</details>

<details>
  <summary>With Mamba</summary>

  ```
  mamba install ..
  ```
</details>

<details>
  <summary>With Pixi</summary>
  
  ```
  pixi add ..
  ```
</details>

Removing mamba seems strange to me, in that sense we could also argue to remove the conda commands. 🤷

conda_smithy/templates/README.md.tmpl
```diff
@@ -228,12 +228,25 @@ or with `mamba`:
 mamba install {{ ' '.join(outputs) }}
 ```
 
+If you are using a [`pixi` project](https://pixi.sh/latest/basic_usage/):
+
+```
+pixi add {{ ' '.join(outputs) }}
+```
+
```
pavelzw, line 236:
```suggestion
```

baszalmstra:
What is the status of this PR? What do we need to change?

PR #1888, opened 2024-03-23
add test for correctly resolving stdlib

### LINKED ISSUE #1840: ENH: correctly rerender CI configs with new stdlib jinja-function

h-vetinari:
Closes #1840, ~needs conda-build 24.3; tested locally with artefacts from https://github.com/conda-forge/conda-build-feedstock/pull/220~

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -509,6 +509,8 @@ def _collapse_subpackage_variants(
         "macos_machine",
         "channel_sources",
         "channel_targets",
+        "c_stdlib",
+        "c_stdlib_version",
         "docker_image",
         "build_number_decrement",
         # The following keys are required for some of our aarch64 builds
--- a/news/1888-ensure-stdlib-renders-correctly.rst
+++ b/news/1888-ensure-stdlib-renders-correctly.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Ensure new ``{{ stdlib("c") }}`` correctly populates CI config. (#1840 via #1888)
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -214,7 +214,9 @@ about:
     home: home
     """
         )
-    with open(os.path.join(config_yaml, "recipe", "stdlib_config.yaml"), "w") as f:
+    with open(
+        os.path.join(config_yaml, "recipe", "stdlib_config.yaml"), "w"
+    ) as f:
         f.write(
             """\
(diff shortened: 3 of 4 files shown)


h-vetinari:
I'm actually surprised that this passes with conda-build 24.1.2; I thought that https://github.com/conda/conda-build/pull/5195 would have been necessary, but I guess by adding `c_stdlib{,_version}` to `always_keep_keys`, we're not actually dependent on that PR after all... CC @mbargull

beckermr (approved):

jakirkham:
This is great! Thanks Axel 🙏

mbargull:
> I'm actually surprised that this passes with conda-build 24.1.2; I thought that [conda/conda-build#5195](https://github.com/conda/conda-build/pull/5195) would have been necessary, but I guess by adding `c_stdlib{,_version}` to `always_keep_keys`, we're not actually dependent on that PR after all... CC @mbargull

Well, sure, without https://github.com/conda/conda-build/pull/5195, the key is not recognized as used; if you override the logic with the (`conda-smithy`-specific) `always_keep_keys` thing, you don't depend on it being recognized, of course.

h-vetinari:
Yeah, I didn't know the inner workings of conda-{build,smithy} well enough to understand that, sorry...

mbargull:
Ah, no need to apologize!
We needed the fix in `conda-build` anyway since we have support uses outside of `conda-smithy rerender` workflows ;).

PR #1891, opened 2024-03-25
add "disabled" bot inspection to schema

ytausch:
Checklist
* [x] Added a ``news`` entry

https://github.com/regro/cf-scripts/pull/2274 added support for the `disabled` bot inspection choice for `conda-forge.yml`, which is added with this PR. Also, `True` was never supported as an inspection option.


ytausch:
Reference: https://github.com/regro/cf-scripts/issues/2272

beckermr (approved):
Thank you!

PR #1897, opened 2024-03-28
small fixes

ytausch:
Checklist
* N/A

- Since Python 3, we no longer subclass `object`
- `logger.warn` is deprecated
- fix typo


beckermr (changes requested):
Please add a news file.

PR #1909, opened 2024-04-18
provide linter hints about adding stdlib, removing outdated ways to do it

h-vetinari:
As discussed in core call today; I wanted to add this into `run_conda_forge_specific` originally (where the other hint logic from https://github.com/conda-forge/conda-smithy/pull/1772 lives[^1], but it's a pain to run that in testing because `run_conda_forge_specific` requires a `GH_TOKEN`. If desired, I can still move it there, especially now that the tests have been debugged (and assuming CI here has a valid token). 🤷  

[^1]: which we cannot reuse because it currently only works for plain package names, no patterns, versions etc.

Note: I'd also want to wait before ~merging~ releasing this until we have updated the stdlib docs in the knowledge base, to avoid people getting confused.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -845,6 +845,47 @@ def lintify_meta_yaml(
             "[here]( https://conda-forge.org/docs/maintainer/adding_pkgs.html#spdx-identifiers-and-expressions )."
         )
 
+    # stdlib-related hints
+    build_reqs = requirements_section.get("build") or []
+    run_reqs = requirements_section.get("run") or []
+    constraints = requirements_section.get("run_constrained") or []
+
+    stdlib_hint = (
+        "This recipe is using a compiler, which now requires adding a build "
+        'dependence on `{{ stdlib("c") }}` as well. For further details, please '
+        "see https://github.com/conda-forge/conda-forge.github.io/issues/2102."
+    )
+    pat_compiler_stub = re.compile(
+        "(m2w64_)?(c|cxx|fortran|rust)_compiler_stub"
+    )
+    has_compiler = any(pat_compiler_stub.match(rq) for rq in build_reqs)
+    if has_compiler and "c_stdlib_stub" not in build_reqs:
+        if stdlib_hint not in hints:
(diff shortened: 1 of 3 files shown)


beckermr:
I didn't follow why a token was needed or w/e with the tests. However, LGTM!

beckermr (approved):

h-vetinari:
> I didn't follow why a token was needed or w/e with the tests. 

Calling `run_conda_forge_specific` immediately runs into
https://github.com/conda-forge/conda-smithy/blob/4006d83ee032f3440dcbb9baf0f5a46619c3586e/conda_smithy/lint_recipe.py#L851-L852

(and faking that token is no use because it is soon after used to fetch info from GH).

Hence testing `run_conda_forge_specific` needs a valid token, whereas "just" calling `lintify_forge_yaml` does not run into this constraint.

beckermr:
We could guard those lines and calls to enable things to be run easily without tokens.

h-vetinari:
> We could guard those lines and calls to enable things to be run easily without tokens.

Sure. Not sure though if it's worth the effort? For now it seems to work OK as is and doesn't seem tooooo ugly? (then again, perhaps I also haven't understood where exactly the split between `run_conda_forge_specific` and `lintify_forge_yaml` should be). How about a follow-up issue?

h-vetinari:
OK, now that we have some basic documentation for stdlib up, let's put this one in. :)

ytausch:
@h-vetinari @beckermr I suggest not introducing any new code that uses `io.open`. In Python 3, the preferred way should be just `open`, which is equivalent to `io.open`.

jakirkham:
Maybe someone can add an appropriate `ruff` configuration to the linters here

h-vetinari:
> I suggest not introducing any new code that uses `io.open`. In Python 3, the preferred way should be just `open`, which is equivalent to `io.open`.

Yeah, completely fine by me, I just happened to overlook this while C&P-ing from pre-existing smithy code.

ytausch:
> Maybe someone can add an appropriate `ruff` configuration to the linters here

Done. See #1919

Issue #1911, opened 2024-04-20
BUG: smithy ignores keys in migrator when affected by zipping logic


h-vetinari:
In https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790, I'm updating some values that are part of a `zip_keys`, as well as another value (`channel_sources`). As far as I can tell (perhaps related to the overhaul for additional zip-keys in #1764 & #1769), as soon as the variant algebra finds some zip_keys, it will "forget" the other keys being updated.

There is another adjacent issue related to https://github.com/conda-forge/conda-smithy/pull/1752, in the sense that
https://github.com/conda-forge/conda-smithy/blob/8a2eba45e40e53abd2e1744475942e133cf6bffe/conda_smithy/configure_feedstock.py#L976-L978
will only ever consider the first entry of migrated_combined_variant_spec, however, it is not a given that this only has one entry.

I have a draft fix, but opening this issue so that we can discuss the right approach (if it turns out I'm on the wrong track).

CC @beckermr @jaimergp @0xbe7a

h-vetinari:
@conda-forge/core - this is a minimal fix that we need for the numpy 2.0 migration AFAICT (even aside from the question of what to do with python 3.8). Could someone comment here or review #1912 please?

h-vetinari:
So the issue here was that `variant_key_add` doesn't know how to determine the "higher of the two" between `conda-forge` and `conda-forge/label/numpy_rc,conda-forge`. This can be solved by providing an `ordering:` to the migrator.

This now completely circumvents `additional_zip_keys` (which still may have some open corner cases, but if they exist and we hit them, we can open separate issues).

In the meantime, I'm closing this.

PR #1912, opened 2024-04-20
Do not ignore zip-unrelated keys in migrators

h-vetinari:
Tentative fix for #1911, needed for https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790 (AFAIU), tested in https://github.com/conda-forge/scipy-feedstock/pull/274 and https://github.com/conda-forge/pythran-feedstock/pull/84

### DIFF
--- a/conda_smithy/variant_algebra.py
+++ b/conda_smithy/variant_algebra.py
@@ -19,7 +19,7 @@ from conda_build.utils import ensure_list
 import conda_build.variants as variants
 from conda.exports import VersionOrder
 from conda_build.config import Config
-from functools import partial
+from functools import partial, reduce
 
 
 from typing import Any, Dict, List, Optional, Union
@@ -168,6 +168,14 @@ def op_variant_key_add(v1: dict, v2: dict):
         result.setdefault("__additional_zip_keys_default_values", {})[
             additional_key
         ] = result[additional_key][0]
+
+    all_zips = result.get("zip_keys", [])
+    all_keys_in_zips = reduce(lambda x, y: set(x) | set(y), all_zips, set())
+    # update result with all other keys specified in the migrator,
+    # which aren't part of the migrator specs or zipped keys
+    already_handled = {"__migrator"} | all_keys_in_zips | newly_added_zip_keys
+    for key in set(v2.keys()) - already_handled:
+        result[key] = v2[key]
 
     for pkey_ind, pkey_val in enumerate(v2[primary_key]):
         # object is present already, ignore everything
--- a/conda_smithy/variant_algebra.py
+++ b/conda_smithy/variant_algebra.py
@@ -174,7 +174,7 @@ def op_variant_key_add(v1: dict, v2: dict):
     # update result with all other keys specified in the migrator,
     # which aren't part of the migrator specs or zipped keys
     already_handled = {"__migrator"} | all_keys_in_zips | newly_added_zip_keys
-    for key in set(v2.keys()) - already_handled:
+    for key in (set(result.keys()) & set(v2.keys())) - already_handled:
         result[key] = v2[key]
 
     for pkey_ind, pkey_val in enumerate(v2[primary_key]):
--- a/news/1912-zip_key-hole.rst
+++ b/news/1912-zip_key-hole.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Do not ignore keys updated by migrator if they do not appear in a `zip_keys:` (#1912)
+
+**Security:**
+
+* <news item>
--- a/tests/test_variant_algebra.py
+++ b/tests/test_variant_algebra.py
@@ -531,7 +531,7 @@ def test_multiple_key_add_migration():
     assert "numpy" not in res3
 
 
-def test_other_keys_in_additional_zip_migrator():
+def test_update_other_keys_not_in_zip_keys():
     """Test updating keys in migrators not related to zips still get updated."""
     # based on global pinning
     base = parse_variant(
--- a/tests/test_variant_algebra.py
+++ b/tests/test_variant_algebra.py
@@ -531,6 +531,153 @@ def test_multiple_key_add_migration():
     assert "numpy" not in res3
 
 
+def test_other_keys_in_additional_zip_migrator():
+    """Test updating keys in migrators not related to zips still get updated."""
+    # based on global pinning
+    base = parse_variant(
+        dedent(
+            """
+            python:
+              - 3.8.* *_cpython
+              - 3.9.* *_cpython
+              - 3.10.* *_cpython
+              - 3.11.* *_cpython
+            python_impl:
+              - cpython
+              - cpython
+              - cpython
+              - cpython
+            numpy:
+              - 1.22
+              - 1.22
+              - 1.22
+              - 1.23
+            # not part of zip
+            channel_sources:
+              - conda-forge
+            zip_keys:
+              -
+                - python
+                - numpy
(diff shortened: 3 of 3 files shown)


beckermr:
Can you add a test recipe for this one? Without a simplified example, I won't be able to understand what changed. Further, this issue seems easy to miss and like a thing we don't want to regress later.

h-vetinari:
I've now added a test based on the current state of https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790, which works fine for feedstocks on py>=39. For py38 we'll need some further fixes, but this one will be necessary in any case.

> Without a simplified example, I won't be able to understand what changed.

I've reproduced the current situation (i.e. global pinning + 3 affected migrators for pypy, py312 & numpy2); though I guess the failure would also show up with one less migrator. Without https://github.com/conda-forge/conda-smithy/pull/1912/commits/3ecc8c4fb4126c28341ffbc0cd3eff99931398e5, the rerender wrongly produces
```
channel_sources:
- conda-forge
```

AFAICT this happens because the rerender will produce a list `["conda-forge", "conda-forge/label/numpy_dev,conda-forge"]` after merging, but we always just take the 0th element since conda build needs a string for the channel_sources.

It's not obvious how to broadcast these two elements to the 6 we need here, so the least we can do is update all of them consistently (in the migrators, and take that into account in smithy). The next step would be using `additional_zip_keys` in all three migrators, but that has two problems:
* AFAICT the current code is not ready for using the same `additional_zip_keys` in two migrations (because once the first one has been applied, the new key already is part of the zip)
* `additional_zip_keys` does not apply to `kind: version` migrators (or, more precisely, while we can add `operation: key_add` to end up calling `op_variant_key_add`, this does not work for a baseline version update). I've hacked around on this and couldn't get it to work.

beckermr (changes requested):
Thanks for adding tests! 

So it appears that the last migrator applied will win if two or migrators override the same non-zip key. 

This will potentially lead to 

 - race conditions if the order the migrators are applied in is randomized somehow (e.g., we glob for all of the migrator files on disk)
 - migrators interfering with each other in complex ways (even if they are applied in a manner with no race conditions)

What should we do about this?

h-vetinari:
> So it appears that the last migrator applied will win if two or migrators override the same non-zip key.

AFAICT it's quite unusual that several migrators are overriding the same key? Basically, this only happens if several different migrators tackle different components of a big zip, as happens with us for python 3.12, pypy & numpy. I don't think there are many other scenarios where this would become relevant, but perhaps I'm not creative enough...

I kind of think we should punt on this because for now it's mostly a hypothetical concern (compare the unhandled corner-cases for `additional_zip_keys`). An alternative would be to raise a warning if we spot an inconsistency (i.e. a key not part of a zip that has length >1 after merging, assuming that's a good metric).

> For channel sources, shouldn't we combine the strings at some point before the final ci support files are made?

They already need to be provided in combined form in the recipes CBC, c.f.
```
conda-forge/label/numpy_dev,conda-forge
^                           ^
1                           2
```

beckermr:
We should not leave this quirk / bug in the code just because we think it is rare. 

For the channel sources, I am saying that when smithy produces CBCs like this

```
channel_sources:
  - blah
  - blah_blah
```

it should automatically convert them to single strings 

```
channel_sources:
  - blah,blah_blah
```

h-vetinari:
> it should automatically convert them to single strings

That sounds really dangerous IMO. So far the entire `channel_sources` (plural!) logic is based on providing this value as a block. Having 
```
channel_sources:
  - blah
  - blah_blah
```
is almost certainly a sign of something gone wrong in the variant merge and should IMO warn/raise, not silently merge.

beckermr:
If we refuse to merge them automatically since it is a sign of an issue and so smithy should error loudly if it ever produces more than one value, wouldn't that preclude the code fixup here since this PR is intended to fix a case where smithy did produce more than one value but chose the wrong one?

beckermr:
In other words, how do I square the "raise on multiple values instead merging them logic" with this

> AFAICT this happens because the rerender will produce a list ["conda-forge", "conda-forge/label/numpy_dev,conda-forge"] after merging, but we always just take the 0th element since conda build needs a string for the channel_sources.

h-vetinari:
> wouldn't that preclude the code fixup here since this PR is intended to fix a case where smithy did produce more than one value but chose the wrong one?

I'm not sure I understand. Before https://github.com/conda-forge/conda-smithy/pull/1912/commits/3ecc8c4fb4126c28341ffbc0cd3eff99931398e5, the merge produced two values but ignored the second. With this PR it is at least able to produce a single value (provided the migrators are consistent between themselves).

> In other words, how do I square the "raise on multiple values instead merging them" logic with this

This was trying to describe the problem before this PR.

beckermr:
When I suggested that we automatically merge the keys, you told me that 

> That sounds really dangerous IMO. So far the entire channel_sources (plural!) logic is based on providing this value as a block. [...] is almost certainly a sign of something gone wrong in the variant merge and should IMO warn/raise, not silently merge. This kind of merge would almost certainly create issues with strict channel priority too.

The PR here is fixing the exact case of multiple values by picking one. This PR does not raise a warning or an error as you advocated above.

The options appear to be:

 - pick one of the values
 - raise a warning
 - raise an error
 - merge them automatically

Which option do you think we should do?

h-vetinari:
> The PR here is fixing the exact case of multiple values by picking one.

That's a misunderstanding I think. This PR is fixing it by ensuring there _is_ only one. It does this by always updating the value if a migrator specifies it, which is not exactly the same as picking one of the values, because it produces the correct result in the case of non-racy migrator specs.

The race condition argument is a fair criticism though, which we could try to warn on.

beckermr:
I am confused sorry. We already specify a key in the global pinnings, so any migrator setting anything creates more than one. I don't understand why the first one set is special relative to the rest. 

I don't like putting a partial solution into smithy and leaving edge cases for us to have to go back and figure out later. 

We should ensure this works in the general case now instead of taking the fast path to get one feature moving.

h-vetinari:
Let's take a step back. Our migrator infrastructure is supposed to help us go from:
```
foo:
  - 1
```
with 
```
__migrator:
  kind: version
foo:
  - 2
```
to 
```
foo:
  - 2
```

That is the absolute core functionality. In general this works well, being able to update multiple keys, or a component of a zip, or even all components of a zip. But it fails if the migrator updates a zip _plus_ something else. I've reduced the test here accordingly in https://github.com/conda-forge/conda-smithy/pull/1916, and it fails on `main`. IMO this must be considered a bug and so fixing it is not a feature.

However, I note that my proposed fix here does not fix the test in #1916, so I guess we can take that as net information gain from this discussion (I'm the first to admit that I have a hazy understanding of the smithy codebase at best 😅).

beckermr:
Yes indeed you have identified the core thing. My point is that this fix is only partial and I think we should take the time to develop something that won't suffer from race conditions and deals properly or raises an error for channels. (A warning is going to be not be visible enough given how our tools currently work.)

h-vetinari:
> something that won't suffer from race conditions

#1916 has only a single migration, so there's nothing to race. If we can fix that without opening other cans of worms, that should be an unequivocal win, right?

beckermr:
We'll start another python soon enough and then run right into this as the numpy one drags on for a while. Really I don't get why we are putting in a partial fix. Let's get this done while we are here so we don't have to come back later. If you are not willing, then i'll happily take this on but it may be a bit.

h-vetinari:
> Really I don't get why we are putting in a partial fix.

I did say that the fix I'm advocating is for #1916 first and foremost, and unrelated to this PR. How that fix will look is unknown for now (too late for me to continue investigating last night), so I'm not sure where you get the "partial fix" from. Let's judge the implementation when one is available?

beckermr:
I mean this PR is a partial fix for the issue it is claiming to fix as I detailed above.

PR #1916, opened 2024-04-30
WIP: add test for updating migrator-keys not part of zip_keys

h-vetinari:
In context of #1911 / #1912


h-vetinari:
OK, chasing down the failure here shows that it happens because `variant_key_add` doesn't know how to determine the "higher of the two" between `conda-forge` and `conda-forge/label/numpy_rc,conda-forge`. This can be solved by providing an `ordering:` to the migrator.

PR #1918, opened 2024-05-03
BUG do not stomp on zipped keys after they are reordered properly

### LINKED ISSUE #1917: BUG: something non-deterministic going wrong in squished_input_variants/preserve_top_level_loops
This issue first showed up a long time ago: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/4190

It occurs with regularity while trying to completely skip pypy builds on an already-migrated feedstock. Of course it would be possible to drop the migrator file completely, but I don't know if that doesn't put the migration bot in an inconsistent state somehow, and feels more intrusive than just a (often temporary) skip.

I hit this again today in https://github.com/conda-forge/scipy-feedstock/pull/274, where we really need to skip PyPy builds for now. 

The first thing to note is that debugging this is hard because the rerender changes based on random print statements or other changes I've made to my local smithy checkout (or even the `ordering:` length in the pypy migrator). This points to some indeterminism (e.g. dict/set order).

So far I was able to narrow it down to the following: things go wrong in
https://github.com/conda-forge/conda-smithy/blob/a04dba97f7f204c6df401bb2cda9c68644c8ad42/conda_smithy/configure_feedstock.py#L626-L627
Before that line, we have
```
used_key_values['python_impl']=['cpython', 'cpython', 'cpython', 'cpython']
```
whereas after it we get (for the failing case)
```
used_key_values['python_impl']=('pypy', 'cpython', 'cpython', 'cpython', 'cpython', 'cpython')
```
This happens against a background of the following values:
```
top_level_loop_vars={'target_platform', 'numpy', 'python', 'python_impl'}
all_used_vars={'docker_image', 'MACOSX_DEPLOYMENT_TARGET', 'libblas', 'channel_sources', 'macos_machine', 'cdt_name', 'liblapack', 'channel_targets', 'BUILD', 'numpy', 'python', 'pin_run_as_build', 'MACOSX_SDK_VERSION', 'python_impl', 'c_compiler', 'target_platform', 'macos_min_version', 'c_stdlib', 'cxx_compiler', 'zip_keys', 'cdt_arch', 'build_number_decrement', 'libcblas'}
preserve_top_level_loops={'python_impl'}
```

Given the definition
https://github.com/conda-forge/conda-smithy/blob/a04dba97f7f204c6df401bb2cda9c68644c8ad42/conda_smithy/configure_feedstock.py#L579

it's possible that `python_impl` should be detected as used (if only from `skip: true  # [python_impl == "pypy"]`) but isn't. Beyond that, I've kind of saturated on smithy code for now 😅  Any help, @beckermr? 🙏

beckermr:
Checklist
* [ ] Added a ``news`` entry

closes #1917

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -6,7 +6,6 @@ import os
 import re
 import subprocess
 import sys
-import pprint
 import textwrap
 import time
 import yaml
@@ -59,6 +58,7 @@ from conda_smithy.validate_schema import (
 from conda_smithy.utils import (
     get_feedstock_about_from_meta,
     get_feedstock_name_from_meta,
+    log_debug_with_pprint,
 )
 
 from . import __version__
@@ -86,25 +86,6 @@ if "CONDA_SMITHY_SERVICE_FEEDSTOCKS" in os.environ:
 CONDA_FORGE_PINNING_LIFETIME = int(
     os.environ.get("CONDA_FORGE_PINNING_LIFETIME", 15 * 60)
 )
-
-
-def _ordereddict_to_dict(od):
-    """convert an ordereddict to a dict"""
-    return {
-        k: _ordereddict_to_dict(v) if isinstance(v, OrderedDict) else v
-        for k, v in od.items()
-    }
-
-
-def _log_debug_with_pprint(msg, obj):
-    """This helper avoids calling pprint.pformat when we don't plan to use the output."""
-    if logger.isEnabledFor(logging.DEBUG):
-        # ordered dicts are ugly when printed, so we convert them to regular dicts
-        if isinstance(obj, OrderedDict) or isinstance(obj, dict):
-            new_obj = _ordereddict_to_dict(obj)
-        else:
-            new_obj = obj
-        logger.debug(msg + "\n" + pprint.pformat(new_obj))
 
 
 # use lru_cache to avoid repeating warnings endlessly;
@@ -348,7 +329,7 @@ def _get_used_key_values_by_input_order(
         if key in squished_input_variants
     }
 
-    _log_debug_with_pprint("initial/input used_key_values:", used_key_values)
+    log_debug_with_pprint(logger, "initial/input used_key_values:", used_key_values)
 
     # we want remove any used key values not in used variants and make sure they follow the
     #   input order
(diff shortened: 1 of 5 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -609,22 +618,27 @@ def _collapse_subpackage_variants(
...
+        # so sometimes a key in a zip will not be used in the recipe and so
+        # gets put into preserve_top_level_loops.
+        # however, if we assign it to used_key_values, the careful reordering
+        # of the zipping in `_get_used_key_values_by_input_order` gets undone.
+        # so we skip it here
+
+        # TODO - this is a hack and needs fixing
+        # it turns out sometimes conda-build just doesn't return some of the variants
+        # even if they are in the zipped vars. So we also keep any key which on output
+        # has a length of 1 but had length > 1 on input.
```
beckermr, line 639: @mbargull Can you look the issue here over? 

In particular, IDK why conda-build when it computes all of the outputs making a list of metas, sometimes does and sometimes does not return outputs for the all of the key values in preserve_top_level_loops. 

Right now I put in a hack to get the tests to pass but I suspect there is something more fundamental going on that needs fixing.

h-vetinari:
Tested this branch on the scipy-PR, and confirmed it works (see https://github.com/conda-forge/scipy-feedstock/pull/274/commits/7a3cbf2767f04e84fc8729f30686f2e9969f5934). Thanks a lot for the quick handling! :)

beckermr:
Awesome! I tested it on my own simpler scipy branch as well. I am not happy with the fix and am waiting for someone with more conda-build expertise to comment on the appearance of top-level loop vars in the configs before moving on.

conda_smithy/configure_feedstock.py
```diff
@@ -2603,7 +2592,8 @@ def main(
     temporary_directory=None,
 ):
     loglevel = os.environ.get("CONDA_SMITHY_LOGLEVEL", "INFO").upper()
-    logger.setLevel(loglevel)
+    _logger = logging.getLogger("conda_smithy")
+    _logger.setLevel(loglevel)
```
beckermr, line 2596:
```suggestion
    logging.getLogger("conda_smithy").setLevel(loglevel)
```

conda_smithy/configure_feedstock.py
beckermr: I figure this out I think. When conda-build renders a recipe, it returns the metadata for each unique file generated. If a key we use is zipped with a key we do not use, conda-build will ignore the variants involving all but one of the values of the key we do not use. This behavior is not what we do in conda-forge (i.e., we want all variants that are not explicitly skipped even if some of the keys in the variants are not explicitly used).

The most robust way to fix this is to write a custom function that returns metadata for each of the variants in the full exploded matrix except the ones that the recipe skips.

jakirkham:
Would it be alright if we separate out the logging change into another PR?

Both changes seem reasonable, but combining them makes it a bit harder to parse what is happening here

beckermr:
Yeah absolutely! Sorry for the mess of a diff. I am still struggling to understand some of thr subtle things in this PR and probably marked it as ready for review too early. :/

conda_smithy/configure_feedstock.py
beckermr: The part about zip keys is wrong above. It is a similar issue but about top-level keys not used by outputs.

beckermr:
closed by #1921

PR #1919, opened 2024-05-03
Introduce Ruff Linter

ytausch:
Checklist
* [x] Added a ``news`` entry

This PR introduces [ruff](https://github.com/astral-sh/ruff) as a source code linter, which will ensure a consistent and modern code style.

Please see the commits to check which fixes where applied manually and which automatically.


tests/test_lint_recipe.py
```diff
@@ -1312,7 +1306,7 @@ def test_version(self):
             "Package version 2.0.0~alpha0 doesn't match conda spec"
         )
         lints, hints = linter.lintify_meta_yaml(meta)
-        self.assertIn(expected_message, lints)
+        assert any(expected_message in lint for lint in lints)
```
h-vetinari, line 1309: If we replace this, it should be
```suggestion
        assert expected_message in lints
```

However, I see a lot of `self.assertIn` that aren't changed here, so I'm wondering what's the rationale for this change?

(if the above suggestion fails for whatever reason, match the expected message exactly)

conda_smithy/lint_recipe.py
```diff
@@ -490,18 +486,17 @@ def lintify_meta_yaml(
...
+    if (version := package_section.get("version")) is not None:
         try:
-            VersionOrder(ver)
-        except:
+            VersionOrder(str(version))
+        except InvalidVersionSpec as e:
             lints.append(
-                "Package version {} doesn't match conda spec".format(ver)
+                f"Package version {version} doesn't match conda spec: {e}"
             )
```
ytausch, line 501: @h-vetinari This needed to be changed because the except clause is too broad.

Therefore I catch the only error that can happen here, `InvalidVersionSpec`, and, in one process, also print the `InvalidVersionSpec` error details because it helps debugging.

tests/test_lint_recipe.py
ytausch: The lint now includes error details that are generated by our dependency `conda` (see the other comment in which I tagged you). Therefore, your suggestion fails. And also, I did not want to match the expected message exactly because `conda` could change their error message slightly and then our tests fails.

beckermr (changes requested):
I'd be in favor of expanding the ruleset for ruff in the pyproject.toml

```
[tool.ruff.lint]
select = ["E", "F", "W"]
preview = true
```

Thoughts everyone?

ytausch:
@beckermr What do you mean by "expanding the ruleset"?

My PR already contains more rulesets than the default (and more than you propose):

https://github.com/conda-forge/conda-smithy/blob/12f4a1fa107b75069d5380e166b2d5d41cbb6ed9/pyproject.toml#L36-L52

Also, what is the rationale behind your proposal to enable `preview` mode? Preview mode contains [unstable features](https://docs.astral.sh/ruff/preview/), which is not something we want here IMO.

conda_smithy/anaconda_token_rotation.py
```diff
@@ -361,9 +338,10 @@ def rotate_token_in_travis(
 
 
 def rotate_token_in_azure(user, project, binstar_token, token_name):
+    from vsts.build.v4_1.models import BuildDefinitionVariable
+
     from .azure_ci_utils import build_client, get_default_build_definition
     from .azure_ci_utils import default_config as config
-    from vsts.build.v4_1.models import BuildDefinitionVariable
```
jakirkham, line 366: Out of curiosity, do we know what led to this change?

conda_smithy/lint_recipe.py
```diff
@@ -1,42 +1,43 @@
...
     import tomllib
 
 from conda.models.version import VersionOrder
+from conda_build.metadata import (
+    FIELDS as _CONDA_BUILD_FIELDS,
+)
 from conda_build.metadata import (
     ensure_valid_license_family,
-    FIELDS as cbfields,
 )
```
jakirkham, line 31: Why is it splitting these into 2 `import`s from the same module? This seems overly verbose. Is there an option to fix this?

conda_smithy/lint_recipe.py
```diff
@@ -490,18 +486,17 @@ def lintify_meta_yaml(
...
                             "the selectors are necessary, please remove "
-                            "`noarch: {}`.".format(noarch_value)
+                            f"`noarch: {noarch_value}`."
                         )
                         break
 
     # 19: check version
-    if package_section.get("version") is not None:
-        ver = str(package_section.get("version"))
+    if (version := package_section.get("version")) is not None:
```
jakirkham, line 494: Think we can simplify this further

```suggestion
    version = str(package_section.get("version", ""))
    if version:
```

conda_smithy/lint_recipe.py
```diff
@@ -490,18 +486,17 @@ def lintify_meta_yaml(
...
         try:
-            VersionOrder(ver)
-        except:
+            VersionOrder(str(version))
```
jakirkham, line 496:
```suggestion
            VersionOrder(version)
```

tests/conftest.py
```diff
@@ -4,17 +4,15 @@
 
 import pytest
 import yaml
-
+from conda_build.utils import copy_into
 from jinja2 import FileSystemLoader
 from jinja2.sandbox import SandboxedEnvironment
```
jakirkham, line 9: Think the idea here was to group `import`s and then `from...import...`s and separate those out

Does `ruff` have a way to do this?

tests/test_condaforge_config_schema.py
```diff
@@ -1,8 +1,7 @@
 import pytest
 from pydantic import ValidationError
-import yaml
-from conda_smithy.schema import ConfigModel
 
+from conda_smithy.schema import ConfigModel
```
jakirkham, line 4: Interesting here it groups some `from...import...`s with `import`s and splits some out. Or is it just trying to group external `import`s separately from internal ones?

jakirkham (commented):
Thanks Yannik! 🙏

Generally this looks good. Had a few questions/observations below

conda_smithy/lint_recipe.py
jaimergp: I think that's part of the black/isort standard too. `as` imports have their own line 🤷

.pre-commit-config.yaml
```diff
@@ -10,6 +10,11 @@ repos:
     rev: 24.4.2
     hooks:
     -   id: black
```
jaimergp, line 12: We should remove black if we have ruff now.

tests/conftest.py
jaimergp: From what I've seen, black/isort distinguish three blocks (one blank line of separation):

- stdlib
- 3rd party
- the package itself

Then within each block, we'll see the `import` lines first, followed by `from X import Y`, alphabetically. No blank line.

jaimergp:
Awesome! I suggest that when we merge this one, we do it via Squash, so then we can easily omit this PR in the blame view via that github file whose name I can't remember :)

conda_smithy/anaconda_token_rotation.py
ytausch: The ruff linter `I` rules sort imports with `isort`. Does that answer your question?

conda_smithy/lint_recipe.py
ytausch: Yes.

conda_smithy/lint_recipe.py
ytausch: I propose to stay defensive here and crash intentionally if `version` is an empty string (as opposed to `None`). Open to split the `:=` assignment though if you think this is too fancy.

conda_smithy/lint_recipe.py
ytausch: see above, I think this relates to the change above

tests/conftest.py
ytausch: I also think so. The reasoning behind `isort` makes sense and we don't need to change something about it IMO.

tests/test_condaforge_config_schema.py
ytausch: This is exactly what happens.

.pre-commit-config.yaml
ytausch: Actually, I think the ruff linter is orthogonal to `black` as a formatter. There is also the ruff formatter (which is a Black alternative) but it is currently disabled. Using the ruff linter together with Black is a [common use case](https://docs.astral.sh/ruff/faq/#is-the-ruff-linter-compatible-with-black).

Open to switch to the ruff formatter in a separate PR if you think that makes sense - but for now, I propose to stick with Black.

beckermr: agreed on using both ruff and black

jaimergp:
Any thoughts about the rationale for the preview mode exposed in https://github.com/conda-forge/conda-forge-webservices/pull/616?

beckermr:
I like all of the flake8 rules so I like preview mode, but that is me.

.pre-commit-config.yaml
beckermr: Actually I am confused now. Do we still get autofixes from ruff if we don't use it as a formatter?

ytausch: Yes, we do.

ytausch:
I would be fine with enabling preview mode if you think that's necessary - is there anything else you want to debate about this PR? All open discussions are solved from my side I want this to get merged.

ytausch:
@conda-forge/core

conda_smithy/lint_recipe.py
h-vetinari: I like John's suggestion much better.

> I propose to stay defensive here and crash intentionally if `version` is an empty string (as opposed to `None`).

That is way too subtle to go without a comment, but in any case, I don't think we need to: We should make the version check unconditional (even for `""`), which should be the best of both worlds AFAICT:
```
version = str(package_section.get("version", ""))
try:
    VersionOrder(version)
except InvalidVersionSpec as e:
    ...
```

I'm not aware of any feedstocks without a version; unless I'm overlooking a use-case here, I'd like to just get rid of that corner case completely.

h-vetinari (commented):
Basically LGTM, but some small things left.

conda_smithy/lint_recipe.py
ytausch: How about this solution? It takes your approach but makes it super-obvious what is going on.

0c04145

h-vetinari: Fine by me. There are two tests that now produce a new linter hint that needs to be handled.

ytausch: Done.

ytausch:
@h-vetinari @beckermr @jaimergp @jakirkham Can we please get this through?

beckermr (approved):

beckermr:
Let's wait for a few more reviews and then merge.

conda_smithy/github.py
```diff
@@ -1,27 +1,24 @@
...
 from git import Repo
-
 from github import Github
 from github.GithubException import GithubException
 from github.Organization import Organization
 from github.Team import Team
-import github
 
-import conda_build.api
 from conda_smithy.utils import get_feedstock_name_from_meta
```
jaimergp, line 12: We are not being too consistent with relative / absolute imports, I think

news/1919-ruff.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Introduced ``ruff`` as pre-commit linter.
```
jaimergp, line 7:
```suggestion
* Introduced ``ruff`` as pre-commit linter. (#1919) 
```

jaimergp (approved):
Just some comments that came to mind while reading the changes. The only blocking one is the PR number in the news file.

conda_smithy/anaconda_token_rotation.py
```diff
@@ -145,9 +139,7 @@ def rotate_anaconda_token(
                         if "DEBUG_ANACONDA_TOKENS" in os.environ:
                             raise e
                         else:
-                            err_msg = (
-                                "Failed to rotate token for %s/%s" " on azure!"
-                            ) % (user, project)
+                            err_msg = f"Failed to rotate token for {user}/{project} on azure!"
```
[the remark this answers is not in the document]
ytausch: I would like, however `rotate_anaconda_token` is in such a bad condition that I don't want to touch anything which is not 1) strictly ruff-lrelated 2) a complete refactoring of the method, or removal (I could not figure out where the CLI command is used)

2) is out of scope for this PR, so I created a separate issue #1962.

conda_smithy/configure_feedstock.py
```diff
@@ -3,26 +3,22 @@
...
 from os import fspath
 from pathlib import Path, PurePath
-import requests
 
-try:
-    from builtins import ExceptionGroup
-except ImportError:
-    from exceptiongroup import ExceptionGroup
+import requests
+import yaml
```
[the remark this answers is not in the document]
ytausch: #1963

conda_smithy/configure_feedstock.py
```diff
@@ -328,9 +322,7 @@ def _get_used_key_values_by_input_order(
         for key in all_used_vars
         if key in squished_input_variants
     }
-    logger.debug(
-        "initial used_key_values {}".format(pprint.pformat(used_key_values))
-    )
+    logger.debug(f"initial used_key_values {pprint.pformat(used_key_values)}")
```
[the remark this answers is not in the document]
ytausch: In general, yes. And there is even a ruff rule to check that (which is currently disabled): [G004](https://docs.astral.sh/ruff/rules/logging-f-string/)

I personally prefer f-strings for logging because I think they are easier to read, and usually there should not be much performance to gain because calculating the string representations of the objects should be fast anyway.

I found https://stackoverflow.com/a/54368109, so it seems like I am in a minority, so if you want to, I am totally open to adding the G-rules enforcing logging with deferred positional arguments. Let me know what you think.

conda_smithy/feedstock_tokens.py
```diff
@@ -208,13 +194,11 @@ def feedstock_token_exists(user, project, token_repo, provider=None):
         if err_msg:
             raise FeedstockTokenError(err_msg)
         else:
+            provider_or_empty = "" if provider is None else " " + provider
```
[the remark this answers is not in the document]
ytausch: done in 0840cc7

conda_smithy/github.py
ytausch: This is how the new file looks like:

https://github.com/conda-forge/conda-smithy/blob/d8a8d8dab3baa97a61374bc14764995931a30053/conda_smithy/github.py#L1-L14

Seems pretty ordered to me. Note https://github.com/conda-forge/conda-smithy/pull/1919#discussion_r1593447856

jaimergp: Sometimes we are doing `from .utils import X` and sometimes `from conda_smithy.utils import X`.

Compare this line to this [other one](https://github.com/conda-forge/conda-smithy/pull/1919/files#diff-27f5c1fcf8e15b2b01490459c0c6589f02d699d19110b6a9a80d66b034f6fc8bR19)

conda_smithy/configure_feedstock.py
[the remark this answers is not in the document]
jaimergp: No strong feelings here. My style is usually deferring as long as possible in logging out of habit so this reads weird, but I'm not going to oppose. Whatever we use, we should be consistent though. So pick one and enforce a rule, it's what I'd say.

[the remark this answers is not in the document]
ytausch: Alright, let's enforce the G rules then.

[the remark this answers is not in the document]
ytausch: 5dd60c1ccaa8eaa6e0c4a1d24972b50541dff88f

conda_smithy/github.py
ytausch: Good point. My proposal: Let's ban all relative imports with the `TID` ruleset and `flake8-tidy-imports.ban-relative-imports = "all"`.

I've done this in 4902d5d.

Quote from [PEP 8](https://peps.python.org/pep-0008/#imports):

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
```python
import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example
```
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
```python
from . import sibling
from .sibling import example
```

So I think we're in line with PEP 8 since the package layout is not complex at all.

ytausch:
@jaimergp

ytausch:
Rebasing this over and over again is quite tedious so I'll wait until we have found a consensus about the rules we want to enforce and then I can do this one last time. Happy if we finally close this after almost 3 months.

beckermr:
We have two approvals from core , so I'd be happy to merge today.

ytausch:
To make rebasing easier and the commit history cleaner, I reran ruff with the final configuration, and reapplied everything we have discussed here. From my side, this should be fine now.

beckermr:
We have two test failures. After those are fixed, we can merge.

ytausch:
Done.

wolfv:
Awesome job, @ytausch

PR #1928, opened 2024-05-12
ensure MACOSX_SDK_VERSION is at least c_stdlib_version

### LINKED ISSUE #1927: BUG: MACOSX_SDK_VERSION not participating in merge logic
I just noticed that is a recipe uses 
```
c_stdlib_version:          # [osx and x86_64]
  - "10.12"                # [osx and x86_64]
MACOSX_SDK_VERSION:        # [osx and x86_64]
  - "10.12"                # [osx and x86_64]
```
the merge logic of smithy will not be applied to `MACOSX_SDK_VERSION`, which will end up being incorrectly populated as 10.12 (while `MACOSX_DEPLOYMENT_TARGET` ends up being 10.13 after the merge with the global pinning).

I guess we should also add a linter warning for c_stdlib_version / MACOSX_SDK_VERSION <10.13?

CC @beckermr

h-vetinari:
Fixes #1927 (minus the linter)

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -439,6 +439,8 @@ def _merge_deployment_target(container_of_dicts, has_macdt):
       For now, take the maximum to populate both.
     - In any case, populate MACOSX_DEPLOYMENT_TARGET, as that is the key picked
       up by https://github.com/conda-forge/conda-forge-ci-setup-feedstock
+    - If MACOSX_SDK_VERSION is lower than the merged value from the previous step,
+      update it to match the merged value.
     """
     result = []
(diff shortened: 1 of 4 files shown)


h-vetinari:
Note that the current state of this PR will end up populating `MACOSX_SDK_VERSION` everywhere, but with a matching version to whatever we get after merging `c_stdlib_version` / `MACOSX_DEPLOYMENT_TARGET`. If we want to avoid that, I can change the logic here to only add `MACOSX_SDK_VERSION` if it is different. But I don't really see how it would hurt to have it in the variant configs.

CC @beckermr @isuruf

beckermr (approved):

PR #1929, opened 2024-05-13
Add linter hints for issues with OSX configuration in CBC

h-vetinari:
Linter portion of #1927, but also adding linter hints for warnings from #1889, which were so far only visible when running smithy locally.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -910,11 +910,19 @@ def lintify_meta_yaml(
     # ```
     cbc_osx = dict(filter(lambda item: item[1] is not None, cbc_osx.items()))
 
-    # for fallback, assume ordering [x64, arm64] which is used (almost) universally
+    def sort_osx(versions):
+        # we need to have a known order for [x64, arm64]; in the absence of more
+        # complicated regex processing, we assume that if there are two versions
+        # being specified, the higher one is osx-arm64.
+        if len(versions) == 2:
+            if VersionOrder(str(versions[0])) > VersionOrder(str(versions[1])):
+                versions = versions[::-1]
+        return versions
+
     baseline_version = ["10.13", "11.0"]
-    v_stdlib = cbc_osx.get("c_stdlib_version", baseline_version)
-    macdt = cbc_osx.get("MACOSX_DEPLOYMENT_TARGET", baseline_version)
-    sdk = cbc_osx.get("MACOSX_SDK_VERSION", baseline_version)
+    v_stdlib = sort_osx(cbc_osx.get("c_stdlib_version", baseline_version))
+    macdt = sort_osx(cbc_osx.get("MACOSX_DEPLOYMENT_TARGET", baseline_version))
+    sdk = sort_osx(cbc_osx.get("MACOSX_SDK_VERSION", baseline_version))
 
     if {"MACOSX_DEPLOYMENT_TARGET", "c_stdlib_version"} <= set(cbc_osx.keys()):
         # both specified, check that they match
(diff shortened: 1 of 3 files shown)


conda_smithy/lint_recipe.py
```diff
@@ -886,6 +887,125 @@ def check_pins_build_and_requirements(top_level):
...
+    # filter None values out of cbc_osx dict, can appear for example with
+    # ```
+    # c_stdlib_version:  # [unix]
+    #   - 2.17           # [linux]
+    #   # note lack of osx
+    # ```
+    cbc_osx = dict(filter(lambda item: item[1] is not None, cbc_osx.items()))
+
+    # for fallback, assume ordering [x64, arm64] which is used (almost) universally
+    baseline_version = ["10.13", "11.0"]
```
h-vetinari, line 914: We could add _yet another_ lint if the order here is "wrong" (from the POV of the rest of the processing), i.e. for
```
c_stdlib_version:           # [osx]
  - 11.0                    # [osx and arm64]
  - 10.13                   # [osx and x86_64]
```

h-vetinari: Or alternatively, if we find `version(first_entry) > version(second_entry)`, we assume that we've got arm64 sorted before, and just reverse the list. That would still leave an exceedingly small corner case open[^1], but that shouldn't be an issue IMO

[^1]: if osx-arm64 genuinely uses a lower deployment-target or sdk than osx-64, which I don't think ever makes sense

beckermr: Why does the ordering matter here? Is there a zip I am not remembering?

h-vetinari: Mostly the whole thing is about having 10.13 somewhere. I added 11.0 for osx-arm64 as well because otherwise I cannot diagnose a handful of cases.

However the tradeoff is the need to assume some order between osx-64 and osx-arm64 (or do a bunch more regex work, which currently just filters out anything that's not explicitly osx-related).

h-vetinari: > Or alternatively, if we find `version(first_entry) > version(second_entry)`, we assume that we've got arm64 sorted before, and just reverse the list.

I ended up going with that. This should have essentially zero real failure modes AFAICT. 🤞

beckermr (approved):

maresb:
I was pretty confused about the `c_stdlib_version` below the current global baseline warning I received [here](https://github.com/conda-forge/nutpie-feedstock/pull/34#issuecomment-2172085476).

After some investigation I realized that it was coming from [this file](https://github.com/conda-forge/nutpie-feedstock/blob/c5a3a49401485e3d751b3285fbaf77bce1035527/recipe/conda_build_config.yaml). (It would have been helpful to indicate the filename `recipe/conda_build_config.yaml`.) I created that file according to [this announcement](https://conda-forge.org/news/2024/01/29/rust-175-osx-sdk/) and then forgot about it.

My confusion was because I don't have the bigger picture of what's going on, and when I created `conda_build_config.yaml` I was blindly following the announcement without fully realizing that I was pinning the `c_stdlib_version` which might need to be updated in the future. In my case it seemed to work to simply delete the entry from `conda_build_config.yaml`. Should this be mentioned as another possible resolution? (Also, I hope this was the correct action in my case?)

h-vetinari:
> I was pretty confused about the `c_stdlib_version` below the current global baseline warning I received [here](https://github.com/conda-forge/nutpie-feedstock/pull/34#issuecomment-2172085476).

Sorry about that...



> In my case it seemed to work to simply delete the entry from `conda_build_config.yaml`. Should this be mentioned as another possible resolution?

PRs to improve the warning messages here (or documentation), so that they would have been more understandable to you, would be greatly appreciated!



> Also, I hope this was the correct action in my case?

Yes, deleting a pin for `c_stdlib_version <10.13` is completely fine.

PR #1930, opened 2024-05-14
Add dependabot configuration and pin Github Actions actions

jaimergp:
Checklist
* [x] Added a ``news`` entry

Trying this out. I'm expecting this config to do the following:

- Update the workflows under `.github/workflows` as usual
- ~Maaaaybe pick the `github-actions.yml.tmpl` file in the templates. If that's true, we should see a PR updating `upload-artifacts`.~ It won't.
- I added a dummy workflow with the same `uses` lines as the GHA template. This will act as a proxy that we reconcile via a test with the actual template.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - '*'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directory: ".github/workflows/"
     schedule:
       interval: "weekly"
     groups:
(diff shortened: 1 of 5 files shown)


jaimergp:
I don't think we'll be lucky with the template... [Dependabot parses the YAML](https://github.com/dependabot/dependabot-core/blob/5d60fd0b3a28a598fd9e977c562290871766d29a/github_actions/lib/dependabot/github_actions/file_parser.rb#L47).

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: ".github/workflows/"
+    schedule:
+      interval: "weekly"
```
beckermr, line 6: monthly would be better. IDK what the syntax is though.

beckermr (changes requested):
See above.

jaimergp:
Addressed.

beckermr (dismissed):
We should not merge this. The goal with our CI is to ensure things are reproducible including rendering with smithy.

beckermr (approved):

PR #1936, opened 2024-05-20
Lint / hint about using Python wheels as sources

### LINKED ISSUE #1935: lint on most wheel-based python installs except `noarch: python` and `none-any.whl`
There are now several feedstocks that use pure python wheels with `noarch: python` recipes. This procedure is in principle safe, but an sdist is preferred. We should lint on any usage of wheels outside this one case. A second PR might also provide a hint that sdists are preferred.

jaimergp:
Checklist
* [X] Added a ``news`` entry

Closes #1935

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -711,7 +711,8 @@ def lintify_meta_yaml(
         formatted_urls = ", ".join([f"`{url}`" for url in compiled_wheel_urls])
         lints.append(
             f"Detected compiled wheel(s) in source: {formatted_urls}. "
-            "This is discouraged. Please consider using a source distribution (sdist) instead."
+            "This is disallowed. All packages should be built from source except in "
+            "rare and exceptional cases."
         )
     if pure_python_wheel_urls:
         formatted_urls = ", ".join(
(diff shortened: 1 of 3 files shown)


beckermr (approved):
We may want stronger language. Maybe "All packages should be built from source except in rare and exceptional cases."

conda_smithy/lint_recipe.py
```diff
@@ -693,6 +693,42 @@ def check_pins_build_and_requirements(top_level):
...
+            for line in f:
+                if match := pure_python_wheel_re.search(line):
+                    pure_python_wheel_urls.append(match.group(1))
+                elif match := wheel_re.search(line):
+                    compiled_wheel_urls.append(match.group(1))
+    if compiled_wheel_urls:
+        formatted_urls = ", ".join([f"`{url}`" for url in compiled_wheel_urls])
+        lints.append(
+            f"Detected compiled wheel(s) in source: {formatted_urls}. "
+            "This is discouraged. Please consider using a source distribution (sdist) instead."
```
jaimergp, line 714:
```suggestion
            "This is disallowed. All packages should be built from source except in "
            "rare and exceptional cases."
```

jaimergp: Something like this @beckermr?

beckermr:
Yeah that's great.

beckermr:
Thank you!

PR #1944, opened 2024-05-30
PROD lower azure maxParallel to 10

beckermr:
Checklist
* [x] Added a ``news`` entry

We've grown a lot and this is now a more sensible default. It can be made bigger on a case by case basis as well.

### DIFF
--- a/conda_smithy/data/conda-forge.yml
+++ b/conda_smithy/data/conda-forge.yml
@@ -4,7 +4,7 @@ azure:
   build_id: null
   force: false
   free_disk_space: false
-  max_parallel: 50
+  max_parallel: 10
   project_id: 84710dde-1620-425b-80d0-4cf5baca359d
   project_name: feedstock-builds
   settings_linux:
--- a/news/azure_max_parallel.rst
+++ b/news/azure_max_parallel.rst
@@ -1,0 +1,4 @@
+**Changed:**
+
+* Lowered default ``maxParallel`` on azure to 10.
+
--- a/news/azure_max_parallel.rst
+++ b/news/azure_max_parallel.rst
@@ -1,4 +1,3 @@
 **Changed:**
 
 * Lowered default ``maxParallel`` on azure to 10.
-


beckermr:
pre-commit.ci autofix

conda_smithy/data/conda-forge.yml
```diff
@@ -4,7 +4,7 @@ azure:
   build_id: null
   force: false
   free_disk_space: false
-  max_parallel: 50
+  max_parallel: 10
```
h-vetinari, line 7: I think dividing by 5 is pretty aggressive. Feedstocks building the full python matrix for all architectures have at least 30 builds (not counting PyPy), so we'd be tripling their CI time for a pretty standard case.

beckermr: What number do you want?

h-vetinari: I'm not sure if we can come up with some good bits of automation there, but it'd be probably more impactful & less disruptive if we could manage to concentrate these efforts on CIs with very large matrices (especially if they "abuse" the microarch levels), coupled with long runtimes.

beckermr: People can make this bigger by overriding in their feedstock. This is lowering the default.

beckermr: Alright. This PR is the extent to which I want to work on this specific item at the moment. I will leave it to you or someone else to make any changes you like and merge, or to simply let it languish. Thanks for the review!

h-vetinari: Overall, my instinct would be not to do this pre-emptively until we run into more concrete trouble. What I mean is that if there's a feedstock that's consuming way too many resources, we could point it out to them to either reduce their build matrix or exercise caution ([example](https://github.com/conda-forge/arrow-cpp-feedstock/issues/825)), and I believe people will be receptive to following such a request.

However, if consensus is that we must reduce already now...

> What number do you want?

I don't know... 30? 25?

> People can make this bigger by overriding in their feedstock. This is lowering the default.

I get it, but a lot of feedstock maintainers will not necessarily discover this, or how to fix it. I guess we could also live with that.

PR #1953, opened 2024-06-20
test: skip test that relies on GH_TOKEN if that is not set

janl:


janl:
this was an accident, apologies.

PR #1955, opened 2024-06-20
chore: remove now obsolete setup.py

janl:
Checklist
* [x] Added a ``news`` entry

There are two places where I’m not sure my changes are desired:

1. https://github.com/neighbourhoodie/conda-smithy/commit/d11dce780edcfc3aee37040438abab3b4de5efa9 drops `--single-version-externally-managed --record=record.txt` from the install command because I did not manage to find any information about migrating those and I humbly request advice on whether this is correct or what the correct variant needs to be (tests pass locally).

2. https://github.com/neighbourhoodie/conda-smithy/commit/57b07858148cd712fcdbf8ca26746a81c4759ae6 migrates away from setup.py but since this is about linting of other projects and not conda-smithy itself (afaict), I’m not sure, this is the right change to make and I’d need some advice from someone working on the project to make a final call.

I’m happy to adjust anything you might request.

Please let me know if I missed anything.

This is part of the STF work for Conda.

### DIFF
--- a/tests/recipes/click-test-feedstock/recipe/meta.yaml
+++ b/tests/recipes/click-test-feedstock/recipe/meta.yaml
@@ -16,7 +16,7 @@ source:
 
 build:
   number: {{ build_num }}
-  script: python -m pip install .
+  script: python setup.py install --single-version-externally-managed --record=record.txt
 
 requirements:
   host:
--- a/tests/recipes/click-test-feedstock/recipe/meta.yaml
+++ b/tests/recipes/click-test-feedstock/recipe/meta.yaml
@@ -16,7 +16,7 @@ source:
 
 build:
   number: {{ build_num }}
-  script: python setup.py install --single-version-externally-managed --record=record.txt
+  script: python -m pip install .
 
 requirements:
   host:
--- a/tests/recipes/noarch_platforms/recipe/meta.yaml
+++ b/tests/recipes/noarch_platforms/recipe/meta.yaml
@@ -10,7 +10,7 @@ source:
 
 build:
   number: 1
-  script: {{ PYTHON }} -m pip install .
+  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record record.txt
   entry_points:
     - pytest = pytest:console_main
     - py.test = pytest:console_main
--- a/tests/recipes/noarch_platforms/recipe/meta.yaml
+++ b/tests/recipes/noarch_platforms/recipe/meta.yaml
@@ -10,7 +10,7 @@ source:
 
 build:
   number: 1
-  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record record.txt
+  script: {{ PYTHON }} -m pip install .
   entry_points:
     - pytest = pytest:console_main
     - py.test = pytest:console_main
--- a/tests/test_lint_recipe.py
+++ b/tests/test_lint_recipe.py
@@ -1002,14 +1002,14 @@ class Test_linter(unittest.TestCase):
         meta = {
             "build": {
                 "skip": "True",
-                "script": "python -m pip install",
+                "script": "python setup.py install",
                 "number": 0,
             }
         }
         lints, hints = linter.lintify_meta_yaml(meta)
         self.assertNotIn(expected_message, lints)
 
-        meta = {"build": {"skip": "True", "script": "python -m pip install"}}
+        meta = {"build": {"skip": "True", "script": "python setup.py install"}}
         lints, hints = linter.lintify_meta_yaml(meta)
         self.assertIn(expected_message, lints)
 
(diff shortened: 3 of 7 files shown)


beckermr:
This file was kept for folks who still used the setup.py directly. However, given that this feature is very much on its way out and has been deprecated for a while now, we can remove.

beckermr (approved):

beckermr (changes requested):
I forgot that we need to remove the bits in the tests that rely on executing the setup.py too.

beckermr:
See this line: https://github.com/conda-forge/conda-smithy/actions/runs/9596450561/job/26465117542?pr=1955#step:5:31

jaimergp (changes requested):
We will need to bump the minimum version in the setuptools dependency in pyproject.toml so we make sure that editable installs still work. `>=70` seems fair, given how there was one more fix in the editable logic, according to the [changelog](https://github.com/pypa/setuptools/blob/main/NEWS.rst).

Also please provide a `news` file as instructed in the PR description template (which ideally is not deleted as part of the submission).

Thanks!

janl:
Thank you for the quick review and my apologies for opening a clearly sub-par PR. I was a bit over-eager to close this out.

I have updated the PR text to reflect the additional work and open questions, thank you in advance for any hints.

Apologies also for missing the news item initially.

tests/recipes/click-test-feedstock/recipe/meta.yaml
```diff
@@ -16,7 +16,7 @@ source:
 
 build:
   number: {{ build_num }}
-  script: python setup.py install --single-version-externally-managed --record=record.txt
+  script: python -m pip install .
```
beckermr, line 19: This change is not needed and is unrelated to the packaging of conda-smithy. Please back it out.

tests/recipes/noarch_platforms/recipe/meta.yaml
```diff
@@ -10,7 +10,7 @@ source:
 
 build:
   number: 1
-  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record record.txt
+  script: {{ PYTHON }} -m pip install .
```
beckermr, line 13: Same here.

tests/test_lint_recipe.py
```diff
@@ -1002,14 +1002,14 @@ def test_missing_build_number(self):
         meta = {
             "build": {
                 "skip": "True",
-                "script": "python setup.py install",
+                "script": "python -m pip install",
```
beckermr, line 1005: Same here.

tests/test_lint_recipe.py
```diff
@@ -1002,14 +1002,14 @@ def test_missing_build_number(self):
...
                 "number": 0,
             }
         }
         lints, hints = linter.lintify_meta_yaml(meta)
         self.assertNotIn(expected_message, lints)
 
-        meta = {"build": {"skip": "True", "script": "python setup.py install"}}
+        meta = {"build": {"skip": "True", "script": "python -m pip install"}}
```
beckermr, line 1012: Same here.

tests/test_lint_recipe.py
```diff
@@ -1444,7 +1444,7 @@ def test_bad_subheader(self):
         meta = {
             "build": {
                 "skip": "True",
-                "script": "python setup.py install",
+                "script": "python -m pip install",
```
beckermr, line 1447: same here.

tests/test_lint_recipe.py
```diff
@@ -1454,7 +1454,7 @@ def test_bad_subheader(self):
         meta = {
             "build": {
                 "ski": "True",
-                "script": "python setup.py install",
+                "script": "python -m pip install",
```
beckermr, line 1457: same here.

beckermr (changes requested):
This PR is looking great! Note however that the references to setup.py in the test suite and associated data are not related to the packaging of conda-smithy itself.

tests/recipes/click-test-feedstock/recipe/meta.yaml
janl: yup, hence me raising these cases in the PR body, thanks for clarifying! I made these separate commits so I can revert them easily.

janl: I’ve done reverts for now, but I was wondering if you want all these as separate commits, or should I rebase these
- [ ] into a single comit
- [ ] separate commits are good, just take out the original commits instead of a revert

beckermr: commit hygene doesn't bother me - we can squash on merge

janl: Thanks!

Issue #1949, opened 2024-06-12
ENH: extend free_disk_space to windows?


h-vetinari:
I'm running into
```
OSError: [Errno 28] No space left on device
```
on a windows build, however, `free_disk_space` is documented to only work on linux:
https://github.com/conda-forge/conda-smithy/blob/0361df53e4ffb856ddd9733abae69b6e17b11ce7/conda_smithy/schema.py#L139-L149

Presumably there are some low-hanging fruit we could clean out also on windows images?

CC @carterbox @bkpoon @mikemhenry @beckermr @jakirkham as people who touched these lines in recent times.

bkpoon:
On Windows, I switch the build to use `C:\` instead of `D:\`. There should be at least 70 GB available. To move the build, you modify `conda-forge.yml` to have
```
azure:
  settings_win:
    variables:
      CONDA_BLD_PATH: C:\\bld\\
```
I used to use a directory in `C:\Miniconda`, but I eventually hit an issue with long path names. `C:\bld` is short enough. I suggested this for the `pytorch` Windows build here (https://github.com/conda-forge/pytorch-cpu-feedstock/pull/231#issuecomment-2041186886).

jakirkham:
There is an implementation of `azure.free_disk_space` in PR: https://github.com/conda-forge/conda-smithy/pull/1966

However it has a runtime of roughly 15mins, which isn't pleasant. Though perhaps acceptable when the alternative is a crash

If others have ideas on how to get the deletion to run faster, would be open to feedback in that PR

PR #1966, opened 2024-06-28
Extend `azure.free_disk_space` to Windows as well

jakirkham:
In addition to Linux, enable `azure.free_disk_space` on Windows. This helps free up some space in cases where it is needed. Also it can fix build issues where a system Python is accidentally pulled into builds causing hard to diagnose issues.

Also switch Miniforge downloads to use PowerShell. This removes the reliance on the system Python for this step (making the aforementioned cleanup possibly). Additionally it slims the step down a bit and makes it a bit easier to parse.

<hr>

Checklist
* [x] Added a ``news`` entry


h-vetinari:
👍 

Xref #1949, which also contains a work-around.

jakirkham:
Thanks Axel! 🙏

Was finding the deletion time on Azure to be pretty slow. Tried every trick I could think of with `robocopy`

If you know of better ways to delete on Windows, please let me know

Agree changing the build directory path is worth considering

h-vetinari (commented):
We probably need to guard this on `CONDA_BLD_PATH` somehow? I don't think it makes sense to spend time cleaning up `C:\` if the build then happens in `D:\` by default
https://github.com/conda-forge/conda-smithy/blob/058670dd19ea43cb3c33a774b7f15dac511b99bb/conda_smithy/schema.py#L226

Speaking of `D:\`, I'm not sure there'd even be anything to clean up there? So I guess we'd have a three-stage ladder of escalating options, each taking more time but providing more disk space:
* default (build on `D:\`; fastest, least space)
* switch `CONDA_BLD_PATH` to `C:\` (slower, more space)
* also set `azure.free_disk_space` (slowest, most space)

jakirkham:
Agreed

We may also want to put this under another option like `azure.free_disk_space: win` so that doing `true` doesn't add this (similar to the `docker` option on Linux)

PR #1967, opened 2024-06-28
HashableDict is removed in conda_build 24.7

minrk:
HashableDict is [deprecated in conda-build 24.5.0](https://docs.conda.io/projects/conda-build/en/24.5.x/release-notes.html#id4) in favor of `frozendict.deepfreeze`, removed in 24.7.

ref: https://github.com/conda/conda-build/pull/5284

rerender's working for me with this change and conda-build 24.5.1.39_gc49182fe. Rerender time is reduced by about 90%  for me with 24.7.x, so I'm looking forward to this release.

PR #1968, opened 2024-06-28
reduce input variants to only _used_ input variants

minrk:
before recomputing output metadata, which involves a number of expensive $O(\texttt{len(inputvariants)})$ operations.

reduces `get_used_vars` and `Metadata.copy()` time dramatically because `input_variants` can be several thousand items long to produce 10s of variants, since it is the cartesian product of all possible variants across conda-forge, including every single unused combination.

Combining this PR and the following upcoming optimizations to conda-build 24.7:

- https://github.com/conda/conda-build/pull/5392
- https://github.com/conda/conda-build/pull/5384
- https://github.com/conda/conda-build/pull/5281
- (also requires #1967 for compatibility with conda-build 24.7)

render time for the petsc4py feedstock, which has 13,824 input variants to produce 72 linux-64 builds, is reduced from over 30 minutes to 55 seconds to produce the same result:

Screenshot 2024-06-28 at 13 58 11

Checklist
* [ ] Added a ``news`` entry


beckermr:
Thanks for this. I don't quite follow it and need to think on it a bit so I understand. I'm concerned the test suite may not catch all of the edge cases.

minrk:
Looks like it doesn't quite work in every case and tests catch it, despite working in all of the recipes I tested.

I think perhaps conda-smithy needs to do its own detection of used vars and reduce them before passing them to cond-build's render. Because the _vast_ majority of variants are never used in all conda-forge recipes (every recipe computes every variant for every input parameter across all of conda-forge _before_ checking if anything will be used, and then caries this exploded matrix through the whole render process, accounting for ~90% of the remaining render time).

I think the upstream PR is probably not going to work either: https://github.com/conda/conda-build/pull/5392

In general, conda-build seems to assume that every input variant will be used, which doesn't match conda-forge where the input config is used for every package and has numerous unused dimensions which should eventually be dropped.

beckermr:
Thanks. The detection of used variables is itself expensive and so we'll need to think on this a bit.

conda_smithy/configure_feedstock.py
```diff
@@ -878,6 +881,39 @@ def migrate_combined_spec(combined_spec, forge_dir, config, forge_config):
...
+
+    # from distribute_variants
+    # explode variants dict to list of variants
+    variants = initial_variants = conda_build.variants.get_package_variants(
+        metadata, variants=input_variants
+    )
+    logger.debug(f"Starting with {len(initial_variants)} variants")
+    metadata.config.variant = variants[0]
+    metadata.config.variants = variants
+    all_used = metadata.get_used_vars(force_global=True)
```
beckermr, line 902: What does `force_global=True` do here? That seems like it is worth a comment.

conda_smithy/configure_feedstock.py
```diff
@@ -878,6 +881,39 @@ def migrate_combined_spec(combined_spec, forge_dir, config, forge_config):
...
+    # trim unused dimensions, these cost a lot in render_recipe!
+    all_used.update(ALWAYS_KEEP_KEYS)
+    all_used.update(
+        {"target_platform", "extend_keys", "ignore_build_only_deps"}
+    )
+    new_input_variants = input_variants.copy()
+    for key, value in input_variants.items():
+        if len(value) > 1 and key not in all_used:
```
beckermr, line 910: Is there a reason to not remove keys with dimension 1 here too if they are unused?

beckermr (changes requested):
Some questions.

conda_smithy/configure_feedstock.py
minrk: No, if this is rigorous and always guarantees to get all used keys every time. Since increased dimensionality is the real source of cost, only considering keys which increase dimensionality reduces the chances of imprecise logic having an impact (see comment below for a little more detail)

conda_smithy/configure_feedstock.py
minrk: Good call. force_global [looks at the whole recipe](https://github.com/conda/conda-build/blob/d1ba529d606ecc886331d1672f49b83c5c3e81bb/conda_build/metadata.py#L2922-L2928) for used variables, rather than parsing outputs individually and only looking in requirements. It's a broader, simpler search that will find used_vars in more complex template logic.

minrk:
The upstream PR https://github.com/conda/conda-build/pull/5392 no longer reduces the variant list because I don't understand what conda-build promises to do with unused variants. Instead it reduces the _cost_ of a very large variant list, so I think it has a better chance of being acceptable.

conda_smithy/configure_feedstock.py
minrk: hundreds of other unused keys contributes to the cost as well, so removing all unused keys here would be ideal and have a benefit. But I'm not confident enough that the logic is perfect every time to apply this across the board.

beckermr (changes requested):
Looking good. Do we have a test that covers a multidimensional key only in an output? This seems like the relevant corner case given the discussion.

minrk:
> Do we have a test that covers a multidimensional key only in an output?

yes, [`test_conda_build_api_render_for_smithy`](https://github.com/conda-forge/conda-smithy/blob/dd6091b80e8bf1f3f66612d773b3127f490b6b94/tests/test_configure_feedstock.py#L1953) exercises this with [multiple_outputs](https://github.com/conda-forge/conda-smithy/tree/dd6091b80e8bf1f3f66612d773b3127f490b6b94/tests/recipes/multiple_outputs) which has multiple variants for each of two outputs.

I believe I have found a case where this will miss a variant: if a variable is only used in the build script of an output in a multi-output recipe, but nowhere in meta.yaml, it will be missed by `get_used_vars(global=True)`. I'm not sure if/when it makes sense to write a recipe like that and I'm not sure conda-smithy needs to handle it. If we do need to handle it, it may be tricky to do without constructing all of the outputs to get `build.script` (which is what we want to avoid doing until _after_ we've reduced the variant matrix), but calling [`find_used_variables_in_shell_script`](https://github.com/conda/conda-build/blob/d1ba529d606ecc886331d1672f49b83c5c3e81bb/conda_build/variants.py#L772) on our own glob of `recipe/*.sh` (and equivalent for bat) might do it. The difference would be that it would search _all_ scripts, while conda-build's logic only searches the ones that are actually `output.script` for an output. I'm not sure if the `.sh` extension can be assumed, even though it's ~always used.

beckermr (changes requested):
Looking at this again, I think in order to merge we need to do either one of the following two things

1. Fix the code getting all used vars check so it is reliable
2. Add a test showing exactly what fails and ensure we are ok with ignoring that case

beckermr:
Given the latest change, I think we can discard the len(1) > 1 check. 

Also, we need to add `.bat` file to the tests if we don't have one.

beckermr (approved):

minrk:
I haven't removed the `len(variant) > 1` check in this PR, and when I tried it, it removed the mpi _provider_ pins in petsc4py. I.e. having:

```yaml
host:
- { mpi }
```

does not result in the pin for the  _value_ of `mpi` (`mpich` or `openmpi`) being in the `used_vars` found by `reduce_variants`. That does mean if `mpich` itself had _multiple_ version values to build, it would render incorrectly. The following pattern would render correctly, if we had done that:

```yaml
host:
- mpich  # { mpi == 'mpich' }
```

I can try to add a test for that tomorrow, if you think we should, or if that means we should hold off on this. One minute is a lot less dire to optimize than 30.

beckermr (changes requested):
Yes. Let's add more tests.

PR #1969, opened 2024-06-30
Add version update skip field to schema

mfisher87:
Checklist
* [x] Added a ``news`` entry

I believe this is the needed change to correspond with this one: https://github.com/regro/cf-scripts/pull/2646

Here's a feedstock I maintain which is failing lint due to leveraging this feature: https://github.com/conda-forge/quarto-feedstock/pull/50

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,7 @@ Below are a few things we ask you kindly to self-check before getting a review. 
 -->
 Checklist
 * [ ] Added a ``news`` entry
+* [ ] Regenerated schema JSON if schema altered (`python conda_smithy/schema.py`)
 
 <!--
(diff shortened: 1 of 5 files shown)


mfisher87:
@beckermr would you mind giving your input on this change? I think I found the right spot, but I assume I'm missing something :)

beckermr (changes requested):
This needs a news item and I am pretty sure we need to regenerate the json schema after this change.

beckermr:
Also thanks for the PR!

mfisher87:
@beckermr I'm not used to this "news" format in use here, can you provide more details or a link to documentation? I'll also update the pull request template with that link as part of this PR to benefit future contributors :)

mfisher87:
Same for generating the JSON schema actually :grin:

beckermr:
Try running 

```
python conda_smithy/schema.py
```

from the top-level directory of the git repo and it should dump new json files that can be added.

beckermr (approved):

PR #1956, opened 2024-06-20
docs: expand developer documentation

espy:
- Add more documentation for local development, testing, linting and formatting

PR #1975, opened 2024-07-12
Modernize testsuite

hulkoba:
Hello 👋!

As part of the STF work for Conda, this PR introduces several improvements to our testing framework by transitioning from unittest to pytest, leveraging pytest-mock for mocking, and adopting snapshot testing with syrupy.

Note that the testing section in the readme is updated at https://github.com/conda-forge/conda-smithy/pull/1956

Thank you!

Checklist

- [x] Added a news entry

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     rev: v4.6.0
     hooks:
     -   id: check-yaml
-        exclude: (tests/recipes|conda_smithy.recipe)
+        exclude: (tests/__snapshots__|tests/recipes|conda_smithy.recipe)
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/psf/black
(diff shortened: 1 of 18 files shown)


hulkoba:
Somehow ignoring the snapshots from linting doesn't work in this branch as it does for the one on the fork https://github.com/neighbourhoodie/conda-smithy/pull/13

beckermr (changes requested):
So the changes to the tests as written are wrong in many cases. 

Many of the tests verify the correctness of the results as opposed to being regression tests handled by snapshotting. For tests that verify result correctness, we hard code the correct answer clearly in the test case. In some cases we even include comments on why the answer is correct. The changes here obfuscate that. 

We need to separate these two cases out and only apply snapshotting to true regression tests that only care if the code output did not change and have no constraints on what is in that output.

beckermr:
I'd say all of them should remain hard coded.

hulkoba:
Hey all :wave: 

Since the snapshot testing is still discussed, I was wondering if you are still interested in the removal of `unittest` in favor of `pytest`?

If so, I could take those commits from this PR and create a new one.
What do you think?

beckermr:
Yes, for sure we should move to pytest!

PR #1979, opened 2024-07-17
feat: add rattler build linting

nichmor:
Checklist
* [ ] Added a ``news`` entry


nichmor:
Hey! @beckermr and @isuruf !
I've added back linting of recipe.yaml.
Could you please take a look?
Thanks!

beckermr (changes requested):
Can we break this down into a set of smaller PRs?

I'd suggest something like

1. A set of PRs to refactor the linting code into separate functions. These would be pure refactors with no functional changes.
2. A set of PRs to add the new recipe format linting functions (as needed) that mirror the refactored ones above that need changes for the new format.
3. A final PR that builds the end to end linter.

Each of these PRs should aspire to have a net diff of at most ~500 lines of code changed (either added or deleted).

nichmor:
> My two cents: now that we are going to have two implementations of many similar checks, is it time to add identifiers to each lint/hint rule in the same way that e.g. [`ruff`](https://docs.astral.sh/ruff/rules/) or [`shellcheck`](https://www.shellcheck.net/wiki/SC1000) do it? I imagine (ending up) having documentation for each rule explaining the rationale and workarounds if needed for example.

hey @jaimergp ! I think it is a very good point!

I think we could have at least two explicit categories - hints ( H ) and lints ( L ) and then something specific for rattler-build and conda-build lint ( similar to how there are pylint or flake lints )

conda_smithy/lint_recipe.py
```diff
@@ -164,20 +200,25 @@ def lintify_meta_yaml(
...
+    lint_should_be_empty_line(recipe_fname, lints)
 
     # 12: License family must be valid (conda-build checks for that)
-    try:
-        ensure_valid_license_family(meta)
-    except RuntimeError as e:
-        lints.append(str(e))
+    if not is_rattler_build:
+        try:
+            ensure_valid_license_family(meta)
```
wolfv, line 208: can you add a comment that `rattler-build` enforces spdx and therefore doesn't need a `license_family`?

tests/test_lint_recipe.py
```diff
@@ -336,7 +612,7 @@ def test_bad_about_license(self):
                 "license": "unknown",
             }
         }
-        lints, hints = linter.lintify_meta_yaml(meta)
```
beckermr, line 339: We cannot push api changes like this without a major version bump and a lot of discussion.  further, if any api changes were pushed in the previous PRs, they need to be backed out.

cc @wolfv

beckermr (changes requested):

tests/test_lint_recipe.py
wolfv: That seems like a internal API, correct?
I haven't found any usage outside of conda-smithy with a Github code search. 

We'll double check to not break any other project import `conda-smithy`.

tests/test_lint_recipe.py
```diff
@@ -94,6 +150,33 @@ def test_osx_lint(where):
...
+
+    with tmp_directory() as recipe_dir:
+        with io.open(os.path.join(recipe_dir, "recipe.yaml"), "w") as fh:
+            fh.write(
+                """
+                package:
+                   name: foo
+                requirements:
+                  run:
+                    # since we're in an f-string: double up braces (2->4)
```
wolfv, line 164: I don't see 4 brackets, though.

tests/test_lint_recipe.py
beckermr: Internal APIs begin with an underscore in Python. Any other API is always external and cannot be changed without a major version change and discussion. We cannot change them.

cc @conda-forge/core

nichmor:
I've used old API name so it will not create any breaking changes. 
Could you please @beckermr review it again?

beckermr (changes requested):
This PR is still way too big. Please submit smaller PRs.

nichmor:
@beckermr I can strip tests out of this PR, so it would be just ~500 changes.
and following PR will add them back so you can potentially focus only on tests.

wdyt?

nichmor:
I understand the rationale behind reviewing code and tests together in smaller chunks. 
However, splitting this into multiple PRs with subsets of lints presents some challenges in my opinion:

* It will lead to a longer overall review process, as each PR would need to go through the entire cycle. ( and I fear that we are already in the middle of the week so we have a higher slippage chance ) 

* Tracking the progress and ensuring all lints are eventually included would be more complex for me

* Maintaining consistency across PRs might be difficult, as the style of the lints are the same, but if in one of the PR's we decided to change the "stylistic" of lint's I would need to repeat in the waterfall manner for all of them again, which leads to the first point.

Maybe we could do some iterations of reviewing this PR? so all the discussions will be here and it would be easier for me to keep track of them and apply.

nichmor:
> I'd go the other direction. Add a subset of the lints in each PR with the tests for those lints. Then we can look at the code+tests together.

after some additional thought I also have a different plan for this:
* I can have a PR that modify only hints
*  2-3 PR's that add 10 new lints for rattler-linter

wdyt?

beckermr:
I am aware that making smaller PRs in the style I asked for is more work for the person making the PRs. On the other hand, the big PRs or reviewing subsets is more work for the reviewers.

The thing to remember is that I am not paid to do the review work here. While I like the new recipe format, I'm personally ok with it taking much longer to be used in conda-forge than the time scale you appear to want. 

I very much appreciate all of the work being done by the prefix.dev folks on conda-forge and very much want to see that work continue. 

We do however need to find a sustainable path to retooling conda-forge for the new recipe format. In a recent PR where I raised these concerns, we had another member of prefix do the review. I'm only marginally ok with this, we missed API breaks that should have been caught, and instead would very much prefer you all to take a bit more time and care with how you ask for volunteer time through large PR review requests.

beckermr:
as long as they group tests + code and are around ~500 lines or, I'm happy with them

wolfv:
I am closing this one since we merged all the items in multiple small PRs :)

PR #1983, opened 2024-07-23
fix: move back needed families

nichmor:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -69,6 +69,8 @@ from conda_smithy.validate_schema import validate_json_schema
 
 from conda_smithy.utils import render_meta_yaml, get_yaml
 
+NEEDED_FAMILIES = ["gpl", "bsd", "mit", "apache", "psf"]
+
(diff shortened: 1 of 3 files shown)


wolfv:
This is to fix a potential breakage here: https://github.com/regro/cf-scripts/blob/c5c49476e9f4b292c1ec64c0cc53194a9e92efae/conda_forge_tick/migrators/license.py#L15

beckermr (commented):
Just to be clear, you are free to move the base definition of NEEDED_FAMILIES to the `.utils` module combined with importing it into `.lint_recipe`. That won't break the API. Removing it all together from `.lint_recipe` is the issue.

nichmor:
> Just to be clear, you are free to move the base definition of NEEDED_FAMILIES to the `.utils` module combined with importing it into `.lint_recipe`. That won't break the API. Removing it all together from `.lint_recipe` is the issue.

It is true but I think it's better to keep them in one place and just pass as some arguments

PR #1985, opened 2024-07-24
feat: add rattler-build noarch hints

wolfv:
Will have to credit @nichmor, but wanted to check the tests in CI :)


wolfv:
Actually, we should start from the top to make it work. Will revive this one later on :)

conda_smithy/lint_recipe.py
```diff
@@ -536,17 +539,40 @@ def _format_validation_msg(error: "jsonschema.ValidationError"):
...
-def main(recipe_dir, conda_forge=False, return_hints=False):
+def main(
+    recipe_dir, conda_forge=False, return_hints=False, feedstock_dir=None
+):
     recipe_dir = os.path.abspath(recipe_dir)
-    recipe_meta = os.path.join(recipe_dir, "meta.yaml")
-    if not os.path.exists(recipe_dir):
-        raise IOError("Feedstock has no recipe/meta.yaml.")
+    build_tool = CONDA_BUILD_TOOL
+    if feedstock_dir:
```
beckermr, line 547: Should we default feedstock_dir to "{recipe_dir}/.." if it is not given?

conda_smithy/linter/rattler_linter.py
```diff
@@ -20,3 +22,39 @@
...
+                isinstance(requirement, dict)
+                for requirement in section_requirements
+            ):
+                no_arch_possible = False
+                break
+
+        if no_arch_possible:
+            hints.append(
+                "Whenever possible python packages should use noarch. "
+                "See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-builds"
```
beckermr, line 59: We may want to have this hint message be pulled from a single global value since if we update it in one spot, we'll want to update it in both.

beckermr (changes requested):
Two comments. PR looks great and thanks for the nice change diff to look through!

conda_smithy/lint_recipe.py
nichmor: I would prefer to not doing this so we can enable linting of recipe.yaml by explicitly passing feedstock-dir. Also, making it a default would require additional changes to the test suite

beckermr: The defaults can be overriden at the command line, right? I don't follow.

nichmor: It can be, but we have the same behavior in conda-forge-ci-setup: `https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/316/files#diff-2d7dc8fce520979cf844640e36ee6becb8e393f62efc53fea141ac8629f2b6eaR362`

I would prefer them to be explicit until we fully integrate the `rattler-build` so we could gradually enabling them

beckermr (approved):

PR #1992, opened 2024-07-26
Add conda recipe format v2 stdlib lint support

wolfv:
~Note: the addition of `recipe_dir` in the lint_stdlib function breaks API.~

No API breaks here.

### DIFF
--- a/tests/test_lint_recipe.py
+++ b/tests/test_lint_recipe.py
@@ -8,6 +8,7 @@ import unittest
 import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
+from pathlib import Path
 
 import github
 import pytest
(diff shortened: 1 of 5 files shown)


beckermr (changes requested):
We should pass the variant.yaml path as the conda_build_config_filename. This won't break the API and is consistent with how the function worked previously.

wolfv:
Good point. I'll refactor it later and update the PR :)

wolfv:
I removed the API breakage. I think this one should be good as well now.

tests/test_lint_recipe.py
```diff
@@ -329,6 +358,144 @@ def test_license_file_empty(is_rattler_build):
...
+        (["10.9", "11.0"], ["10.13", "11.0"], ["10.12"], "You are"),
+        # only sdk -> no warning
+        (None, None, ["10.13"], None),
+        (None, None, ["10.14", "12.0"], None),
+        # only sdk, but below global baseline -> warning
+        (None, None, ["10.12"], "You are"),
+        (None, None, ["10.12", "11.0"], "You are"),
+    ],
+)
+def test_rattler_cbc_osx_hints(
```
beckermr, line 412: textwrap.dedent can be used here to make things a bit more legible.

beckermr (changes requested):

tests/test_lint_recipe.py
wolfv: agreed. We just copied the test from a few lines above though :)

nichmor: applied dedent

wolfv: thanks @nichmor

beckermr: Ahhhh sorry for the random request then. IDK what we did for the other one. Thanks for fixing here. Much appreciated!

beckermr (approved):

PR #1993, opened 2024-07-26
fix: regenerate schema

wolfv:
I ran `python conda_smithy/schema.py`.

### DIFF
--- a/conda_smithy/data/conda-forge.yml
+++ b/conda_smithy/data/conda-forge.yml
@@ -11,19 +11,19 @@ azure:
     pool:
       vmImage: ubuntu-latest
     swapfile_size: 0GiB
-    timeoutInMinutes: 360
+    timeout_in_minutes: 360
     variables: {}
   settings_osx:
     pool:
       vmImage: macOS-12
     swapfile_size: null
-    timeoutInMinutes: 360
+    timeout_in_minutes: 360
     variables: {}
   settings_win:
     pool:
       vmImage: windows-2022
     swapfile_size: null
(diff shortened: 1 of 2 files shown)


conda_smithy/data/conda-forge.yml
```diff
@@ -11,19 +11,19 @@ azure:
     pool:
       vmImage: ubuntu-latest
     swapfile_size: 0GiB
-    timeoutInMinutes: 360
+    timeout_in_minutes: 360
```
ytausch, line 14: I am a bit confused how this will affect us. This change appeared because the corresponding Pydantic model field was renamed from `timeoutInMinutes` to `timeout_in_minutes` in the ruff PR, so that is right, @wolfv.

To not change our schema though, I set this alias here: https://github.com/conda-forge/conda-smithy/blob/0bf152c460bf0e73f39274fdfd268a7e182892c6/conda_smithy/schema.py#L111-L115

However, because we save the default values with `model_dump`, which [by default does not respect aliases](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump) this defaults yaml file now has a different field name.

This actually breaks the logic setting field defaults, I suppose:

https://github.com/conda-forge/conda-smithy/blob/0bf152c460bf0e73f39274fdfd268a7e182892c6/conda_smithy/configure_feedstock.py#L2170-L2205

ytausch:
I think merging this might have broken conda-smithy.

ytausch:
See the PR which is linked above.

PR #1994, opened 2024-07-26
Revert "fix: regenerate schema"

ytausch:
These changes were incorrect because the azure runners expect the field name to be `timeoutInMinutes`, not `timeout_in_minutes`.


ytausch:
Oops, I am wrong here. Will leave a comment at #1993

PR #1997, opened 2024-07-29
start using inline-snapshot

wolfv:
Let's discuss! :)

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -60,7 +60,6 @@ def test_init_with_custom_config(py_recipe, request):
     init_obj = cli.Init(subparser)
     recipe = py_recipe.recipe
     # expected args object has
-    print("Type: ", py_recipe.config["conda_build_tool"])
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -82,9 +82,18 @@ def test_init_with_custom_config(py_recipe):
     data = yaml.safe_load(
         open(os.path.join(destination, "conda-forge.yml")).read()
     )
-    assert data.get("bot") is not None
-    assert data["bot"]["automerge"] is True
-    assert data["bot"]["run_deps_from_wheel"] is True
+
+    if data.get("conda_build_tool") == "rattler-build":
```
nichmor, line 86: wdyt about using "rattler-build" value from conda_smithy.utils module?

wolfv: Good point.

wolfv:
Benefits: updating snapshots is really easy and nice. Adding snapshot tests is also trivial.
Downside: the tests that I converted here are now more verbose.

PR #2001, opened 2024-07-30
feat: wip on noarch and runtime deps

nichmor:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/news/2001-lint-noarch-deps.rst
+++ b/news/2001-lint-noarch-deps.rst
@@ -1,4 +1,3 @@
 **Added:**
 
 * add lint for noarch and runtime deps (#2001)
-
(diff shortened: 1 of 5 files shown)


beckermr (changes requested):
We do allow platform-only selectors in noarch recipes. Our current linter should have that edge case as a guide.

nichmor:
> We do allow platform-only selectors in noarch recipes. Our current linter should have that edge case as a guide.

added lint for noarch platforms use-case. could you please take a look?

wolfv:
Fixed up the merge conflicts and rebased on main.

news/2001-lint-noarch-deps.rst
```diff
@@ -0,0 +1,3 @@
+**Added:**
+
+* add lint for noarch and runtime deps (#2001)
```
beckermr, line 3: Let's add this the single item we already have going.

beckermr (changes requested):
One nit.

PR #2018, opened 2024-08-05
fix pin_subpackage / pin_compatible lints with v2 recipes

wolfv:
This PR fixes how we lint `pin_subpackage` and `pin_compatible`.


beckermr (changes requested):
Please add a news item or append onto the one we already have.

wolfv:
Added a news entry.

PR #2035, opened 2024-08-12
Add a bash/sh/zsh guard to avoid users running build-localy with them

hmaarrfk:
I keep typing

```
bash build-\tab
```

which causes all kinds of problems. I think this would help me and likely others

Checklist
* [ ] Added a ``news`` entry <--- Is this needed?


conda_smithy/feedstock_content/build-locally.py
```diff
@@ -1,8 +1,10 @@
-#!/usr/bin/env python3
+#!/bin/sh
+'''exec' "python3" "$0" "$@" #'''
```
hmaarrfk, line 2: Hmm this is even cooler.

My fear is that this encourages bad habits. but ok!

hmaarrfk:
Well CIs are green, so this is here if there is some enthusiasm around it.

conda_smithy/feedstock_content/build-locally.py
```diff
@@ -1,8 +1,11 @@
-#!/usr/bin/env python3
+#!/bin/sh
+"""exec' "python3" "$0" "$@" #"""  # fmt: off # fmt: on
```
ocefpaf, line 2: This broken the script for those calling it with ./build-locally.py

Issue #2034, opened 2024-08-12
Pass `--extra-meta` for rattler-build


nichmor:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

Right now we don't pass the `--extra-meta` flag for osx/linux builds: https://github.com/conda-forge/conda-smithy/blob/5dafc25e86b8a59887e7b4f885a1a092233b8bf6/conda_smithy/templates/run_osx_build.sh.tmpl#L117

We need to fix this and pass them as we do for windows

### Installed packages

```shell
The error can be found on new rattler-build feedstock
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1000560&view=logs&jobId=a70f640f-cc53-5cd3-6cdc-236a1aa90802&j=a70f640f-cc53-5cd3-6cdc-236a1aa90802&t=6119ccbe-9301-594f-7c27-f792b80a7fcc
```

### Environment info

```shell
^ same https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1000560&view=logs&jobId=a70f640f-cc53-5cd3-6cdc-236a1aa90802&j=a70f640f-cc53-5cd3-6cdc-236a1aa90802&t=6119ccbe-9301-594f-7c27-f792b80a7fcc
```

PR #2037, opened 2024-08-16
fix: pass --extra-meta for rattler-build

nichmor:
Checklist
* [ ] Added a ``news`` entry

Closes https://github.com/conda-forge/conda-smithy/issues/2034

Actual run can be seen here: https://github.com/conda-forge/jolt-physics-feedstock/pull/2


beckermr (changes requested):
Does rattler not support the --no-test option? We need that for some cross-compile cases when emulation is turned off or we are working with osx-arm on osx-64.

baszalmstra:
There is [a `--no-test`](https://prefix-dev.github.io/rattler-build/latest/reference/cli/#modifying-result) for rattler-build but Im not entirely sure if it does the same thing as conda-build. Can you check @nichmor ?

nichmor:
should also fix: https://github.com/conda-forge/crane-feedstock/pull/6

baszalmstra:
@wolfv Can you verify that `--no-test` for `rattler-build` do the same thing as `--no-test` for `conda-build`?

wolfv:
I did just double check, and rattler-build unconditionally skips test execution which is what conda-build also does.

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -76,11 +76,11 @@ ulimit -n 1024
 # make the build number clobber
 make_build_number "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
 
-{% if test == "native" -%}
+{% if test == "native" and conda_build_tool != "rattler-build" -%}
 if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]] && [[ "${BUILD_WITH_CONDA_DEBUG:-0}" != 1 ]]; then
     EXTRA_CB_OPTIONS="${EXTRA_CB_OPTIONS:-} --no-test"
 fi
-{%- elif test == "native_and_emulated" -%}
+{%- elif test == "native_and_emulated" and conda_build_tool != "rattler-build" -%}
```
wolfv, line 83: is this correct? If we need `--no-test` then this is going to be missing in the rattler-build case, right?

nichmor: Initially it was disabled by me, but now that `--no-test` is working for rattler-build I need to remove this check

beckermr: I agree this looks wrong.

baszalmstra: I reverted these changes.

wolfv (approved):

wolfv:
Do we need to set the rattler-build version somewhere to `>=0.21.0`?

Looks good to me!

nichmor:
> Do we need to set the rattler-build version somewhere to `>=0.21.0`?
> 
> Looks good to me!

I think the minimal should be `>=0.20.0` as it was released there, and we need to set it here: https://github.com/conda-forge/rattler-build-conda-compat-feedstock/blob/main/recipe/meta.yaml#L24.

but we can merge this anyway,no?

wolfv:
Yep!

PR #2042, opened 2024-08-29
Fix corner-case in stdlib linting

h-vetinari:
Hit by @jaimergp in https://github.com/conda-forge/staged-recipes/pull/27419

### DIFF
--- a/conda_smithy/linter/lints.py
+++ b/conda_smithy/linter/lints.py
@@ -796,10 +796,15 @@ def lint_stdlib(
     # this check needs to be done per output --> use separate (unflattened) requirements
     for build_reqs in all_build_reqs:
         has_compiler = any(pat_compiler_stub.match(rq) for rq in build_reqs)
-        stdlib_search = (
-            "c_stdlib_stub" if recipe_version == 0 else "${{ stdlib"
+        stdlib_regex = (
+            # we need the C stdlib, not just any invocation of the stdlib jinja
+            "c_stdlib_stub"
+            if recipe_version == 0
+            else r"\$\{\{ stdlib\(['\"]c['\"]"
         )
-        if has_compiler and not any(stdlib_search in x for x in build_reqs):
+        if has_compiler and not any(
+            re.search(stdlib_regex, x) for x in build_reqs
+        ):
             if stdlib_lint not in lints:
                 lints.append(stdlib_lint)
 
--- a/conda_smithy/linter/lints.py
+++ b/conda_smithy/linter/lints.py
@@ -796,8 +796,10 @@ def lint_stdlib(
     # this check needs to be done per output --> use separate (unflattened) requirements
     for build_reqs in all_build_reqs:
         has_compiler = any(pat_compiler_stub.match(rq) for rq in build_reqs)
-        stdlib_stub = "c_stdlib_stub" if recipe_version == 0 else "${{ stdlib"
-        if has_compiler and stdlib_stub not in build_reqs:
+        stdlib_search = (
+            "c_stdlib_stub" if recipe_version == 0 else "${{ stdlib"
+        )
+        if has_compiler and not any(stdlib_search in x for x in build_reqs):
             if stdlib_lint not in lints:
                 lints.append(stdlib_lint)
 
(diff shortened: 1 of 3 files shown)


conda_smithy/linter/lints.py
```diff
@@ -796,8 +796,10 @@ def flatten_reqs(reqs):
     # this check needs to be done per output --> use separate (unflattened) requirements
     for build_reqs in all_build_reqs:
         has_compiler = any(pat_compiler_stub.match(rq) for rq in build_reqs)
-        stdlib_stub = "c_stdlib_stub" if recipe_version == 0 else "${{ stdlib"
-        if has_compiler and stdlib_stub not in build_reqs:
+        stdlib_search = (
+            "c_stdlib_stub" if recipe_version == 0 else "${{ stdlib"
```
beckermr, line 800: Do we always want "c" or maybe we search for "_stdlib_stub"? 

Right now for v1 recipes we'll catch anything whereas for v0 we'll catch only things that end in "c".

beckermr (changes requested):

conda_smithy/linter/lints.py
h-vetinari: IMO the v1 lint could/should be improved, because we really need the C stdlib here (which is why I had been using `c_stdlib_stub`). Otherwise, using
```yaml
requirements:
  build:
    - {{ stdlib("cxx") }}
    - {{ compiler("c") }}
```
would shut up the linter, but lead to broken packages (well, it would first lead to missing dependencies à la `cxx_stdlib_linux-64` unless `cxx_stdlib` gets defined in CBC, but still...)

beckermr: Ahhh I see. I guess we'll need a regex or to hardcode the possible cases.

h-vetinari: Do you want me to do that in this PR?

beckermr: whatever you think is best. I don't care much here.

h-vetinari: I ended up doing it anyway in https://github.com/conda-forge/conda-smithy/pull/2042/commits/31ececa8780c60b05cc3880ea562a283c78529c0. For the v0 recipes, a regex is kinda pointless (as we're searching for a literal string), but I think that's not a problem.

conda_smithy/linter/lints.py
```diff
@@ -796,8 +796,15 @@ def flatten_reqs(reqs):
     # this check needs to be done per output --> use separate (unflattened) requirements
     for build_reqs in all_build_reqs:
         has_compiler = any(pat_compiler_stub.match(rq) for rq in build_reqs)
-        stdlib_stub = "c_stdlib_stub" if recipe_version == 0 else "${{ stdlib"
-        if has_compiler and stdlib_stub not in build_reqs:
+        stdlib_regex = (
+            # we need the C stdlib, not just any invocation of the stdlib jinja
+            "c_stdlib_stub"
+            if recipe_version == 0
+            else r"\$\{\{ stdlib\(['\"]c['\"]"
```
beckermr, line 803: Won't this mis the original error for v1 recipes where the c stdlib was <something>_c?

beckermr (changes requested):

conda_smithy/linter/lints.py
h-vetinari: I don't understand your question. By v1 you mean the new format, right? For that, we're only changing the search from the literal `"${{ stdlib"` to the regex that also ensures the `("c"` part, which you had asked for.

I don't know where `something_c` would ever come into play, either for v0 or v1 (there's `c_stdlib_stub` in v0, which I'm using here).

beckermr: The linter originally failed because it didn't recognize m2w64_c as a valid stdlib. In the v1 recipe this would be "${{ stdlib('m2w64_c') }}". The code was written would fail to recognize this stdlib and thus have the same bug, but now on v1 recipes.

h-vetinari: yeah... that one's on me. I forgot the first part of the reason of this PR when I picked it up again. sorry

beckermr (approved):

PR #2044, opened 2024-08-29
Remove unittest

hulkoba:
Hello 👋!

As part of the STF work for Conda, this PR introduces several improvements to our testing framework by transitioning from `unittest` to `pytest` and leveraging `pytest-mock` for mocking.

It is split up from https://github.com/conda-forge/conda-smithy/pull/1975. 

Thank you!

Checklist

- [x] Added a news entry

### DIFF
--- a/tests/test_feedstock_io.py
+++ b/tests/test_feedstock_io.py
@@ -6,7 +6,6 @@ import shutil
 import stat
 import string
 import tempfile
-import unittest
 
 import git
 from git.index.typ import BlobFilter
@@ -46,7 +45,7 @@ def parameterize():
                 shutil.rmtree(tmp_dir)
 
 
-class TestFeedstockIO(unittest.TestCase):
+class TestFeedstockIO:
     def setUp(self):
         self.old_dir = os.getcwd()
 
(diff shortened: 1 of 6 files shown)


hulkoba:
Linting fails because of an unused var ` F841 Local variable git_mock is assigned to but never used`
But when I remove the line(s), the tests fail.

tests/test_feedstock_io.py
```diff
@@ -46,7 +45,7 @@ def parameterize():
                 shutil.rmtree(tmp_dir)
 
 
-class TestFeedstockIO(unittest.TestCase):
+class TestFeedstockIO:
```
beckermr, line 48: Unit test executes the setup method before tests. If we simply remove the class inheritance here, I am not sure if that will happen or not given the code now. Can we convert the setup to a test fixture and the methods to functions?

tests/test_feedstock_io.py
```diff
@@ -191,41 +184,33 @@ def test_copy_file(self):
...
+            assert write_text == read_text
 
             if repo is not None:
                 blob = next(repo.index.iter_blobs(BlobFilter(filename2)))[1]
                 read_text = blob.data_stream[3].read().decode("utf-8")
 
-                self.assertEqual(write_text, read_text)
+                assert write_text == read_text
 
     def tearDown(self):
```
beckermr, line 211: The same comment applies to teardown. This should be part of a test fixture IIUIC.

tests/test_lint_recipe.py
```diff
@@ -532,18 +532,18 @@ def test_v1_cbc_osx_hints(
             assert any(lint.startswith(exp_lint) for lint in lints)
 
 
-class TestLinter(unittest.TestCase):
+class TestLinter:
```
beckermr, line 535: The methods on this class should be converted to pure functions.

beckermr (changes requested):

PR #2053, opened 2024-09-05
Fix `./build-locally.py` in feedstocks

pkgw:
There's an unbalanced quotation mark in this script that makes the bash-to-python trampoline not actually work.

Checklist
* [x] Added a ``news`` entry
* [N/A] Regenerated schema JSON if schema altered (`python conda_smithy/schema.py`)


hmaarrfk:
Hmmm. Maybe black broke the previous. Have you tried running black locally?

PR #2068, opened 2024-09-29
feat: generate and use deploy keys

beckermr:
Checklist
* [x] Added a ``news`` entry

This PR adds a CLI utility and integrations into the webservices for SSH deploy keys. The idea is as follows

- A deploy key is made for each feedstock. The public part is registered with github with read+write permissions. The private part is stored in the github actions secrets.
- In the webservices github actions jobs, the deploy key is used to write to the repo.

The goal is to use the SSH key to be able to rerender github actions workflow files.

### DIFF
--- a/conda_smithy/cli.py
+++ b/conda_smithy/cli.py
@@ -825,6 +825,39 @@ class GenerateFeedstockToken(Subcommand):
                 )
 
 
+class SetFeedstockDeployKey(Subcommand):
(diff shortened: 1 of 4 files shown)


beckermr:
And deploy keys cannot push to forks with maintainer edit permissions. 🤦‍♂️ 

cc @jaimergp

PR #2091, opened 2024-10-13
fix moji-bake with rattler-build on Windows

wolfv:
We were using `conda.exe run ...` to execute `rattler-build, but it does something undesirable to the current code page on Windows which leads to "Moji-bake".

When calling rattler-build directly, things seem to work much better for printing fancy tables.

Tested in https://github.com/conda-forge/jolt-physics-feedstock/pull/6

Before: 
Screenshot 2024-10-13 at 12 08 26

After: 

Screenshot 2024-10-13 at 12 08 32


beckermr (changes requested):
Needs a news entry

wolfv:
Added news entry!

beckermr (approved):

PR #2099, opened 2024-10-17
Support `pixi` as an install tool

jaimergp:
Checklist
* [x] Added a ``news`` entry

Only an experiment for now. Testing at https://github.com/conda-forge/dav1d-feedstock/pull/21.

Will close https://github.com/conda-forge/conda-forge.github.io/issues/2323.


jaimergp:
This is mostly functional. See last runs at https://github.com/conda-forge/dav1d-feedstock/pull/21.

It provisions the base environment in a similar way to what `micromamba` does, but the dependency list is coming from the `pixi.toml` file. Since we have to populate that anyway, I also rendered a few convenience tasks for local users, so they can just do `pixi run build`, `pixi run rerender`, etc. Not perfect or a 1:1 correspondence to CI (for that they have `pixi run build-locally`), but a first approach to see how it feels.

To note:
- The installation paths do not correspond 1:1 either, but the "spirit" is there; e.g. on Windows we respect the drives so we get all the performance gains from D:.
- There's an ugly hack to limit which platforms get solved in CI (we patch `platforms` in the TOML file), but hopefully that'll get added soon in Pixi so we just use a flag like `--platforms=linux-64` for overrides.
- Pixi doesn't seem to offer PowerPC binaries?

Anyway, fun stuff! The linux timings are impressive; around 10s thanks to the shared `pkgs` cache. macOS and Windows stay under a minute.

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -64,7 +80,7 @@ export CONDA_LIBMAMBA_SOLVER_NO_CHANNELS_FROM_INSTALLED=1
 {{ conda_install_tool }} update --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
     pip {{ conda_install_tool_deps }} {{ conda_build_tool_deps }} {{ " ".join(remote_ci_setup_update) }}
 {%- endif %}
-{% if local_ci_setup %}
+{% if conda_install_tool != "pixi" and local_ci_setup %}
```
beckermr, line 83: Having local ci setup configuration/code split over multiple files on a single platform is not ideal. Is there a way we can leave this here?

conda_smithy/templates/pixi.toml.tmpl
```diff
@@ -0,0 +1,51 @@
...
+
+[project]
+name = "{{ feedstock_name }}"
+version = "{{ smithy_version }}"
+description = "Pixi configuration for conda-forge/{{ feedstock_name }}"
+authors = ["@conda-forge/{{ feedstock_name[:-11] }}"]
+channels = ["conda-forge"]
+platforms = {{ platforms }}
+
+[dependencies]
```
beckermr, line 13: Do we need to remove the ci setup package from conda in this list if local ci setup is in use?

beckermr (changes requested):
The handling of local_ci_setup might be flawed right now IIUIC. 

The current procedure on main installs the conda package, then force uninstalls it to leave the deps in the env, and then does the pip install without deps. 

The code in this PR does a pip install on top of the conda package. 

IDK if those two things are equivalent. 

We should test this in the ci-setip feedstock.

conda_smithy/templates/pixi.toml.tmpl
beckermr: Well actually not since we want the deps.

jaimergp:
Agreed, I haven't tested that, mostly left it there as a reminder of sorts. Worst case `conda-forge-ci-setup` can stay with `conda_install_tool = "micromamba"`. Let me give it a try in the feedstock though.

conda_smithy/templates/build_steps.sh.tmpl
jaimergp: Ok, I refactored this bit now, so we inject `conda` and `pip` for `local_ci_setup` cases (which is a rare exception anyway). Had to create a couple helpers to avoid duplication but it looks ok. Let me test it.

conda_smithy/templates/pixi.toml.tmpl
```diff
@@ -0,0 +1,46 @@
+# This file was generated automatically from conda-smithy. To update this configuration,
+# update the conda-forge.yml and/or the recipe/meta.yaml.
+# -*- mode: yaml -*-
```
beckermr, line 3: This mode is wrong.

beckermr (changes requested):

conda_smithy/configure_feedstock.py
```diff
@@ -2177,6 +2205,34 @@ def render_github_actions_services(jinja_env, forge_config, forge_dir):
...
+    if os.path.exists(ci_support_path):
+        for filename in os.listdir(ci_support_path):
+            if filename.endswith(".yaml"):
+                variant_name, _ = os.path.splitext(filename)
+                variants.append(variant_name)
+    platforms = [
+        platform.replace("_", "-")
+        for platform, service in forge_config["provider"].items()
+        if service
+    ]
```
beckermr, line 2225: Does this need to be a set?

jaimergp: It's guaranteed to be unique because we are only keeping the keys of a dictionary, unless the dictionary had mixed `_` and `-` so... yea, let's dedupl just in case.

jaimergp:
See https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/361 for ci-setup tests.

beckermr (approved):
LGMT to me assuming all of the feedstock tests work. I've been merging a bit fast I think since I've seen a few comments on PRs after I have merged them here. So i'll bump @conda-forge/core to look at this one too.

jaimergp:
No probs, thank you for the reviews! I'm in no rush here, was just following the momentum of the micromamba changes because I had all that mental model in the "hot cache" part of my brain :P 

For those checking the PR now or later, this comment summarizes the changes: https://github.com/conda-forge/conda-smithy/pull/2099#issuecomment-2422320419.

h-vetinari:
Haven't reviewed, but I think this is a great initiative and thanks a lot for pushing it forward. I checked out https://github.com/conda-forge/dav1d-feedstock/pull/21 and I think it's really cool (aside from the speed gains) that with pixi tasks, we'll provide a much simplified interface for building things locally, e.g. `pixi run build-linux_64_`. 😎

mfisher87:
> we'll provide a much simplified interface for building things locally, e.g. `pixi run build-linux_64_`

Oooo... amazing! I love that re-rendering is now a task, too!

beckermr:
@conda-forge/core it has been two weeks on this PR and there are no other reviews. Shall we merge?

PR #2124, opened 2024-11-07
Set rattler cache to /tmp when building linux from osx

hadim:
Checklist
* [x] Added a ``news`` entry


hadim:
Similar to https://github.com/conda-forge/staged-recipes/pull/28119

hadim:
LGTM

hadim:
> This should go in build_steps.sh.tmpl, I think. There's a rattler-specific env var we are already using in other platforms, which would be preferred instead of the system-wide XDG setting.

I can move the logic to `build_steps.sh.tmpl` but I don't see any rattler env var I can re-use there. I am also not aware of another env var to set the rattler cache than `XDG_CACHE_HOME`. 

Also, the current logic is specific to building linux from macOS with docker. Should I keep that same logic if moving to `build_steps.sh.tmpl`.

jaimergp:
> I am also not aware of another env var to set the rattler cache than `XDG_CACHE_HOME`.

See https://pixi.sh/v0.21.1/features/environment/#caching. I think `RATTLER_CACHE_DIR` applies here.

> Also, the current logic is specific to building linux from macOS with docker. 

Ah, ok, in that case build-locally is ok. Sorry!

conda_smithy/feedstock_content/build-locally.py
```diff
@@ -26,6 +26,12 @@ def setup_environment(ns):
             os.path.dirname(__file__), "miniforge3"
         )
 
+    # The default cache location might not be writable using docker on macOS.
+    if ns.config.startswith("linux") and platform.system() == "Darwin":
+        os.environ["CONDA_FORGE_DOCKER_RUN_ARGS"] = (
+            "-e XDG_CACHE_HOME=/tmp/rattler_cache"
+        )
```
jaimergp, line 33: Let's make sure we are not overwriting a maybe-present CONDA_FORGE_DOCKER_RUN_ARGS. Something like:

```suggestion
    if ns.config.startswith("linux") and platform.system() == "Darwin":
        os.environ["CONDA_FORGE_DOCKER_RUN_ARGS"] = (
            os.environ.get("CONDA_FORGE_DOCKER_RUN_ARGS", "")
            + " -e RATTLER_CACHE_DIR=/tmp/rattler_cache"
        )
```

hadim: Yes good catch. Thanks.

hadim:
I just tested locally and it does not work. It seems like `RATTLER_CACHE_DIR` only works with pixi.

- [The variable shows in the pixi source code.](https://github.com/search?q=repo%3Aprefix-dev%2Fpixi%20RATTLER_CACHE_DIR&type=code)
- [Not in the rattler repo](https://github.com/search?q=repo%3Aconda%2Frattler%20RATTLER_CACHE_DIR&type=code)
- [Not in the rattler-build repo](https://github.com/search?q=repo%3Aprefix-dev%2Frattler-build%20RATTLER_CACHE_DIR&type=code)

hadim:
I opened a ticket at https://github.com/prefix-dev/rattler-build/issues/1171

jaimergp:
Looks like [0.31 is out](https://github.com/prefix-dev/rattler-build/releases/tag/v0.31.0). Is that what we need for the new env var? Reopening just in case.

hadim:
Let's wait for 0.31 to make it in https://github.com/conda-forge/rattler-build-feedstock

hadim:
Actually for conda-smithy, you can merge @jaimergp !

news/rattler-cache.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* Set rattler cache to `RATTLER_CACHE_DIR=/tmp/rattler_cache` when building a linux package from a macOS machine with docker.
```
jaimergp, line 3:
```suggestion
* Set rattler cache to `RATTLER_CACHE_DIR=/tmp/rattler_cache` when building a linux package from a macOS machine with Docker. (#2124)
```

beckermr (approved):

jaimergp (approved):

PR #2126, opened 2024-11-08
fix: soften language for the `noarch: python` hint, add override note, and fix a bug for v1

beckermr:
Checklist
* [x] Added a ``news`` entry

Trying to address this [comment](https://github.com/conda-forge/conda-forge.github.io/issues/2351#issuecomment-2464374140) from @maresb.

### DIFF
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -281,13 +281,10 @@ def hint_noarch_python_use_python_min(
                 )
                 + hint
                 + (
-                    "\n   - You can override the `python_min` variable in the `conda_build_config.yaml`/`variants.yaml` "
-                    "if the package requires a newer Python version than the currently supported minimum "
-                    "version on `conda-forge`."
-                )
-                + (
-                    "\n   - You will need to [rerender the feedstock](https://conda-forge.org/docs/maintainer/"
-                    "infrastructure/#conda-forge-admin-please-rerender) after making any of these changes."
+                    "\n   - If the package requires a newer Python version than the currently supported minimum "
+                    "version on `conda-forge`, you can override the `python_min` variable by adding a "
+                    "Jinja2 `set` statement at the top of your recipe (or using an equivalent `context` "
+                    "variable for v1 recipes)."
                 )
             )
             hints.append(hint)
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -227,7 +227,8 @@ def hint_pip_no_build_backend(host_or_build_section, package_name, hints):
 
         if not found_backend:
             hints.append(
-                f"No valid build backend found for Python recipe for package `{package_name}` using `pip`. Python recipes using `pip` need to "
+                f"No valid build backend found for Python recipe for package `{package_name}` using `pip`. "
+                "Python recipes using `pip` need to "
                 "explicitly specify a build backend in the `host` section. "
                 "If your recipe has built with only `pip` in the `host` section in the past, you likely should "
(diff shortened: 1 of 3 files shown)


beckermr:
@conda-forge/core @maresb this is ready for review!

conda_smithy/linter/hints.py
```diff
@@ -266,21 +266,26 @@ def hint_noarch_python_use_python_min(
...
+                    f"\n  - For the `{section_name}` section of the recipe, you should usually use `{syntax}` "
+                    f"for the `python` entry."
                 )
 
         if hint:
             hint = (
                 (
-                    "noarch: python recipes should almost always follow the syntax in "
-                    "our [documentation](https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-python). "
+                    "`noarch: python` recipes should usually follow the syntax in "
```
maresb, line 276: I was thinking more something like this:

```suggestion
                    "**Experimental:** (A migration for this is planned for the near future.) `noarch: python` recipes should usually follow the syntax in "
```

Since they usually indicate some form of typo, I personally don't think of the linter "suggestions" / "hints" as optional, even though they technically are. So I'm trying to emphasize here more something like, "You don't need to do this now, but we'll be automatically doing most of this for you soon."

maresb: Thanks so much @beckermr, I really appreciate your diligence!!

beckermr: Yeah this is not experimental though. We passed a CFEP and this hint will remain in place even after the migration.

maresb: Ya, "experimental" was not the best choice of words.

Although verbose, perhaps `**Safe to temporarily ignore:**`?

beckermr: But one can always ignore hints. I am not quite sure what you are going for here. We've had some linter bugs and made changes to clarify things, but otherwise the hint is correct.

maresb: What I'm going for is to try and prevent casual recipe maintainers from needing to read CFEP-25.

> But one can always ignore hints.

Ok, sure, it's definitely partially my fault that I wasn't reading carefully enough. I didn't read anything from the linter's message above "For **recipe/meta.yaml**:" since it's mostly boilerplate, and thus I didn't realize. This linter is very quiet (which is very good!), so whenever I see something, I think that I messed something up and action is required.

Due to my inattentive reading I ended up going down a much bigger rabbit hole than was necessary, so I'm looking to preempt that for other maintainers.

But everything in the message is technically correct, so it's fine as-is.

beckermr: Ok thanks for working with me! Resolving.

beckermr:

beckermr:
OK this PR is ready to go again @conda-forge/core. I will wait through the weekend and merge on monday with a bugix release if I do not hear otherwise.

conda_smithy/linter/hints.py
```diff
@@ -250,12 +251,13 @@ def hint_noarch_python_use_python_min(
...
         ]:
-            test_syntax = syntax.replace("{{ python_min }}", "9999")
-
             if recipe_version == 1:
                 syntax = syntax.replace(
                     "{{ python_min }}", "${{ python_min }}"
                 )
+                test_syntax = syntax
+            else:
+                test_syntax = syntax.replace("{{ python_min }}", "9999")
```
maresb, line 260: @beckermr, is `9999` intentional? Looks like it could be remnants from a test?

beckermr: It is intentional.

PR #2142, opened 2024-11-19
Make sure $MINIFORGE_HOME exists during osx build

hadim:
Checklist
* [x] Added a ``news`` entry
* [ ] Regenerated schema JSON if schema altered (`python conda_smithy/schema.py`)

I noticed this issue when using pixi as conda install tool. The `mkdir` command is outside the pixi condition but I can also move it inside it if you prefer. Let me know.

PR #2146, opened 2024-11-19
Move `MINIFORGE_HOME` fix to `pixi` block

jakirkham:
Creating `MINIFORGE_HOME` up top breaks the Miniforge install step. As this appears this is only needed for `pixi` move to that block to limit the effect.

cc @hadim @beckermr

xref: https://github.com/conda-forge/conda-smithy/pull/2142#discussion_r1849286289

Checklist
* [x] Added a ``news`` entry


hadim:
Thank you @jakirkham and my bad for the trouble!

jakirkham:
Not at all and thank you for the `pixi` fix

There are just a few moving parts here given we now have 3 supported installers

h-vetinari (approved):

Issue #2145, opened 2024-11-19
`validate_recipe_outputs` fails with `pixi`


hadim:
### Solution to issue cannot be found in the documentation.

- [x] I checked the documentation.

### Issue

`validate_recipe_outputs` is looking for the built packages into a pixi env directory instead of conda-bld. It only happens on osx and not linux:

```
Adding in variants from internal_defaults
Adding in variants from argument_variants
Traceback (most recent call last):
  File "/Users/runner/work/1/s/.pixi/envs/default/bin/validate_recipe_outputs", line 11, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/runner/work/1/s/.pixi/envs/default/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/1/s/.pixi/envs/default/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/runner/work/1/s/.pixi/envs/default/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/1/s/.pixi/envs/default/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/1/s/.pixi/envs/default/lib/python3.12/site-packages/conda_forge_ci_setup/feedstock_outputs.py", line 158, in main
    distributions = built_distributions_from_recipe_variant(recipe_dir=recipe_dir, variant=variant)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/1/s/.pixi/envs/default/lib/python3.12/site-packages/conda_forge_ci_setup/utils.py", line 126, in built_distributions_from_recipe_variant
    for dist in built_distributions(subdirs=allowed_subdirs)
[...]

PR #2148, opened 2024-11-20
Set CONDA_BLD_PATH when using pixi during osx build

hadim:
Checklist
* [x] Added a ``news`` entry

This is an attempt to fix https://github.com/conda-forge/conda-smithy/issues/2145. You can see it in action at https://github.com/conda-forge/docling-parse-feedstock/pull/3

Note the fix is a simple and naive patch but looks fragile to me. Please let me know if you can think of another solution.

### DIFF
--- a/conda_smithy/templates/run_osx_build.sh.tmpl
+++ b/conda_smithy/templates/run_osx_build.sh.tmpl
@@ -6,17 +6,9 @@ source .scripts/logging_utils.sh
 
 set -xe
 
-MINIFORGE_HOME=${MINIFORGE_HOME:-${HOME}/miniforge3}
-MINIFORGE_HOME=${MINIFORGE_HOME%/} # remove trailing slash
-
-# Set CONDA_BLD_PATH when using pixi
-# to prevent the pixi env path from being used
-# during validation when looking for the built packages.
-# The export is required for the variable to propagate to
-#inspect_artifacts.
-{%- if conda_install_tool == "pixi" %}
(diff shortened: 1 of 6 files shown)


beckermr (dismissed):
I wonder if we should set this on all platforms?

Also, the code that finds the outputs is in the ci-setup feedstock. We need to consult that code to ensure what we are doing here works with it.

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -9,6 +9,15 @@ set -xe
 MINIFORGE_HOME=${MINIFORGE_HOME:-${HOME}/miniforge3}
 MINIFORGE_HOME=${MINIFORGE_HOME%/} # remove trailing slash
```
jaimergp, line 10: I think we can include [this blob](https://github.com/conda-forge/staged-recipes/blob/778e1c767ae07a9ba9eecc60e927325ce05f93d3/.scripts/run_osx_build.sh#L9-L11) from `staged-recipes` here instead.

jaimergp:
> I wonder if we should set this on all platforms?

Yea I think it wouldn't hurt. We are doing it in staged-recipes.

jaimergp:
Testing at https://github.com/conda-forge/docling-parse-feedstock/pull/5

beckermr:
I wonder why we set it in ci setup for win, but in smithy for Unix. I don't fully understand when stuff goes in one place versus the other. We should maybe clarify in future prs.

PR #2199, opened 2025-01-01
fix: Fix removing dir in feedstock_io, and add a test case

mgorny:
Checklist
* [x] Added a ``news`` entry

Fix `remove_file_or_dir()` to correctly remove directories from index, and add a test case to cover it.


mgorny:
Thanks, and sorry for my original mistake.

PR #2213, opened 2025-01-10
fix test with latest rattler-build

wolfv:
The strict parsing made this fail. With a `*` things should be fine! :)


beckermr (changes requested):
Needs a news item.

wolfv:
Added the news item, thx!

beckermr (approved):

PR #2217, opened 2025-01-13
Fix handling package version 0, refactor and deduplicate

### LINKED ISSUE #2215: Package version "0" causes a false positive

### Issue

If the recipe has:

```yaml
package:
  ...
  version: "0"
```

```
$ conda smithy lint recipe
recipe has some lint:
  Package version is missing.
```

Not a big deal but looks like something is misinterpreting `0` as `False`.

### Installed packages

```shell
n/a
```

### Environment info

```shell
n/a
```

mgorny:
Checklist
* [x] Added a ``news`` entry

Fix the boolean logic that caused version of 0 (that is a valid package version) to be incorrectly interpreted as missing, and emit linter errors.  While at it, deduplicate the recipe v0 linter to reuse the utility code added along with v1 lints, and refactor the v1 linter to reduce the number of redundant conditions and make the logic easier to follow.

Fixes #2215


mgorny:
Thanks!

Issue #2188, opened 2024-12-16
BUG: false negative for CFEP-25 linting


h-vetinari:
Found a situation where [there's](https://github.com/lucascolley/scipy-stubs-feedstock/blob/0f60173aa1e5e035bdafa8fd7014df572cd5a476/recipe/meta.yaml) an output
```yaml
  - name: scipy-typed
    build:
      noarch: python
    requirements:
      host:
        - python
      run:
        - python
        - scipy =={{ scipy_version }}
        - {{ pin_subpackage('scipy-stubs', exact=True) }}
    test:
      requires:
        - basedmypy
      imports:
        - scipy
```
and yet, the linter (from 3.45.0) says it's fine.

beckermr:
Yes this is a long standing missing feature. The noarch python linter never looked at outputs even before CFEP-25

mgorny:
Is this merely a matter of implementing that part, i.e. processing output requirements, or is there a more tricky reason for this?

beckermr:
I think a missing implementation is the main issue.

mgorny:
Thanks. I'm working on it now, though I guess it may take a round or two to get an optimal solution (read: do not reinvent too many wheels).

PR #2218, opened 2025-01-14
Support outputs in `hint_noarch_python_use_python_min`, also fix if/then/else flattening for non-list clauses

mgorny:
Checklist
* [x] Added a ``news`` entry

Add support for checking `python_min` use in multi-output recipes (#2188). This involves some minimal refactoring to move shared code into common functions.

While at it, accidentally discovered a bug that the following dependency:

```yaml
- if: foo
  then: bar
  else: baz
```

would be incorrectly flattened (`bar` would be iterated over as a list).

### DIFF
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -238,7 +238,7 @@ def hint_pip_no_build_backend(host_or_build_section, package_name, hints):
             )
 
 
-def hint_noarch_python_use_python_min_inner(
+def _hint_noarch_python_use_python_min_inner(
     host_reqs,
     run_reqs,
     test_reqs,
@@ -321,11 +321,11 @@ def hint_noarch_python_use_python_min(
                 output_host_reqs = requirements.get("host")
                 output_run_reqs = requirements.get("run")
             else:
-                output_host_reqs = requirements
+                output_host_reqs = None
                 output_run_reqs = requirements
 
             hint.extend(
-                hint_noarch_python_use_python_min_inner(
+                _hint_noarch_python_use_python_min_inner(
                     output_host_reqs or [],
                     output_run_reqs or [],
                     get_all_test_requirements(output, [], recipe_version),
@@ -338,7 +338,7 @@ def hint_noarch_python_use_python_min(
             )
     else:
         hint.extend(
-            hint_noarch_python_use_python_min_inner(
+            _hint_noarch_python_use_python_min_inner(
                 host_reqs,
                 run_reqs,
                 test_reqs,
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -3,6 +3,7 @@ import re
 import shutil
 import subprocess
 import sys
+from collections.abc import Mapping
 from glob import glob
 from typing import Any
 
@@ -12,6 +13,7 @@ from conda_smithy.linter.utils import (
     VALID_PYTHON_BUILD_BACKENDS,
     find_local_config_file,
     flatten_v1_if_else,
+    get_all_test_requirements,
     is_selector_line,
 )
 from conda_smithy.utils import get_yaml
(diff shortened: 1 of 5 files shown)


conda_smithy/linter/hints.py
```diff
@@ -291,24 +292,67 @@ def hint_noarch_python_use_python_min(
...
+        for output in outputs_section:
+            requirements = output.get("requirements", {})
+            if isinstance(requirements, Mapping):
+                output_host_reqs = requirements.get("host")
+                output_run_reqs = requirements.get("run")
+            else:
+                output_host_reqs = requirements
+                output_run_reqs = requirements
+
+            hint.update(
```
beckermr, line 323: update or add?

conda_smithy/linter/hints.py
```diff
@@ -291,24 +292,67 @@ def hint_noarch_python_use_python_min(
...
+        )
+
+    if hint:
+        hint = (
+            (
+                "`noarch: python` recipes should usually follow the syntax in "
+                "our [documentation](https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-python) "
+                "for specifying the Python version."
+            )
+            + "".join(sorted(hint))
```
beckermr, line 350: This sorts the entire set of hints based on the full hint. Don't we only want to use the name of the output?

conda_smithy/linter/hints.py
mgorny: Update. The `inner` function constructs a set of hints applying to a single output, then here we merge them into all hints applicable to a recipe. Not the most pretty solution, I admit, but I figured we don't want to repeat the same hint multiple times when it applies to multiple outputs.

conda_smithy/linter/hints.py
mgorny: We don't include the output name right now. We just give a list of "generic" hints that apply to at least one output in the recipe. Do you prefer per-output hints?

conda_smithy/linter/hints.py
beckermr: We do need to make it clear that a given hint applies to a given output. IIRC, In other parts of the linter we tolerate the repeats and report everything for each output. Alternatively we could aggregate the output names that each hint applies to and explicitly list them in each hint. I don't think the code as is will be clear enough.

mgorny: Okay, I'll update it tomorrow.

conda_smithy/linter/hints.py
beckermr: Yes we should do them per output. People will fix one and miss one and then be scratching their heads on why the linter is mad at them. We need to be specific.

conda_smithy/linter/hints.py
mgorny: I've added the output name to the message, when it applies to a specific output. Please let me know if this form is fine or do you prefer putting it there some other way.

conda_smithy/linter/hints.py
```diff
@@ -291,24 +293,74 @@ def hint_noarch_python_use_python_min(
...
+    hint = []
+
+    if outputs_section:
+        for output_num, output in enumerate(outputs_section):
+            requirements = output.get("requirements", {})
+            if isinstance(requirements, Mapping):
+                output_host_reqs = requirements.get("host")
+                output_run_reqs = requirements.get("run")
+            else:
+                output_host_reqs = requirements
```
beckermr, line 324:
```suggestion
                output_host_reqs = None
```

beckermr: Requirements that are not in a mapping are always run for outputs of v0 recipes.

conda_smithy/linter/utils.py
```diff
@@ -237,14 +237,44 @@ def load_linter_toml_metdata_internal(time_salt):
...
+                "run"
+            ) or []
+
+            if (
+                "python" in test_element
+                and test_element["python"].get("python_version") is not None
+            ):
+                test_reqs.append(
+                    f"python {test_element['python']['python_version']}"
+                )
```
beckermr, line 276: We should always at least add python here.

```suggestion
            if (
                "python" in test_element
            ):
                if test_element["python"].get("python_version") is not None:
                    test_reqs.append(
                        f"python {test_element['python']['python_version']}"
                    )
                else:
                    test_reqs.append("python")
```

beckermr:
pre-commit.ci autofix

beckermr (approved):

mgorny:
Thanks. Just wanted to say it's always a pleasure working with you.

PR #2228, opened 2025-01-22
fix up indentation error causing reduced test coverage

h-vetinari:
Oversight from #2227

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -560,7 +560,8 @@ def _collapse_subpackage_variants(
     # either v0 or v1 recipe must exist; no fall-back if missing
     with open(recipe_path) as f:
         meta_text = f.read()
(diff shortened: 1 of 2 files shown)


h-vetinari:
For reasons I don't really understand, 27bab567bdc0caab9a9fe250398806b2c4fe884a caused a regression here. If I copy the text and do it manually, then
```
>>> bool(re.match(r"\{\{ python_min \}\}", meta_text))
False
>>> bool(re.match(r".*\{\{ python_min \}\}", meta_text))
False
>>> bool(re.match(r".*\{\{ python_min \}\}.*", meta_text))
False
>>> bool(re.match(r"(?m).*\{\{ python_min \}\}.*", meta_text))
False
```
which makes no sense to me. If I go back to
```
>>> pm_pat = re.compile(r".*\{\{ python_min \}\}")
>>> any(pm_pat.match(x) for x in meta_text.splitlines())
True
```
it works.

PR #2244, opened 2025-02-17
consistently set `encoding=utf-8` for opening files

h-vetinari:
I haven't had issues with running the linter on windows, but today I ran into encoding errors while linting a multi-output rattler recipe (no idea what triggered it exactly). Since python won't default to utf-8 until 3.15(!), it's better to be explicit about the encoding where we're interacting with files. I only needed the changes in `lints.py`, but while I was at it, I also looked for other uses of bare `open()` (with the exception of some obviously unix-only files, e.g. those using paths like `~/...`)

### DIFF
--- a/conda_smithy/ci_skeleton.py
+++ b/conda_smithy/ci_skeleton.py
@@ -21,7 +21,7 @@ def _render_template(template_file, env, forge_dir, config):
     print("Generating " + target_fname, file=sys.stderr)
     new_file_contents = template.render(**config)
     os.makedirs(os.path.dirname(target_fname), exist_ok=True)
-    with open(target_fname, "w") as fh:
+    with open(target_fname, "w", encoding="utf-8") as fh:
         fh.write(new_file_contents)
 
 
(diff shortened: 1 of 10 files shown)


hmaarrfk (commented):
I use this in my codebase as well. The only thing is that some LICENSE.txt from packages are going to be non-utf8 encoded. do these changes touch any of those codepaths?

h-vetinari:
I don't see a reason why license files cannot be encoded in utf-8 themselves, it's not like the encoding is part of the license. Also, I don't think we do any parsing (or even `open()`-ing`) of license files, so the question shouldn't arise AFAICT.

h-vetinari:
BTW, I'm now hitting this in rerendering too, which makes this a fair bit more urgent for me...

hmaarrfk:
> I don't see a reason why license files cannot be encoded in utf-8 

I'm just saying I saw feedstocks' source code, from the package they point to, have a license with characters that are illegal for UTF-8

wolfv:
We could maybe also use `export PYTHONIOENCODING=utf8`?

Or perhaps:

```
import sys
sys.setdefaultencoding('utf-8')
```

PR #2245, opened 2025-02-18
fix superfluous encoding kwarg for reading in binary mode

h-vetinari:
Fix an oversight from applying changes mechanically in #2244; files opened in binary mode don't take an encoding, and this causes
```
    with open(dest, "wb", encoding="utf-8") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: binary mode doesn't take an encoding argument
```

Surprising that this wasn't caught by our tests... 🤔

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2620,7 +2620,7 @@ def get_cfp_file_path(temporary_directory):
 
     response = requests.get(pkg.url)
     response.raise_for_status()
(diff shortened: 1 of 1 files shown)


h-vetinari:
> wonder if one of the linters inside ruff would have caught it?

I don't think they're smart enough to lint one kwarg based on the _value_ of another. It's a bit of a complicated contract...

h-vetinari:
Thanks for the quick review!

PR #2248, opened 2025-02-25
Fix false positive when v1 recipe name is non-context variable

### LINKED ISSUE #2224: False positive in `lint_recipe_name`

### Issue

In `lint_recipe_name`, you can get a false positive with the following recipe:

```yml
# recipe.yaml
context:
  version: '1.20.0'

# Note: This recipe is specifically designed to work well with the autotick bot.
# Also refer to https://github.com/conda-forge/rust-feedstock/blob/main/recipe/meta.yaml.
package:
  name: ${{ polars_variant }}
  version: ${{ version }}

# ...
```

You need to specify `polars_variant` in `conda_build_config.yaml`:

```yml
polars_variant:
  - polars  # [not ppc64le]
  - polars-lts-cpu
  - polars-u64-idx  # [not ppc64le]
```

this occurred in https://github.com/conda-forge/polars-feedstock/pull/303

https://github.com/conda-forge/conda-smithy/blob/fcbd049cb268eb5c8732f4b5e498bf011398f5ea/conda_smithy/linter/conda_recipe_v1_linter.py#L132-L140

### Installed packages

```shell
-
```

### Environment info

```shell
-
```

mgorny:
Checklist
* [x] Added a ``news`` entry

Fixes #2224

Skip checking v1 recipe name if it contains unresolved variables, in order to fix false positives when a variable
from `conda_build_config.yaml` is used.

Iterating over all possible variants and verifying the resulting package names is unlikely to be worth the extra complexity.  This also roughly matches the linter behavior for v0 recipes, where `{{ var_name }}` is converted into `var_name`, and therefore is never properly linted anyway.

### DIFF
--- a/conda_smithy/linter/conda_recipe_v1_linter.py
+++ b/conda_smithy/linter/conda_recipe_v1_linter.py
@@ -134,6 +134,11 @@ def lint_recipe_name(
     lints: list[str],
 ) -> None:
     name = get_recipe_name(recipe_content)
+    # Avoid false positives if the recipe is using variables
+    # from conda_build_config.yaml.
+    # https://github.com/conda-forge/conda-smithy/issues/2224
+    if "${{" in name:
+        return
 
     lint_msg = _lint_recipe_name(name)
     if lint_msg:
--- a/news/2248-v1-recipe-name-var.rst
+++ b/news/2248-v1-recipe-name-var.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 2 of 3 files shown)


beckermr (changes requested):
Why can't we load the variables? For v1 this should be very fast. For v0 I agree this would be too slow.

mgorny:
> Why can't we load the variables? For v1 this should be very fast. For v0 I agree this would be too slow.

By "complexity", I meant the coding effort to make that happen.

Are you suggesting we re-lint the recipe for all variants, or just special-case the package name? I suppose the former could make sense, though we'd probably need to filter out duplicates somehow, and that probably isn't going to be easy.

beckermr:
Package name but ok. I'm not going to stand in the way here.

beckermr (approved):

mgorny:
Thanks. I mean, if you insist I can look into doing it, but it feels a bit, well, inefficient, to call `rattler-build` to render all the variants from within an individual check.

PR #2249, opened 2025-02-25
Use explicit path in "recipe not found" exception

mgorny:
Checklist
* [x] Added a ``news`` entry

Replace the "Feedstock has no ..." error message with a more clear "recipe not found" with explicit path, and a suggestion to pass the `recipe` subdirectory.  This should help people like me, who keep forgetting that in top feedstock directory you need to run:

    conda smithy lint recipe/

instead of just:

    conda smithy lint

and get the message:

    OSError: Feedstock has no recipe/meta.yaml

which is confusing since clearly there is a `recipe/meta.yaml` in the current directory.  Unless you're using v1 recipes, then you get even more confused why it tries to use a v0 recipe file.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -743,7 +743,7 @@ def main(
 
     if not os.path.exists(recipe_file):
         raise OSError(
-            f"Feedstock has no recipe/{os.path.basename(recipe_file)}"
+            f'Recipe not found: {recipe_file}; did you pass the "recipe" subdirectory?'
         )
 
     if build_tool == CONDA_BUILD_TOOL:
--- a/news/2249-recipe-not-found-path.rst
+++ b/news/2249-recipe-not-found-path.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Improve the "reciep not found" exception to include the path used and a suggestion what path to pass. (#2249)
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/2249-recipe-not-found-path.rst
+++ b/news/2249-recipe-not-found-path.rst
@@ -4,7 +4,7 @@
 
 **Changed:**
 
-* Improve the "reciep not found" exception to include the path used and a suggestion what path to pass. (#2249)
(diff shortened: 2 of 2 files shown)


conda_smithy/lint_recipe.py
```diff
@@ -743,7 +743,7 @@ def main(
 
     if not os.path.exists(recipe_file):
         raise OSError(
-            f"Feedstock has no recipe/{os.path.basename(recipe_file)}"
+            f'Recipe not found: {recipe_file}; did you pass the "recipe" subdirectory?'
```
jaimergp, line 746: I think there's a [setting for `recipe`](https://github.com/conda-forge/conda-smithy/blob/b447e7473fe172a90dde003a74af760a2a0d886a/conda_smithy/schema.py#L964-L975), so this shouldn't be hardcoded.

news/2249-recipe-not-found-path.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Improve the "reciep not found" exception to include the path used and a suggestion what path to pass. (#2249)
```
jaimergp, line 7:
```suggestion
* Improve the "recipe not found" exception to include the path used and a suggestion what path to pass. (#2249)
```

jaimergp:
Thanks! Added a comment, but I wonder if we could a good default so that CLI path is indeed not necessary though.

conda_smithy/lint_recipe.py
mgorny: Well, "kinda". If you aren't passing the correct directory, then we can't get the setting — and I dare say changing the name is rare enough that "recipe" will be a good hint most of the time. Perhaps it would work if I removed the quotes:

```suggestion
            f'Recipe not found: {recipe_file}; did you pass the recipe subdirectory?'
```

?

mgorny:
> I wonder if we could a good default so that CLI path is indeed not necessary though.

Make it work both with top feedstock directory and the subdirectory? I suppose that would make sense, and I can't see an obvious case where it would backfire.

mgorny:
So, how about this?

1. Look for `meta.yaml` + `recipe.yaml` in the specified directory per the current logic.
2. Otherwise, look for `conda-forge.yml` in the specified directory, and use its recipe path.
3. Otherwise, look for `recipe` subdirectory with one of the two files.

mgorny:
The alternative approach was implemented in #2250.

beckermr:
We merged the other PR, so I am closing this one. Thanks @mgorny!

PR #2270, opened 2025-03-05
feat: Add a lint on obsolete `os_version`

### LINKED ISSUE #2251: Hint on obsolete `os_version`
We [have](https://github.com/search?q=org%3Aconda-forge+%2Fos_version%3A%2F+path%3Aconda-forge.yml&type=code) >150 recipes that do a variation of

```yaml
os_version:
  linux_64: cos7
```

The vast majority of them are left-overs from a time when cos7 was needed to get newer runtime glibc (long obsolete since we switch the default image to cos7), but after https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6626, it actually has the opposite effect - keeping recipes on outdated images, and very intransparently at that (because the setting is tucked away in a corner where few people look).

IMO we should lint (or at least hint) on this along the lines of:
> This recipe is overriding the default image version from alma9 to cos7 (see `conda-forge.yml` in the feedstock root directory).  Unless you are in the very rare case of repackaging binary artefacts, you should consider removing this override.

Xref #2155

mgorny:
Checklist
* [ ] Added a ``news`` entry

Fixes #2251

Originally, I wanted to add this to `lintify_forge_yaml()` but that function uses `jsonschema.ValidationError`s as output.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -50,6 +50,7 @@ from conda_smithy.linter.lints import (
     lint_noarch,
     lint_noarch_and_runtime_dependencies,
     lint_non_noarch_builds,
+    lint_os_version,
     lint_package_version,
     lint_pin_subpackages,
     lint_recipe_have_tests,
@@ -372,6 +373,9 @@ def lintify_meta_yaml(
     lint_go_licenses_are_bundled(
         build_requirements, lints, recipe_version=recipe_version
     )
+
+    # 30. Check for obsolete os_version
+    lint_os_version(forge_yaml, lints)
 
     # hints
     # 1: suggest pip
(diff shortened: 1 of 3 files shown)


mgorny:
Updated. Also added support for skipping it.

news/2270-obsolete-os_version.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* Added a lint for feedstocks using obsolete ``os_version``. (#2270)
```
beckermr, line 3:
```suggestion
* Added a hint for feedstocks using obsolete ``os_version``. (#2270)
```

beckermr (approved):

mgorny:
Thanks!

PR #2274, opened 2025-03-10
feat: add some line numbers to linting

nh13:
This is a draft PR that adds line numbers to various lints as an example how they can implemented.  If the implementation looks ok, I can add a few more comments in the code, and get it ready for review.

Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/linter/lints.py
+++ b/conda_smithy/linter/lints.py
@@ -32,7 +32,36 @@ from conda_smithy.utils import get_yaml
 logger = logging.getLogger(__name__)
 
 
+def _lineno(data, *key):
+    """Gets the liner number for a series of keys in meta.
+
+    Args:
+        data: the node in the meta data dictionary loaded with ruamel.yaml so that
+              line numbers are available.
+        key: one or more keys to traverse meta data with.
+
+    Returns:
+        the line number of the last key with a line number, otherwise '???'.
+    """
+    # Note: data.lc.line is zero-based
+    lineno = "???"
+    if not hasattr(data, "lc"):
+        return lineno
+    if len(data) == 0:
+        return "1"
+    if not all(isinstance(k, str) for k in key):
+        raise Exception()
+    for k in key:
+        data = data.get(k)
+        if data is None or not hasattr(data, "lc"):
+            lineno = f"~{lineno}"  # tilde for when line number is not accurate
+            break
+        lineno = f"{data.lc.line + 1}"
+    return lineno
+
+
 def lint_section_order(
+    meta: Any,
     major_sections: list[str],
     lints: list[str],
     recipe_version: int = 0,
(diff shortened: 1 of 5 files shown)


conda_smithy/linter/lints.py
```diff
@@ -32,7 +32,36 @@
 logger = logging.getLogger(__name__)
 
 
+def _lineno(data, *key):
+    """Gets the liner number for a series of keys in meta.
```
nh13, line 36:
```suggestion
    """Gets the line number for a series of keys in meta.
```

nh13:
pre-commit.ci autofix

nh13:
pre-commit.ci autofix

conda_smithy/linter/lints.py
```diff
@@ -110,16 +158,19 @@ def lint_recipe_have_tests(
...
                         has_outputs_test = True
                     elif test_out.get("script", "").endswith((".bat", ".sh")):
                         has_outputs_test = True
                     else:
+                        line_no = _lineno(meta["outputs"][index], "name")
                         no_test_hints.append(
                             "It looks like the '{}' output doesn't "
-                            "have any tests.".format(out.get("name", "???"))
+                            "have any tests on line {}.".format(
+                                out.get("name", "???"), line_no
```
beckermr, line 172:
```suggestion
                                out.get("name", f"<output {index+1}>"), line_no
```

beckermr (commented):
I'm concerned about the technical debt of us having to maintain our own custom ruamel.yaml classes. Is there another way to do this?

nh13:
@beckermr thats the best I could find unfortunately, and the API is meant for these cases. Alternatively I could create a new Python package to move the burden of maintenance there? I’d be happy to maintain.

beckermr:
Yep totally understood. I am wondering how much we could handle via searching the text and simple parsing of certain kinds of blocks by hand. It is hacky, but easier to maintain in the end. We are not being very precise on line numbers anyways given how v0 recipes are parsed and our need to default to `~???` if we cannot find anything.

nh13:
Would it be more of a maintenance burden to have a bespoke implementation to get line numbers in each lint function.  I imagine searching by text would eventually evolve to maintaining our own parser. 

I’d be happy to put up a new library on pypi that could be used?

beckermr:
The thing is that we (interpreted broadly as the conda ecosystem + conda-smithy) have had a lot of trouble over the years with using yaml parsers outside of `ruamel.yaml` and `pyyaml`. 

We used to depend on an ancient `ruamel.yaml` version vendored into a package called `ruamel_yaml`. That package was not maintained and unwinding that technical debt was not fun. 

I don't doubt your good intentions on a separate package. I just think it is actually a huge amount of maintenance and I really want to avoid it. 

From the looks of the current PR, and thinking through other things the linter does, all of the line number lookups boil down to 

1. Find the the data for last element nested within a sequence of maps or lists (i.e. `about.license`, or `outputs[0].name` in ~`jq`-like syntax)
2. Return the line number of this last thing.

This specific set of steps I think could be a single function that does the line number lookup by tracking the current indent level, where it encounters `-`, and looking for the right text that matches a given key.

nh13:
@beckermr I am sorry for the trouble that caused, I can only imagine how frustrating that must have been.  I'll consider taking another look at a simpler approach, though in my spare time I did manage to publish a separate package [`ruamel-yaml-line-info`](https://pypi.org/project/ruamel-yaml-line-info/) in case I or others ever needed it

PR #2283, opened 2025-03-21
add ubi8 to image versions being linted on

h-vetinari:
This is a minor fix-up of #2270. Since #2155, we're limiting the possible values of `os_version` through schema validation to 
https://github.com/conda-forge/conda-smithy/blob/22169a89bc72614918bdba42a6fbe7c7feec0e52/conda_smithy/schema.py#L49

but #2270 did not add a hint for that. The ubi8 images are still in use for CUDA 11.8, which is also the only place that needs them. They can continue to be used for that purpose (and any feedstock building for CUDA 11.8 will pull them in automatically), but users shouldn't set `os_version: [linux_64: ubi8]` in `conda-forge.yml` anymore.

CC @mgorny

### DIFF
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -466,7 +466,7 @@ def hint_os_version(
     hints: list[str],
 ) -> None:
     default_os_version = "alma9"
-    obsolete_os_versions = ("cos7", "alma8")
+    obsolete_os_versions = ("cos7", "alma8", "ubi8")
     matches = {
         k: v
         for k, v in forge_yaml.get("os_version", {}).items()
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
Needs a news item.

h-vetinari:
> Needs a news item.

This is such a minor fix, I don't think a news item is justified. It's more for completeness rather than anything else - I don't think any feedstock actually sets this value.

beckermr:
Alright I'll do it myself. Thanks.

beckermr (approved):
LGTM and thank you!

h-vetinari:
> Alright I'll do it myself. Thanks.

If it's important to you, I can add it (going forward I guess, because in the time it took me to write this on a phone, you already pushed 😅)

PR #2285, opened 2025-03-28
fix(schema): $ref and properties should not be combined

ytausch:
Checklist
* [x] Added a ``news`` entry

Cc @pavelzw


ytausch:
@jaimergp @beckermr Can you have a look at this?

PR #2289, opened 2025-03-31
regen schema with pydantic 2.11.1

minrk:
pydantic 2.11 adds `additionalProperties: true` to empty `type: object` fields


beckermr:
Should we add a constraint to the recipe too?

beckermr:
Woops missed it. Needs a news items.

Issue #2296, opened 2025-04-11
Re-rendering with multiple `channel_targets` generates inconsistent README instructions


jakirkham:
In the `python` feedstock, the `recipe/conda_build_config.yaml` file has the following `channel_targets` set

```yaml
channel_targets:
  - conda-forge main
  - conda-forge python_debug   # [not win]
```

When re-rendering with conda-smithy, it appears that one of these `channel_targets` is chosen (seemingly at random? though this could be subtle difference between re-rendering conditions)

It would be good pick one value from `channel_targets` (maybe the first one?). And/or have some way to pick which channel is used in the README instructions generated

matthiasdiener:
Here is the output of conda info for the re-render referenced above:

```console
$ conda info

     active environment : ceesd
    active env location : /Users/mdiener/Work/emirge/miniforge3/envs/ceesd
            shell level : 1
       user config file : /Users/mdiener/.condarc
 populated config files : /Users/mdiener/Work/emirge/miniforge3/.condarc
                          /Users/mdiener/.condarc
          conda version : 25.3.1
    conda-build version : not installed
         python version : 3.9.19.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=25.3.1=0
                          __osx=15.4=0
                          __unix=0=0
       base environment : /Users/mdiener/Work/emirge/miniforge3  (writable)
      conda av data dir : /Users/mdiener/Work/emirge/miniforge3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/osx-arm64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /Users/mdiener/Work/emirge/miniforge3/pkgs
                          /Users/mdiener/.conda/pkgs
       envs directories : /Users/mdiener/Work/emirge/miniforge3/envs
                          /Users/mdiener/.conda/envs
               platform : osx-arm64
             user-agent : conda/25.3.1 requests/2.32.3 CPython/3.9.19 Darwin/24.4.0 OSX/15.4 solver/libmamba conda-libmamba-solver/25.3.0 libmambapy/2.0.8
                UID:GID : 501:20
             netrc file : None
           offline mode : False
```

jakirkham:
It is possible that this was fixed in PR: https://github.com/conda-forge/conda-smithy/pull/2300

Played around with that PR and re-rendering the `python-feedstock`

[...]

PR #2300, opened 2025-04-18
fix: resolve non-stable rerenders by producing CI jobs for the intersection of loop vars over all outputs

### LINKED ISSUE #2100: BUG: smithy 3.43.0 creates too many jobs on ctng-compiler-activation
Since smithy 3.43.0, the rerender of https://github.com/conda-forge/ctng-compiler-activation-feedstock is incorrect - it creates one job per clang & GCC version, when there should only be a job per GCC version that builds all clangs. Here's is an example commit where this goes wrong: https://github.com/conda-forge/ctng-compiler-activation-feedstock/commit/5ca997aa0e832bedcc96aa173f719a741b9f5a2e

Going back to 3.42.1 fixes the problem.

beckermr:
Checklist
* [x] Added a ``news`` entry

This PR fixes an issue where rerenders got different numbers of CI jobs depending on which output appeared first in the list of metadata objects for each output (but only when there is not an output corresponding to the top-level package name.) The first element depends on the python hash seed, hence things not being stable.

Per the code comment, we resolve this ambiguity by producing CI jobs for all combinations of the intersection of top-level loop vars over all outputs.

Another possibility would be to let people add or exclude variables to be expanded over in the CI support jobs by hand. I'd rather not do that yet, so let's try this first.

I tested this version of smithy against the ctng-compiler-activation-feedstock locally and it rendered the expected set of CI jobs.

This PR also resolves something that I have wondered about for a long time, which is why we always used index 0 to get the loop vars.

closes #2100

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -567,10 +567,37 @@ def _collapse_subpackage_variants(
     # on osx, merge MACOSX_DEPLOYMENT_TARGET & c_stdlib_version to max of either; see #1884
     all_variants = _merge_deployment_target(all_variants, has_macdt)
 
-    top_level_loop_vars = list_of_metas[0].get_used_loop_vars(
-        force_top_level=True
-    )
-    top_level_vars = list_of_metas[0].get_used_vars(force_top_level=True)
+    # we grab the shortest list of top-level loop vars
+    # for some recipes, there can be outputs which do not depend on some
+    # of the variables used by other outputs. these outputs will get
+    # duplicate builds if we naively expand the entire build matrix into
+    # one .ci_support file per entry.
+    # for this reason, we want to force the ci_support files to only be
+    # produced for the smallest set of top-level loop vars over all outputs.
+    # this heuristic will fail if we have a recipe that builds two totally unrelated
+    # packages. in that case, we'd need to group the outputs into non-overlapping
+    # subsets (as determined by overlaps between their top-level loop vars), use the
+    # smallest set of top-level vars per group, and then produce a set of ci support jobs
+    # that is a union of the jobs from each group.
+    # we don't do that here since we should not be building unrelated packages in the
+    # same recipe anyways.
+    top_level_loop_vars = None
+    top_level_vars = None
+    for meta in list_of_metas:
+        curr_top_level_loop_vars = meta.get_used_loop_vars(
+            force_top_level=True
+        )
+        curr_top_level_vars = meta.get_used_vars(force_top_level=True)
+
+        if top_level_loop_vars is None:
+            top_level_loop_vars = curr_top_level_loop_vars
+            top_level_vars = curr_top_level_vars
+            continue
+
+        if len(curr_top_level_loop_vars) < len(top_level_loop_vars):
+            top_level_loop_vars = curr_top_level_loop_vars
+            top_level_vars = curr_top_level_vars
+
     if "target_platform" in all_used_vars:
         top_level_loop_vars.add("target_platform")
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -567,7 +567,8 @@ def _collapse_subpackage_variants(
     # on osx, merge MACOSX_DEPLOYMENT_TARGET & c_stdlib_version to max of either; see #1884
     all_variants = _merge_deployment_target(all_variants, has_macdt)
 
-    # we use the intersection of all of the loop vars in the outputs.
+    # we use the intersection of all of the loop vars in the outputs
+    # to distribute CI jobs.
     # this eliminates cases where a fully expanded build matrix would produce
     # duplicate outputs if an output used some variables but not others.
     # For outputs that do not use some variables, conda-build will properly
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1205,6 +1205,11 @@ def _render_ci_provider(
         for meta in metas:
             if not meta.skip():
                 enable_platform[i] = True
+
+        # sort the list of metas by the distribution string to ensure stable
+        # rendering later
+        metas = sorted(metas, key=lambda x: x.dist())
+
         metas_list_of_lists.append(metas)
 
     if not any(enable_platform):
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -567,39 +567,27 @@ def _collapse_subpackage_variants(
     # on osx, merge MACOSX_DEPLOYMENT_TARGET & c_stdlib_version to max of either; see #1884
     all_variants = _merge_deployment_target(all_variants, has_macdt)
 
-    # we grab the shortest list of top-level loop vars
-    # for some recipes, there can be outputs which do not depend on some
-    # of the variables used by other outputs. these outputs will get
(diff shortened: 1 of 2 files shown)


beckermr:
This one is ready for review @conda-forge/core!

beckermr:
@jaimergp @xhochy @h-vetinari can you all send me an example of a feedstock that produces the same noarch package on more than one platform with it depending on non-noarch packages?

I want to see what happens to the CI jobs with this pr. 

Is one of the arrow feedstocks a good place to try?

conda_smithy/configure_feedstock.py
```diff
@@ -567,10 +567,37 @@ def _collapse_subpackage_variants(
...
+    # this heuristic will fail if we have a recipe that builds two totally unrelated
+    # packages. in that case, we'd need to group the outputs into non-overlapping
+    # subsets (as determined by overlaps between their top-level loop vars), use the
+    # smallest set of top-level vars per group, and then produce a set of ci support jobs
+    # that is a union of the jobs from each group.
+    # we don't do that here since we should not be building unrelated packages in the
+    # same recipe anyways.
+    top_level_loop_vars = None
+    top_level_vars = None
+    for meta in list_of_metas:
```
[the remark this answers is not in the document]
beckermr: This would resolve degeneracies between the same number of min vars. Yes we should!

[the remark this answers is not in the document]
beckermr: done!

beckermr:
And the sorting of the metas yield more bugs since apparently there is more than one set of loop vars that have the same number of elements. :/

I'll need to think more on how to break that degeneracy.

beckermr:
OK. I checked all of the feedstocks from @isuruf and I checked the feedstock from the original issue. They all render as expected with this change. 

Ready for another look @conda-forge/core!

news/2300-min-ci-support-job.rst
```diff
@@ -0,0 +1,24 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed unstable rerenders in an edge case by producing the smallest number of CI support jobs. (#2300)
```
beckermr, line 19:
```suggestion
* Fixed unstable rerenders in an edge case by using the intersection of top-leve loop vars over all outputs. (#2300)
```

h-vetinari:
Thank you so much for tackling this @beckermr! The timing is great, because I've been struggling for 2 days with rerendering some fixes for the ctng activation feedstock. I've now opened a [PR](https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/152) based on your fixes here - the bad news is that it still does not collapse the `clang_version` dimension into the CI jobs.

> Is one of the arrow feedstocks a good place to try?

This [one](https://github.com/conda-forge/pyarrow-feedstock/pull/119) is certainly worth a try, though it's currently not buildable due to bugs in conda-build. It's also more complicated than the other cases because of some other yesarch outputs withing the same job depending on the noarch one.

conda_smithy/configure_feedstock.py
```diff
@@ -567,13 +567,28 @@ def _collapse_subpackage_variants(
     # on osx, merge MACOSX_DEPLOYMENT_TARGET & c_stdlib_version to max of either; see #1884
     all_variants = _merge_deployment_target(all_variants, has_macdt)
 
-    top_level_loop_vars = list_of_metas[0].get_used_loop_vars(
-        force_top_level=True
-    )
-    top_level_vars = list_of_metas[0].get_used_vars(force_top_level=True)
+    # we use the intersection of all of the loop vars in the outputs.
```
beckermr, line 570:
```suggestion
    # we use the intersection of all of the loop vars in the outputs
    # to distribute CI jobs.
```

beckermr:
Well that stinks. I am looking at this now.

beckermr:
@h-vetinari you rendered with the wrong commit I think. Even if you installed smithy, you need to use `conda-smithy render` with the dash to ensure you get smithy from your local env and not the global one potentially. Also my dev version locally is `conda-smithy-3.47.2.dev11+g6b6c9255` whereas yours in the commit you made is `3.47.1.dev7+g9f7066fb`.

I just pushed a rerender to your branch from the correct commit and it looks fine.

beckermr:
When you are rendering locally, it will print a new info log about which variables it is using like this

```
NFO:conda_smithy.configure_feedstock:variables used to distribute packages over CI jobs:
    {'target_platform', 'cross_target_platform', 'gcc_version'}
INFO:conda_smithy.configure_feedstock:variables used to distribute packages over CI jobs:
    {'target_platform', 'cross_target_platform', 'gcc_version'}
INFO:conda_smithy.configure_feedstock:variables used to distribute packages over CI jobs:
    {'target_platform', 'cross_target_platform', 'gcc_version'}
INFO:conda_smithy.configure_feedstock:variables used to distribute packages over CI jobs:
    {'target_platform', 'cross_target_platform', 'gcc_version', 'clang_version'}
INFO:conda_smithy.configure_feedstock:variables used to distribute packages over CI jobs:
    {'target_platform', 'cross_target_platform', 'gcc_version'}
```

If you do not see those lines, you are using the wrong version.

h-vetinari:
> Also my dev version locally is `conda-smithy-3.47.2.dev11+g6b6c9255` whereas yours in the commit you made is `3.47.1.dev7+g9f7066fb`.

Yeah, sorry, I had had a typo when doing `gi tcheckout render-latest` and so my local install did not match this branch.

h-vetinari (approved):
Just confirmed that the local rerender now works for me as well. Thank a bunch @beckermr! 🙏

jakirkham:
Would this help with issue ( https://github.com/conda-forge/conda-smithy/issues/2296 )?

beckermr:
I doubt it. Conda build detects that channel targets is not used for the build itself and so that variable is likely already removed or ignored in the sections of code worked on.

jakirkham:
We added to `always_keep_keys`

https://github.com/conda-forge/conda-smithy/blob/717ba104aa163a6ddc0e46e091ead512cf804321/conda_smithy/configure_feedstock.py#L627-L628

beckermr:
Ahhh yes I am wrong. Worth a try for sure.

conda_smithy/configure_feedstock.py
```diff
@@ -567,13 +567,29 @@ def _collapse_subpackage_variants(
...
+    top_level_vars = set.intersection(*top_level_vars)
+    top_level_loop_vars = set.intersection(*top_level_loop_vars)
+
     if "target_platform" in all_used_vars:
         top_level_loop_vars.add("target_platform")
 
+    logger.info(
+        "variables used to distribute packages over CI jobs:\n%s",
+        textwrap.indent(pprint.pformat(top_level_loop_vars), "    "),
+    )
```
jakirkham, line 591: FWIW do see the following when testing with the `python-feedstock`

```python
INFO:conda_smithy.configure_feedstock:variables used to distribute packages over CI jobs:
    {'build_type', 'channel_targets', 'target_platform'}
```

Note the presence of `channel_targets`

beckermr: Right I am just not sure if this means the README will be stable, which was the question.

jakirkham: Posted more info about testing for that issue in comment: https://github.com/conda-forge/conda-smithy/issues/2296#issuecomment-2816417890

PR #2304, opened 2025-04-22
Allow skipping lint_stdlib

xhochy:
Checklist
* [x] Added a ``news`` entry


xhochy:
We need to use `MACOSX_DEPLOYMENT_TARGET` in https://github.com/conda-forge/rust-activation-feedstock and thus this would be useful to please the linter.

PR #2308, opened 2025-04-28
Only show stdlib lint for go-cgo compilers, not go-nocgo

xhochy:
Checklist
* [x] Added a ``news`` entry


beckermr (changes requested):
I don't understand how these changes prevent the linter from showing the hint on one version of go but not the other. Neither were in the regex previously, so how does adding one prevent the other from being shown?

Also, this PR needs a test.

PR #2309, opened 2025-04-28
Add support for rattler-build-config

pavelzw:
Checklist
* [ ] Added a ``news`` entry

### DIFF
--- a/conda_smithy/templates/build_steps.sh.tmpl
+++ b/conda_smithy/templates/build_steps.sh.tmpl
@@ -150,6 +150,9 @@ else
 
     {%- else %}
 
+    if [[ -f "${RECIPE_ROOT}/rattler-build-config.toml" ]]; then
+        EXTRA_CB_OPTIONS="${EXTRA_CB_OPTIONS:-} --config-file ${RECIPE_ROOT}/rattler-build-config.toml"
+    fi
     {{ BUILD_CMD }} --recipe "${RECIPE_ROOT}" \
      -m "${CI_SUPPORT}/${CONFIG}.yaml" \
      ${EXTRA_CB_OPTIONS:-} \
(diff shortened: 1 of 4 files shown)


beckermr (changes requested):
Why are we forcing users to use a new thing in order to specify channels etc? 

Rattler should be reading the variant config keys. 

I don't like this change at all.

beckermr:
Also this configuration will not work with the ci-setup package and how it handles channels.

pavelzw:
my main idea was to reuse pixi-config for mirror + s3 configuration. channels were a natural addition since they are also in the pixi config crate.

but i can also adjust the behavior and read them from the variant config instead

wolfv:
I think this might be orthogonal.

Yes, `rattler-build` should read the channels from the variant config. 

But just like `conda-build`, the default channels should be read from the "configuration file" (which is `pixi.toml` or `rattler-build.toml` file on the system). `conda-build` reads the config from the `.condarc` file.

While I think it's OK to read the configuration for channels from the variant config, I am not sure in how far the "variant config" is the "only" config for conda build tools.
There are certainly things outside the variant config (e.g. how many parallel processes should be used, compression level, ...) that fall outside the "variant config".

beckermr:
Ah yes. That makes sense. We need to redirect this pr to the ci-setup package. This is where we handle conda build options.

beckermr:
@wolfv @pavelzw We set items in the condarc on the fly via the ci-setup package here: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/main/recipe/conda_forge_ci_setup/build_utils.py. 

Items go into the ci-setup package when they can be done without rerenders. I am guessing for the rattler build settings this is true. So we should have a similar mechanism for rattler build that reads settings out of the conda-forge.yml and applies them properly during the build.

PR #2315, opened 2025-05-03
fix: run post-link scripts on Windows

wolfv:
We ran into issues because `m2-git` requires post-link scripts to be run.

This should fix: https://github.com/conda-forge/conda-smithy/issues/2239

### DIFF
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -70,6 +70,9 @@ move /y pixi.toml pixi.toml.bak
 set "arch=64"
 if "%PROCESSOR_ARCHITECTURE%"=="ARM64" set "arch=arm64"
 powershell -NoProfile -ExecutionPolicy unrestricted -Command "(Get-Content pixi.toml.bak -Encoding UTF8) -replace 'platforms = .*', 'platforms = [''win-%arch%'']' | Out-File pixi.toml -Encoding UTF8"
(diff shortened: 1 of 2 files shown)


news/fix-rattler-build-test-strict-parsing.rst
```diff
@@ -0,0 +1,3 @@
+**Fixed:**
+
+* Fix a test that uses `match` in skip to work with strict parsing (#2213)
```
h-vetinari: This looks like a left-over or C&P error

h-vetinari (commented):
Could you add a test that uses `git_url:` (or however it's called now) in a v1 recipe?

news/fix-rattler-build-test-strict-parsing.rst
wolfv: Yep just force pushed, sry!

h-vetinari: Ah, this just disappeared with the force-push.

wolfv:
Do we have this kind of "end-to-end" testing? It requires a proper feedstock to run. I could change this line in an existing feedstock / PR that was failing though.

wolfv:
This feedstock already had the change applied: https://github.com/conda-forge/katago-feedstock/pull/14#issuecomment-2772971915

And it "works" (triggers a different error).

beckermr (approved):

PR #2318, opened 2025-05-11
don't hint on v1 python_version list with python_min

bollwyvl:
Checklist
* [x] Added a ``news`` entry

Thanks for the tireless effort on maintaining `conda-smithy`!

While CFEP-25 specifies packages SHOULD be validated against the ecosystem (or feedstock) `python_min`, validating against _some other python_ can find _interesting_ upstream issues.

This PR updates `hint_noarch_python_use_python_min` to allow [v1 recipes](https://github.com/prefix-dev/recipe-format/blob/fa2df/schema.json#L3320-L3337) to specify testing against `python_min` **and** some other python, at the maintainer cost of a single line, and a single, generally very fast, solve.

### DIFF
--- a/conda_smithy/linter/utils.py
+++ b/conda_smithy/linter/utils.py
@@ -269,10 +269,13 @@ def get_all_test_requirements(
             ) or []
 
             if "python" in test_element:
-                if test_element["python"].get("python_version") is not None:
-                    test_reqs.append(
-                        f"python {test_element['python']['python_version']}"
-                    )
+                py_version = test_element["python"].get("python_version")
+
+                if isinstance(py_version, str):
+                    py_version = [py_version]
+
+                if isinstance(py_version, list):
+                    test_reqs += [f"python {pv}" for pv in py_version]
                 else:
                     test_reqs.append("python")
     else:
--- a/news/v1-python-version-matrix.rst
+++ b/news/v1-python-version-matrix.rst
@@ -1,0 +1,27 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* A v1 recipe ``tests`` item with a matrix-style ``python_version`` will not be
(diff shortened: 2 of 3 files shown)


bollwyvl:
pre-commit.ci autofix

bollwyvl:
Boo:

> The recipe is not parsable by parser `conda-recipe-manager`. The recipe can only be automatically migrated to the new v1 format if it is parseable by conda-recipe-manager.

beckermr:
OK @bollwyvl - this needs a news item but is otherwise good to go!

bollwyvl:
> needs a news item

Sure, will do.

But perhaps the `can only be automatically migrated to the new v1 format` suggests `conda-recipe-manager` lint is irrelevant for recipes that are _already_ v1? I'd need to dig into what other value it might be presently providing...

beckermr:
Conda-recipe-manager is definitely not irrelevant for v1 recipes. Eventually I'd like to use it to parse and modify recipes within the bot.

beckermr (approved):

bollwyvl:
I'll keep looking over there, then... I know i've seen the issue in logs (it's trying to `jinja.Template(['${{ python_min }}.*', '...'])`) but i recall it being a little cryptic where the error chain started.

bollwyvl:
Thanks! Yeah, that `parent = node_stack[-1]` is... pretty opaque, and having some issues even reproducing it over there...

beckermr:
Yeah so CRM is a combined custom jinja2+yaml parser aimed at v0 recipes. In practice, for v1 recipes you only need a regular off-the-shelf yaml parser plus any off-the-shelf jinja2 templating engine. The goal with conda-recipe-manager is to have one tool that handles both. The opaque error is because you are seeing an error in the parser itself (which is a somewhat complicated algorithm).

PR #2321, opened 2025-05-19
enable dev mode on azure-pipelines-win to support symlink

trim21:
Checklist
* [ ] Added a ``news`` entry

more details: https://github.com/conda-forge/mediainfo-feedstock/pull/5

### DIFF
--- a/conda_smithy/templates/azure-pipelines-win.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-win.yml.tmpl
@@ -6,6 +6,8 @@ jobs:
 - job: win
   {{ azure_yaml|indent(2) }}
   steps:
+    # enabled dev mode to allow symlink
+    - reg add "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock" /t REG_DWORD /f /v "AllowDevelopmentWithoutDevLicense" /d "1"
 {%- if clone_depth is not none %}
     - checkout: self
       fetchDepth: {{ clone_depth }}


wolfv (approved):

wolfv:
@jaimergp do you know if our content checks catch symlinks in packages on Windows? I think that should be disallowed for maximum compatibility.

h-vetinari:
> more details: https://github.com/conda-forge/mediainfo-feedstock/pull/5

A reference is not a suitable argument - please don't make reviewers chase down rabbitholes; at least provide a short summary.

In general I would be against enabling symlinks simply because that makes it much more likely that we accidentally publish a package with symlinks, which would be broken on a large portion of windows clients. But given that the checkout in https://github.com/conda-forge/mediainfo-feedstock/pull/5 fails because the  _sources_ contain a symlink, I can at least see the argument.

> [do] content checks catch symlinks in packages on Windows? I think that should be disallowed for maximum compatibility.

This would of course be a solution. Assuming that they _don't_ already check for symlinks, what I'd prefer to avoid is that this gets pushed off as unrelated here, and then never happens.

h-vetinari:
the title is the what, not the why

trim21:
If we postpone this, is there an alternative solution I can put in place?

trim21:
A workaround would be checkout git repo in build scripts.....

isuruf (changes requested):
Adding this might prevent checks about symlinks in conda packages right?

trim21:
> Adding this might prevent checks about symlinks in conda packages right?

Looks like we currently doesn't have symlinks checks for windows in conda-forge.

PR #2326, opened 2025-05-31
Fix macos self hosted labels

wolfv:
cc @aktech


isuruf (changes requested):
This is correct AFAIK. cirun might be overriding defaults.

See for eg: https://github.com/guyrt/arrow/blob/master/dev/tasks/python-wheels/github.osx.arm64.yml#L44

aktech:
>@aktech there is also this code that always adds "self-hosted" as a label. Should we keep it?

Nope, that isn't required.

Issue #2332, opened 2025-06-13
Increase black linelength


h-vetinari:
The following is way too short and leads to badly formatted lines, especially with some of our longer variable names
https://github.com/conda-forge/conda-smithy/blob/41d734946cbe125f32ab93d1f324001a12a9f0c7/pyproject.toml#L34-L35

It is not appropriate in an age where 80-character terminals are decades out of date, especially as this is even tighter than black's default of 88 characters. We should _at least_ match that, or go to 100 characters directly.

CC @conda-forge/core

PR #2333, opened 2025-06-13
Use black's native linelength (Option A)

h-vetinari:
One alternative for fixing #2332, using the default linelength by black.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -84,9 +84,7 @@ SERVICE_FEEDSTOCKS = [
     "cf-autotick-bot-test-package-feedstock",
 ]
 if "CONDA_SMITHY_SERVICE_FEEDSTOCKS" in os.environ:
-    SERVICE_FEEDSTOCKS += os.environ["CONDA_SMITHY_SERVICE_FEEDSTOCKS"].split(
-        ","
-    )
+    SERVICE_FEEDSTOCKS += os.environ["CONDA_SMITHY_SERVICE_FEEDSTOCKS"].split(",")
 
 # Cache lifetime in seconds, default 15min
 CONDA_FORGE_PINNING_LIFETIME = int(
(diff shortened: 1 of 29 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -370,12 +366,7 @@ def _get_used_key_values_by_input_order(
...
-                tuple(
-                    [
-                        tup[used_keyset_ind]
-                        for used_keyset_ind in used_keyset_inds
-                    ]
-                )
+                tuple([tup[used_keyset_ind] for used_keyset_ind in used_keyset_inds])
                 for tup in tuples
             ]
         )
```
h-vetinari, line 372: A good example how the too-short line-length generates garbage sometimes.

h-vetinari (commented):
I understand that linelength is subjective, but looking through the diff here, this would be such a massive improvement from my POV 🤩

conda_smithy/configure_feedstock.py
```diff
@@ -1873,15 +1809,9 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
...
-        )
-        platform_templates["osx"].append(
-            ".scripts/create_conda_build_artifacts.sh"
-        )
-        platform_templates["win"].append(
-            ".scripts/create_conda_build_artifacts.bat"
-        )
+        platform_templates["linux"].append(".scripts/create_conda_build_artifacts.sh")
+        platform_templates["osx"].append(".scripts/create_conda_build_artifacts.sh")
+        platform_templates["win"].append(".scripts/create_conda_build_artifacts.bat")
```
h-vetinari, line 1814: I'd be especially glad to not split such trivial statements across multiple lines, which just makes it that much harder to digest.

xhochy (approved):
I'm in favour of this, just the default because it is the default. Otherwise, this will turn into some long bikeshedding.

beckermr:
pre-commit.ci autofix

beckermr (approved):
I second this PR as it is the default in black itself.

PR #2341, opened 2025-06-25
Detect plain cargo calls

xhochy:
In draft while I prepare the CFEP

Checklist
* [x] Added a ``news`` entry


jaimergp:
With #2408 we have revamped how linter messages are documented and emitted. Would you like to update this PR, @xhochy? I can also do that, but not sure if this is blocked by something else because it's been in draft since June.

Issue #1971, opened 2024-07-09
ENH: Unified interface for creating swap files?


h-vetinari:
From #1751, where I had noted at the time:

> If we're making a new switch for swap files, it would probably make sense to tie this in with https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155 / https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/158, which is currently activated by:
> ```yaml
> azure:
>   settings_win:
>     variables:
>       SET_PAGEFILE: "True"
> ```
>
> I had an unmerged [PR](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/160) that would have allowed setting the size too, if we want to match it in capabilities with linux.
>
> Not sure about where the PowerShell script would have to live, but IIUC, conda-smithy would just generate the call in `azure-pipelines-wind.yml.tmpl`, but could rely on the fact that the respective PowerShell script is present after `conda-forge-ci-setup` has been installed.

In contrast, the interface on linux looks like:
```yaml
azure:
  settings_linux:
    swapfile_size: 15GiB
```
A potential unification was descoped at the time to not hold up the PR, but it would be nice to have a uniform way to do this cross-platform.

_Originally posted by @h-vetinari in https://github.com/conda-forge/conda-smithy/issues/1751#issuecomment-1670548575_

jakirkham:
This seems like a good idea

Do we see them staying roughly the same across platforms? Or do we need some escape hatch for customizing per OS?

PR #2347, opened 2025-07-03
Fix artefact generation on GHA for windows

### LINKED ISSUE #2345: Fix artefact persistence on windows when using GHA
On the pytorch feedstock, we're having issues with a package upload due to a combination of large size, slow uploads of the azure machine, and some internal timeout somewhere in Anaconda's infrastructure.

We wanted to use artefact persistence as a work-around (so that if the upload in CI fails, we can download the artefact and push it manually), but that doesn't work yet on windows when using:
```yaml
provider:
  win_64: github_actions
```

@jaimergp provided the following analysis:
> It's erroring here
> https://github.com/conda-forge/pytorch-cpu-feedstock/blob/55fb9b87412b2ce8aacbe61622555ee9ff66d3eb/.scripts/create_conda_build_artifacts.sh#L28-L32
> See at
> https://github.com/conda-forge/pytorch-cpu-feedstock/blob/55fb9b87412b2ce8aacbe61622555ee9ff66d3eb/.github/workflows/conda-build.yml#L166
> But apparently conda-build is writing to `C:\bld`
> Probably because of this
> https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/880c86261316b03c374efce7fc4161006dacd598/recipe/run_conda_forge_build_setup_win.bat#L19-L22
> so  the GHA workflow needs to set CONDA_BLD_PATH and adjust the expected paths

Looking at the ci-setup code, that doesn't look like it needs changing, so AFAIU we need to fix this in smithy

h-vetinari:
Fixes #2345 

Fix developed and tested in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/395; I kept the commit history (with references) where it made sense, to provide a logical progression.

### DIFF
--- a/conda_smithy/templates/github-actions.yml.tmpl
+++ b/conda_smithy/templates/github-actions.yml.tmpl
@@ -97,6 +97,7 @@ jobs:
 {%- endif %}
 
     - name: Build on Linux
+      id: build-linux
       if: matrix.os == 'ubuntu'
       env:
         CONFIG: {% raw %}${{ matrix.CONFIG }}{% endraw %}
@@ -134,6 +135,7 @@ jobs:
         ./.scripts/run_docker_build.sh
 
     - name: Build on macOS
+      id: build-macos
       if: matrix.os == 'macos'
       env:
         CONFIG: {% raw %}${{ matrix.CONFIG }}{% endraw %}
@@ -167,6 +169,7 @@ jobs:
 {% endfor %}
 
     - name: Build on windows
+      id: build-windows
       if: matrix.os == 'windows'
       shell: cmd
       run: |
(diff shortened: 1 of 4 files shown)


beckermr (approved):

h-vetinari:
Thanks for the quick review! :)

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -167,13 +169,18 @@ jobs:
...
       shell: cmd
       run: |
         set "flow_run_id=github_%GITHUB_RUN_ID%"
         set "remote_url=https://github.com/%GITHUB_REPOSITORY%"
         set "sha=%GITHUB_SHA%"
         call ".scripts\run_win_build.bat"
       env:
+        # default value; make it explicit, as it needs to match with artefact
+        # generation below. Not configurable for now, can be revisited later
+        CONDA_BLD_DIR: C:\bld
```
jaimergp, line 183: This is going to undo the performance improvements on MS-hosted Windows runners, not that we use those currently, but making a note here for the future.

h-vetinari: In what way does it "undo" them? AFAIU we don't have them yet on the GHA runners, i.e. we were defaulting to `C:\bld` when `CONDA_BLD_DIR` wasn't being set.

We currently only have an interface to do this on azure
```
azure:
  settings_win:
    variables:
      CONDA_BLD_PATH: C:\\bld\\
```
It would be nice if we could generalize some of that into something that's independent of the provider. Setting environment variables should be a no-brainer, swapfiles maybe (#1971), and artefact generation would also be a candidate (while it clearly has to be implemented per provider, the demand itself that artefacts be created shouldn't have to be provider-specific).

But once I started thinking about how to do this in a non-hacky way, it became clear that this would be a substantial undertaking that's almost 100% orthogonal to the fix that we need for pytorch right now, so I just deferred that. 😅 

I'm sincere about the "can be revisited later" though! 😁

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -185,47 +192,72 @@ jobs:
...
+        fi
+        echo "status=$STATUS" >> $GITHUB_OUTPUT
+
     - name: Prepare conda build artifacts
+      continue-on-error: true
       id: prepare-artifacts
       shell: bash
-      if: {% raw %}${{ always() }}{% endraw %}
+      # we do not want to trigger artefact creation if the build was cancelled
+      if: {% raw %}${{ steps.determine-status.outputs.status != 'cancelled' }}{% endraw %}
```
h-vetinari, line 223: OK, I didn't know GHA's inner workings well enough to understand that the `always()` here is necessary, as otherwise GHA might decide to skip steps after failure. Pushed a fix-up in 4c30cf1d9c446ce77553c298dc652418e6595a20

PR #2335, opened 2025-06-13
Fix incorrect merge for non-None ordering

h-vetinari:
Fixes #2331

PR #2353, opened 2025-07-13
fix variant merge for migrator with both `primary_key:` & `ordering:`

h-vetinari:
#2335 & #2338 didn't yet cover all bases, as we now find out since we bumped the global pinning to use GCC 14, while CUDA 12.6 is still on GCC 13. In particular, it wasn't possible to get the correct rendering in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/396.

This PR fixes that and adds a test.


tests/test_variant_algebra.py
```diff
@@ -124,6 +124,50 @@ def test_ordering_with_tail():
     assert res["cuda_compiler_version"] == ["None", "12.9"]
 
 
+@pytest.mark.parametrize("gcc_for_12dot6", ["13", "14", "15"])
+def test_ordering_with_primary_key(gcc_for_12dot6):
+    # ensure that GCC pins matching the respective CUDA versions get merged
+    # correctly; the GCC pin for 12.6 should never get picked since that CUDA
+    # version gets dropped in the merge of the primary keys (based on ordering).
```
beckermr, line 131: Can you remind me why a variant add operation drops a version instead of simply adding a version on top?

h-vetinari: The thing is, we're not in a variant-add operation; those get handled further up, i.e.
https://github.com/conda-forge/conda-smithy/blob/482e2de0d2f0617a9b3e8ae58ed5270b14d3832c/conda_smithy/variant_algebra.py#L308-L310

I don't know the origin/reasons behind the naming of the functions, but as the test shows (not least by lack of absence of `operation: key_add`), we're doing a normal "update-one-value-with-another" migrator.

The particularities are that we have two values per key, and they get merged with an `ordering:`. The ordering for the `primary_key:` must then be respected by the other keys in the zip (this is the core of the PR), otherwise the merged values may be jumbled completely (e.g. not form valid variants per set of values).

PR #2370, opened 2025-08-22
Remove slashes and commas from variant names

### LINKED ISSUE #2366: When using rattler cpython 3.14rc migration has / in files under .ci_support

h-vetinari:
Fix issues discovered by py3.14 migration. Currently, the fact that we have values like
```yaml
channel_sources:
- conda-forge,conda-forge/label/python_rc
```
in a zip-key, which ends up in the variant file names as
```
.ci_support/linux_aarch64_channel_sourcesconda-forge,conda-forge/label/python_rcpython3.14.____cp314.yaml
                                                                ^     ^
                                                                !!!!!!!
```
This causes obvious issues because these slashes cause folders where there shouldn't be any (c.f. #2366). Just removing slashes from the filenames fixes this.

Fixes #2366

### DIFF
--- a/tests/test_lint_recipe.py
+++ b/tests/test_lint_recipe.py
@@ -4030,7 +4030,6 @@ def test_lint_recipe_parses_ok():
 
 def test_lint_recipe_parses_forblock():
     with tempfile.TemporaryDirectory() as tmpdir:
-        # CRM cannot parse this one
         with open(os.path.join(tmpdir, "meta.yaml"), "w") as f:
             f.write(
                 textwrap.dedent(
@@ -4065,12 +4064,6 @@ def test_lint_recipe_parses_forblock():
         ), lints
         assert not any(
             hint.startswith("The recipe is not parsable by parser `conda-forge-tick")
-            for hint in hints
-        ), hints
-        assert any(
-            hint.startswith(
-                "The recipe is not parsable by parser `conda-recipe-manager"
-            )
             for hint in hints
         ), hints
         assert not any(
@@ -4159,12 +4152,6 @@ def test_lint_recipe_parses_v1_spacing():
             )
             for lint in lints
         ), lints
-        assert any(
-            hint.startswith(
-                "The recipe is not parsable by parser `conda-recipe-manager"
-            )
-            for hint in hints
-        ), hints
         assert not any(
             hint.startswith("The recipe is not parsable by parser `ruamel.yaml")
             for hint in hints
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -801,6 +801,9 @@ def dump_subspace_config_files(metas, root_path, platform, arch, upload, forge_c
             f"{platform}_{arch}",
             package_key(config, top_level_loop_vars, metas[0].config.subdir),
         )
+        # remove slashes that may occur in variant values and thus the filename
+        config_name = config_name.replace("/", "").replace(",", "")
+
         short_config_name = config_name
         conf_hash = hashlib.sha256(config_name.encode("utf-8")).hexdigest()[:8]
         # drone has a limit of 50, see https://github.com/conda-forge/conda-smithy/issues/1188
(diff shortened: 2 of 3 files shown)


beckermr (approved):

beckermr:
Does this get squash merged?

h-vetinari:
> Does this get squash merged?

I'd say no, but as you wish; https://github.com/conda-forge/conda-smithy/pull/2370/commits/b8a55f3c9d657f30480cb8176f49c424a0e28ca9 is really independent and just necessary to unbreak CI. I could squash the other two if you prefer

isuruf (approved):

schuylermartin45:
I think I'm missing some context, but here are my initial thoughts:

1) We recently changed how exceptions are handled. We catch all exceptions from the parsing stage, re-raise them as a `ParsingException` and then log the entire trace. That way folks down the line can write better exception handlers: https://github.com/conda/conda-recipe-manager/pull/412/files#diff-fb0998d8d273c4070015eb731daf03346cd252dcfde16dd46d14c0fa9270ab04R603

2) There have been a lot of improvements to our parsing success rate lately. I wonder if the confusion is around something that was once not supported and now is supported(?)

Do you have the offending file recipe file that we can look at?

tests/test_lint_recipe.py
```diff
@@ -4030,7 +4030,6 @@ def test_lint_recipe_parses_ok():
 
 def test_lint_recipe_parses_forblock():
     with tempfile.TemporaryDirectory() as tmpdir:
-        # CRM cannot parse this one
```
h-vetinari, line 4033: @schuylermartin45, this test (just unfold below) started not triggering the hint `"The recipe is not parsable by parser conda-recipe-manager"` anymore. Not sure what conda-smithy does internally there, perhaps just something changed w.r.t. what smithy expects CRM to do

schuylermartin45: Oh I see it/understand now. Let me take a look. We did not intentionally add support for the JINJA for loop syntax, so this is surprising.

beckermr: Thanks @h-vetinari. This is indeed why I started looking at this more carefully.

h-vetinari: One of the reasons I wanted to keep the commits separate is that now you can just revert a801ab1d27721926b10df0a916e396669fb81511 on top of main, and go from there. :)

schuylermartin45: I went through the release tags and this difference is a few months old. CRM v0.5.1 is the last version to throw-on-parse, CRM v0.6.0 (released in June) is the first to not throw.

schuylermartin45: Tracking: https://github.com/conda/conda-recipe-manager/issues/430

This is a little nuanced. I'm guessing this change is a consequence of our efforts to improve parsing compatibility. So to re-introduce this exception on purpose feels weird to me.

At the same time, I understand the reasoning why we should. We don't "support" editing or manipulating or rendering such loops.

One could argue the CRM CLI command `crm bump-recipe` is improved by ignoring this support gap (because a recipe can still be bumped to a new version) and `crm convert` should at least warn the user about the unsupported syntax.

There is a function I've left unimplemented for a long time, just in case we ever needed it: [`RecipeReader::has_unsupported_statements()`](https://github.com/conda/conda-recipe-manager/blob/main/conda_recipe_manager/parser/recipe_reader.py#L856). Maybe it is time to circle back on that idea.

PR #2391, opened 2025-09-04
avoid silent misrenders for missing zip_keys when using additional_zip_keys

h-vetinari:
When debugging an application of the 3.14 migration to the pyarrow feedstock (which uses the opt-in CUDA 11.8 migrator), I needed to add a local override due to some minor pinning changes (i.e. handle `"None"` in `cuda_compiler_version_min`). That caused me to miss the addition of `c_stdlib_version` to the migrator from https://github.com/conda-forge/conda-forge-pinning-feedstock/commit/17b696cae62aec00abe5732bb3639afd53c350fc

Surprisingly, the render didn't fail, but rather just produced the wrong result. The reason is that we didn't handle missing keys when we're in the `additional_zip_keys:` branch, but rather just ended up producing something like
```
{'c_stdlib_version': ['2.17', '2.28', None],
 'cuda_compiler': ['cuda-nvcc', 'cuda-nvcc', 'nvcc'],
 'cuda_compiler_version': ['None', '12.9', '11.8'],
 'cuda_compiler_version_min': ['11.8'],
 'migrator_ts': -1.0,
 'zip_keys': [['cuda_compiler_version', 'c_stdlib_version', 'cuda_compiler']]}
 ```
This was the result before this PR of `variant_add(start, cuda118_migrator)` in the test I added; note the `None` that got inserted into `c_stdlib_version`.

The solution to simply fail on missing keys, as usual.

### DIFF
--- a/news/fail-on-missing-key.rst
+++ b/news/fail-on-missing-key.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* Avoid spuriously adding values for missing zip_keys for migrations using `additional_zip_keys`.
+* Avoid spuriously adding values for missing ``zip_keys`` for migrations using ``additional_zip_keys``.
 
 **Security:**
 

### CODE REVIEW

File: news/fail-on-missing-key.rst [hunk 1]
```diff
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* Avoid spuriously adding values for missing keys for migrations using `additional_zip_keys`.
+* Avoid spuriously adding values for missing zip_keys for migrations using `additional_zip_keys`.
```
beckermr:
```suggestion
* Avoid spuriously adding values for missing ``zip_keys`` for migrations using ``additional_zip_keys``.
```

PR #2405, opened 2025-10-19
Update macOS GH-hosted runners

jaimergp:
Checklist
* [x] Added a ``news`` entry

GH will deprecated macos-13 in December, with brownouts during November. macos-15-intel replaces it, and will be removed in Fall 2027. At that point, we would have moved away from Intel builds anyway.

Part of https://github.com/conda/infrastructure/issues/1227.


h-vetinari (approved):
Thanks!

Issue #2234, opened 2025-01-28
Revamp the linter


jaimergp:
conda-smithy includes the linter, but in the past (way back!) it was already discussed whether to [break it out to a separate package](https://github.com/conda-forge/conda-smithy/issues/386), [merge with conda-verify](https://github.com/conda-forge/conda-smithy/issues/512), etc. That didn't happen and we are still here, with an [increasing backlog of bugs and feature requests](https://github.com/conda-forge/conda-smithy/issues/512).

My suggestion here is to:

- Create a separate package, e.g. `conda-forge-linter` or something like that.
- Architect it in a more modular way so we can enumerate all lint and hint rules, with their own unique identifier. This would allow for:
  - Better documentation (e.g. we could have a [site like `shellcheck`'s with all the rules](https://www.shellcheck.net/wiki/SC1000), with each lint linked from the CI logs).
  - Optional configurability. While lints are required, hints are not. Users could either silence some of them out if noisy, or conversely, we can introduce optional hints that users can opt into. Maybe this is not a goal, but it would be possible if we ever consider that.
- The new architecture should make it easier to contribute new lints/hints. I envision different "hook levels" where we can add new rules. For example:
  - Raw recipe lint: takes the whole recipe, unparsed.
  - Rendered recipe lint.
  - Per section lints (e.g. only requirements).
  - Non-recipe lints (e.g. conda_build_config.yaml).

... and in general, seek inspiration in other Python linters like [`pylint`](https://github.com/pylint-dev/pylint) or [`flake8`](https://github.com/PyCQA/flake8) (the code in this one is particularly easy to follow, and is well structured).

We have made great advances with the recent refactors for `recipe.yaml`, so I'm confident this is workable! Let me know your thoughts.

beckermr:
I support this but we need to maintain the current smithy apis both at the command line and in the code unless we plan to bump the smithy major version.

beckermr:
We've had trouble with api breaks during past refactors.

msarahan:
[...]

PR #2408, opened 2025-10-29
Consolidate linter messages in self-documented dataclasses

### LINKED ISSUE #92: Document conda-forge linting rules
Part of https://github.com/conda-forge/conda-smithy/issues/2234

To be closed by:

- https://github.com/conda-forge/conda-smithy/pull/2408

jaimergp:
Checklist
* [x] Added a ``news`` entry
* [X] Regenerated schema JSON if schema altered (`python conda_smithy/schema.py`)

Part of #2234. The whole point of this refactor is to generate a self-documenting JSON blob we can refer to (and import) from the website.

🔎 Preview at https://deploy-preview-2816--conda-forge-previews.netlify.app/docs/maintainer/linter/

We took all messages ever returned as a lint or hint, and made it a dataclass with an identifier and (obviously) a class name. The `message` itself is what gets sent to the CLI output (after rendering it with optional variables). The docstring makes for most of the auto-generated help. Note that I left many docstrings minimal to offer a smaller diff. We will add more docstrings little by little in subsequent PRs.

The `identifier` and unique class name serve several purposes:

- Unambiguous way to refer to a message and report problems
- Can be used as the anchor in the Markdown page to link to it
- Can be used as the key to pass to a potential "ignore_lint" rules in conda-forge.yml

This is the first step in a series of improvements to make the linter more helpful (e.g. record affected file, line and column numbers would be awesome too).

To add a new lint, one would just create a new subclass in the best fitting category, and then use that subclass as the message added to `lints` or `hints`.


conda_smithy/linter/lints.py
```diff
@@ -12,6 +12,54 @@
 from ruamel.yaml import CommentedSeq
 
 from conda_smithy.linter import conda_recipe_v1_linter
+from conda_smithy.linter.messages import (
```
bollwyvl: maybe generally using `from conda_smithy.linter import messages` (or shorter, like `msg`) would save some electrons?

conda_smithy/linter/lints.py
```diff
@@ -834,33 +776,18 @@ def flatten_reqs(reqs):
...
-                lints.append(stdlib_lint)
-
-    sysroot_lint = (
-        "You're setting a requirement on sysroot_linux-<arch> directly; this should "
-        f"now be done by adding a build dependence on {jinja_stdlib_c}, and "
-        "overriding `c_stdlib_version` in `recipe/conda_build_config.yaml` for the "
-        "respective platform as necessary. For further details, please see "
-        "https://github.com/conda-forge/conda-forge.github.io/issues/2102."
-    )
+            if not any(isinstance(lint, RecipeStdlibJinja) for lint in lints):
```
bollwyvl: Perhaps these kinds of `isinstance` checks could be deconflicted more tersely?

```py
class _BaseMessage:
    #: only report this class of message once
    singleton: ClassVar[bool] = false
```

... then

```py
class RecipeStdlibJinja:
    singleton = True
```

... then either with an overloaded `list` which checks for new items, or some other technique. 

```py
RecipeStdlibJinja().append_to(lints)
```

conda_smithy/linter/messages.py
```diff
@@ -0,0 +1,1300 @@
...
+If you want to add a lint or a hint, this is where you define the
+text, its identifier and the necessary variables.
+"""
+
+from dataclasses import asdict, dataclass
+from inspect import cleandoc
+from pathlib import Path
+from typing import ClassVar, Literal
+
+CATEGORIES: dict[str, str] = {
```
bollwyvl: These seem like good candidates for splitting out this juggernaut of a file...

conda_smithy/linter/messages.py
```diff
@@ -0,0 +1,1300 @@
...
+                [
+                    f"<a id='{cls.identifier}'></a>",
+                    f"### `{cls.identifier}`: `{cls.__name__}`",
+                    "",
+                    "",
+                    cleandoc(cls.__doc__),
+                    "",
+                    (
+                        f"<details>\n\n<summary>{cls.kind.title()} message</summary>\n\n"
+                        f"```text\n{cls.message}\n```\n\n</details>"
```
bollwyvl: maybe `markdown` for at least _some_ syntax highlighting?

conda_smithy/linter/messages.py
jaimergp: Done in [0c60144](https://github.com/conda-forge/conda-smithy/pull/2408/commits/0c601442936633077b3360d140ad9e3cc32794b1). `recipe.py` is still large.

conda_smithy/linter/messages.py
jaimergp: I've decided for a `.samples()` classmethod that provides good examples that render pretty too. Otherwise the placeholders look very weird.

conda_smithy/linter/lints.py
jaimergp: Hm, let me look into this, I like it.

jaimergp: Compromised with an `.append_if_absent` method that can check with either `isinstance` or `str(a) == str(b)`.

conda_smithy/linter/lints.py
jaimergp: Done

conda_smithy/linter/messages/__main__.py
```diff
@@ -0,0 +1,142 @@
...
+        cat: module for module, cats in module_to_categories.items() for cat in cats
+    }
+
+    if output_file is None:
+        # Let's check if we are in a repo or installed
+        maybe_repo_root = Path(__file__).parents[3]
+        if (maybe_repo_root / "README.md").is_file() and (
+            maybe_repo_root / ".git"
+        ).is_dir():
+            output_file = maybe_repo_root / "LINTER.md"
```
jaimergp: Not sure if this (repo root) is the best location for the output. To be discussed.

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,105 @@
...
+    deprecated_in = "3.56.0"
+    message = (
+        "Conflicting specification for minimum macOS deployment target!\n"
+        "If your conda_build_config.yaml sets `MACOSX_DEPLOYMENT_TARGET`, "
+        "please change the name of that key to `c_stdlib_version`!\n"
+        "Continuing with `max(c_stdlib_version, MACOSX_DEPLOYMENT_TARGET)`."
+    )
+
+
+# endregion
```
pb01ka: These classes are named with a CBC prefix but use RC identifiers:

- `CBCMacOSDeploymentTargetRename` has identifier `"RC-000"`
- `CBCMacOSDeploymentTargetBelow` has identifier `"RC-001"`
- `CBCMacOSDeploymentTargetBelowStdlib` has identifier `"RC-002"`

This is confusing because CBC semantically implies only `conda_build_config.yaml`, while RC implies recipe variants category. The naming and identifier scheme should agree.

I would suggest,

- Either rename class names to `RCMacOS...` to match RC identifiers
- Or change identifiers to CBC-00x if these are truly CBC-only rules

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -0,0 +1,96 @@
...
+
+    This is usually caused by a misconfiguration of your recipe file (e.g. `build.skip` is always
+    `true`, disabling all builds).
+    """
+
+    kind = "lint"
+    identifier = "CF-004"
+    message = (
+        "The feedstock has no `.ci_support` files and thus will not build any packages."
+    )
```
pb01ka: CF identifiers currently appear as CF-001, CF-002, CF-004, CF-005, CF-006.

If CF-003 was intentionally removed, the absence should be documented with a short note (for example, reserved/retired/deprecated) so that it is not interpreted as an accidental skip or copy/paste mistake.

conda_smithy/linter/messages/base.py
```diff
@@ -0,0 +1,83 @@
...
+from typing import ClassVar, Literal, Self
+
+CATEGORIES: dict[str, str] = {
+    "CBC": "Variants configuration (`conda_build_config.yaml`, `variants.yml`)",
+    "FC": "Feedstock configuration (`conda-forge.yml`)",
+    "R": "All recipe versions",
+    "R0": "Recipe v0 (`meta.yaml`)",
+    "R1": "Recipe v1 (`recipe.yaml`)",
+    "CF": "Issues specific to conda-forge",
+}
```
pb01ka: This file defines `CATEGORIES`. However it seems like `conda_smithy/linter/messages/__main__.py` collects category maps from each module directly. As a result, `base.CATEGORIES` appears unused. I would suggest to either,

- Remove it if not needed, or,
- make it the single source of truth used by the doc generator

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+
+from dataclasses import dataclass
+
+from conda_smithy.linter.messages.base import _BaseMessage
+
+CATEGORIES = {
+    "RC": "All recipe variants files",
+    "CBC": "Only `conda_build_config.yaml`",
+    "VC": "Only `variants.yml`",
```
pb01ka: Category VC is declared here but no `VC-***` classes exist.

This can cause conceptual noise and may produce empty category sections in generated docs depending on logic.

I would suggest fix,

- Remove VC until there are actual VC rules
- Or keep VC and add a clear comment that it is intentionally reserved for future rules (**most probable case**)

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1115 @@
...
+class ScriptShellcheckReport(_BaseMessage):
+    """
+    This issue is raised when `shellcheck` is enabled and detects problems
+    in your build `.sh` scripts.
+
+    See https://www.shellcheck.net/wiki/ for details on the shellcheck error codes.
+    """
+
+    kind = "hint"
+    identifier = "R-042"
```
pb01ka, line 808: Most recipe related messages in `recipe.py` use a `Recipe...` class prefix, but two use `Script...`,

- `ScriptShellcheckReport` (R-042)
- `ScriptShellcheckFailure` (R-043) (see below)

This is just a cosmetic review because naming consistency helps in discoverability and code navigation.

I would suggest renaming to `RecipeShellcheckReport` and `RecipeShellcheckFailure` for consistency.

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1115 @@
...
+                "We are collecting information to see which recipes are compatible with grayskull."
+            )
+        elif self.parser == "conda-recipe-manager":
+            msg += (
+                "The recipe can only be automatically migrated to the new v1 format "
+                "if it is parseable by conda-recipe-manager."
+            )
+        else:
+            msg += (
+                "Your recipe  may not receive automatic updates and/or may not be compatible "
```
pb01ka:
```suggestion
                "Your recipe may not receive automatic updates and/or may not be compatible "
```

Minor double space typo

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1115 @@
...
+    """
+
+    kind = "lint"
+    identifier = "R0-006"
+    message = (
+        "Using toolchain directly in this manner is deprecated.  Consider "
+        "using the compilers outlined "
+        "[here](https://conda-forge.org/docs/maintainer/knowledge_base.html#compilers)."
+    )
+
```
pb01ka: R0-006 class i.e., `RecipeLegacyToolchain` appears in the source before R0-004 i.e., `RecipeNoarchSelectorsV0` and R0-005 i.e., `RecipeJinjaDefinitions` (**see below**).

This is not behaviour breaking because docs are sorted by identifier, but source ordering can still confuse maintainers.

I would suggest to move `RecipeLegacyToolchain` after `RecipeJinjaDefinitions` (purely cosmetic and optional suggestion).

pb01ka (commented):
This is my first round of review which involves cosmetic suggestions (naming patterns, positioning of classes, seemingly unused variables, keys, etc).

conda_smithy/linter/messages/base.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        Provides one or more example instances of the error message. Used in documentation.
+        """
+        return []
+
+    def _render(self) -> str:
+        """
+        Formats the `.message` text by using the dataclass attributes.
+        """
+        return cleandoc(self.message.format(**self._render_attributes()))
```
pb01ka: I have a concern about `self.message.format` call here. This call means that every returned message (via `@property message`) string is parsed by Python format.

In Python format strings,

- `{placeholder}` is a replacement field and expects a key named `placeholder`
- `{}` is a positional replacement field
- Literal braces must be escaped as `{{` (or `\{`?) and `}}` (or `\}`?)

So if a property based message returns plain text that contains `{` or `}`, format will treat them as formatting syntax and may error.

**Typical failure modes**

- `KeyError` when a token like `{foo}` exists but `foo` is not in render attributes
- `ValueError`/`IndexError` for malformed or positional placeholders in unexpected text

**Why this is especially relevant for property based messages**

Note that property based messages may assemble text dynamically, including external strings (for example tool output), which may contain braces that are not formatting fields.

If those braces are not escaped, the unconditional format pass becomes fragile.

Some simple but concrete examples

1) Intended placeholder:

- Input template: `"Hello {name}"`
- `format(name="Karan")` -> `Hello Karan` (succeeded)

2) Accidental placeholder in literal text:

- Input text: `"parser output: {unknown}"`
- `format(...)` -> `KeyError` if `unknown` is not provided (failed because parser output was not in our control)

3) Safe literal brace text:

- Input text: `"parser output: {{unknown}}"`
- `format(...)` -> `"parser output: {unknown}"` (we successfully escaped the literal braces in parser output with `{{` and `}}`)

**What the current code already does well**

I noticed that `ScriptShellcheckReport` explicitly escapes braces in generated text (the `self.output_lines` member) via replacing `{` and `}` with doubled braces before returning, which avoids accidental format interpretation.

I would recommend to document this contract clearly:

- Any message string entering _render must either use valid placeholders supported by _render_attributes, or have all literal braces escaped (especially the ones obtained dynamically inside the definition of message property).

I think hardening the render behaviour to bypass format when there are no attributes may or may not be a good idea, but documenting the contract is likely the best option for now.

pb01ka (commented):
In this review, I have written about my concerns regarding `.format` processing in `_BaseMessage.render`.

conda_smithy/linter/messages/base.py
jaimergp: Good point. Hm... Can we enforce this somehow? I wish format had a graceful fallback mode that leaves things unrendered unless correct. Is there a different formatting option in Python that allows this? Template-strings? `template` module? Do we really need to use Jinja for this?

jaimergp: Addressing it a per https://github.com/jaimergp/conda-smithy/pull/2

conda_smithy/linter/messages/recipe_variants.py
jaimergp: They maybe CBC-only for now but they should actually cover variants.yaml too, not sure about it.

conda_smithy/linter/messages/recipe_variants.py
jaimergp: I think we lost `VC` in the review, but I'll add it back as a comment.

pb01ka (approved):
With the merging of,

- https://github.com/jaimergp/conda-smithy/pull/1
- https://github.com/jaimergp/conda-smithy/pull/2

All of my reviews are now addressed.

conda_smithy/linter/messages/__main__.py
bollwyvl: Would love to see a docs page rather than some at-rest markdown. With e.g. `sphinx-autodoc-typehints`, this script (and the workflow to keep the `.md` up-to-date) wouldn't be needed, and would encourage using language-level features rather than custom conventions.

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -0,0 +1,109 @@
...
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from conda_smithy.linter.messages.base import _BaseMessage
+
+if TYPE_CHECKING:
+    from typing import Self
+
+CATEGORIES = {
+    "CF": "conda-forge specific rules",
```
bollwyvl, line 16: Again thinking from a DX/LSP perspective: it looks like some things have a category prefix and some don't. Perhaps the categories could be subclasses/separate files:

```python
# cf.py
from .base import _BaseMessage, Lint

class CondaForge(_BaseMessage):
   """CF: conda-forge specific rules"""

class MaintainerExists(Lint, CF): 
    ...
```

Then, when imported, this would collapse various things into more-completable namespace-style forms, and avoid the need for big `*` imports.

```python
from . import messages as msg
msg.cf.MaintainerExists(maintainer=foo)
```

conda_smithy/linter/messages/__main__.py
jaimergp: I plan to import this in the Docusaurus website at conda-forge.org, so I need as simple as possible Markdown. I wouldn't be against generating _both_ things but for now I want to start simple.

conda_smithy/linter/messages/conda_forge.py
jaimergp: The categories are already separate modules, sort of. Some modules have several categories because they are related (e.g. recipe v0 and v1).

I'm planning that class names are also rule identifiers that we can use for ignores and other configuration, so we need to make sure they are unique across modules.

I can of course homogeneize the class names, but how do you suggest we implement `msg.cf.MaintainerExists(maintainer=foo)`? Right now you can do that via `msg.conda_forge.CFMaintainerExists`.

We can already distinguish the category via the identifiers, so I don't see the point of having a mixin to implement that too. (I'll add a property to obtain it directly from the identifier)

conda_smithy/linter/messages/__main__.py
jaimergp: Also, I'm checking https://github.com/tox-dev/sphinx-autodoc-typehints and we will need some custom bits to render dynamic fields anyway (and the examples), so I think the custom script is simpler for now.

conda_smithy/linter/messages/conda_forge.py
jaimergp: I gave `msg.cf.MaintainerExists`-style imports a try in the last few commits, let me know what you think! 🙏

conda_smithy/linter/messages/__main__.py
bollwyvl: Docosaurus doesn't appear to be on conda-forge, or have any built-in capability for documenting a Python package. The conda-forge.org site uses 3mb of JS/CSS/ttf (why no woff2?) to draw a static page: maybe not the right tool for this specific job of acting as a long-term, but always up-to-date identifier source? 

> `sphinx-autodoc-typehints`

One can combine/extend `autodoc` with _other_, potentially custom directives to generate intermediate (or checked in) files for the build, which again would stay up-to-date.

jaimergp:
pre-commit.ci autofix

conda_smithy/linter/messages/__main__.py
jaimergp: It's where all of our documentation is, including `conda-forge.yml` schema that lives in this repository. We used to have a few repositories producing sub-websites, which led to a very broken UX with inconsistent theming and poor accessibility practices. I don't want to open the door to that situation again.

The static Markdown page will be available in the repo always. We also get versioned help for free, since we can see the rendered MDs per version tag. Having it additionally rendered in the Docusaurus site will allow the Algolia backend to index it for the site search engine.

I'm ok with copying the Markdown file to a GH Pages deployment so we can link to `conda-forge.org/conda-smithy/lint-rules/` or similar if it's about having a lightweight version of it with a permalink, but I don't really want to maintain a separate Sphinx site with custom autodoc extensions, when a easy to understand script does the job.

beckermr (commented):
Have we considered storing the linter docs as a json blob within the package like we store the config schema?

This would ensure there is only one place to look for docs (the website) and might enable others to parse the json should they care to.

jaimergp:
> Have we considered storing the linter docs as a json blob within the package like we store the config schema?


The thought crossed my mind, yes. I thought that could be a good follow-up PR, but if we are ok with a slightly bigger diff, sure, I can add it. Do you mean in addition to the Markdown file, or as a replacement?

beckermr:
I mean as a replacement. This will help with the confusion on where docs live.

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -0,0 +1,109 @@
...
+}
+
+
+@dataclass(kw_only=True)
+class MaintainerExists(_BaseMessage):
```
h-vetinari, line 21:
```suggestion
class MaintainerMissing(_BaseMessage):
```

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    kind = "hint"
+    identifier = "CF-002"
+    message = "${package_hint}"
+    package_hint: str
+
+
+@dataclass(kw_only=True)
+class NoCiSupport(_BaseMessage):
```
h-vetinari, line 62: Aside from avoiding inconsistent capitalization of acronyms, I think the following would be clearer
```suggestion
class NoVariantConfigs(_BaseMessage):
```

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -0,0 +1,109 @@
...
+
+    kind = "lint"
+    identifier = "CF-003"
+    message = (
+        "The feedstock has no `.ci_support` files and thus will not build any packages."
+    )
+
+
+@dataclass(kw_only=True)
+class NoEmptyVariantsFile(_BaseMessage):
```
h-vetinari: the use of "No" here is inconsistent; `NoCiSupport` says "there are no variant configs, this is an error", while here, "no" is used as "files must not be empty". In one case the "no" is descriptive (of a state that exists but must be avoided), in one case the "no" is proscriptive (and not satisfied). I'd like us to pick one style or the other.

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1163 @@
...
+        "https://github.com/conda-forge/conda-forge.github.io/issues/2102."
+    )
+    recipe_version: RECIPE_VERSIONS
+
+    def _render_attributes(self):
+        return {"dollar": "$" if self.recipe_version == 1 else ""}
+
+
+@dataclass(kw_only=True)
+class StdlibOsx(_BaseMessage):
```
h-vetinari, line 623: I'd prefer to be consistent with keeping acronyms capitalized (like GHA & SPDX elsewhere)
```suggestion
class StdlibOSX(_BaseMessage):
```

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1163 @@
...
+    def _render_attributes(self):
+        bad_specs_list = []
+        for req_type, specs in self.bad_specs.items():
+            specs = [f"`{spec}`" for spec in specs]
+            bad_specs_list.append(f"- In section {req_type}: {', '.join(specs)}")
+        return {"output": self.output, "bad_specs_list": bad_specs_list}
+
+
+@dataclass(kw_only=True)
+class OsVersion(_BaseMessage):
```
h-vetinari, line 933:
```suggestion
class OSVersion(_BaseMessage):
```

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1163 @@
...
+    kind = "hint"
+    identifier = "R-050"
+    message = (
+        "Whenever possible python packages should use pip. "
+        "See https://conda-forge.org/docs/maintainer/adding_pkgs.html#use-pip"
+    )
+
+
+@dataclass(kw_only=True)
+class UsePyPiOrg(_BaseMessage):
```
h-vetinari, line 966:
```suggestion
class UsePyPIOrg(_BaseMessage):
```

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,106 @@
...
+    "CBC": "Recipe variants in `conda_build_config.yaml`",
+    # None yet, but let's reserve the VC- prefix:
+    # "VC": "Recipe variants in `variants.yaml`",
+}
+
+# region RC
+
+
+@dataclass(kw_only=True)
+class MacOSDeploymentTargetRename(_BaseMessage):
```
h-vetinari, line 20: I would stick to existing terms; either `MacOSX` like in the variable in question, or `OSX`; it may be even better though to leave it out:
```suggestion
class DeploymentTargetRename(_BaseMessage):
```

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,106 @@
...
+    identifier = "RC-000"
+    message = (
+        "The `MACOSX_DEPLOYMENT_TARGET` key in ${recipe_config_file} needs to be "
+        "removed or replaced by `c_stdlib_version`, appropriately restricted to osx"
+    )
+    recipe_config_file: str
+
+
+@dataclass(kw_only=True)
+class MacOSDeploymentTargetBelow(_BaseMessage):
```
h-vetinari, line 35:
```suggestion
class DeploymentTargetBelowBaseline(_BaseMessage):
```

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,106 @@
...
+    identifier = "RC-001"
+    message = (
+        "You are setting `c_stdlib_version` on osx below the current global "
+        "baseline in conda-forge (${baseline_version})."
+    )
+    baseline_version: str
+
+
+@dataclass(kw_only=True)
+class MoreThanOneFile(_BaseMessage):
```
h-vetinari, line 50:
```suggestion
class MoreThanOneConfigFile(_BaseMessage):
```

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,106 @@
...
+        "while `variants.yaml` may only be used with v1 recipes."
+    )
+
+
+# endregion
+# region CBC
+
+
+@dataclass(kw_only=True)
+class MacOSDeploymentTargetConflict(_BaseMessage):
```
h-vetinari, line 69:
```suggestion
class DeploymentTargetConflict(_BaseMessage):
```

conda_smithy/linter/messages/recipe_variants.py
```diff
@@ -0,0 +1,106 @@
...
+    message = (
+        "Conflicting specification for minimum macOS deployment target!\n"
+        "If your conda_build_config.yaml sets `MACOSX_DEPLOYMENT_TARGET`, "
+        "please change the name of that key to `c_stdlib_version`!\n"
+        "Continuing with `max(c_stdlib_version, MACOSX_DEPLOYMENT_TARGET)`."
+    )
+
+
+@dataclass(kw_only=True)
+class MacOSDeploymentTargetBelowStdlib(_BaseMessage):
```
h-vetinari, line 86: This can be deleted; it's become obsolete as of #2473

h-vetinari (commented):
Overall I like this a lot! 🤩 

Some comments on naming of the classes, but nothing major.

conda_smithy/linter/messages/recipe_variants.py
jaimergp: Added `deprecated_in = "3.56.0"`

conda_smithy/linter/messages/recipe_variants.py
jaimergp: I'd rather leave some indication of `MacOS` here because it may be used (well, not this one because it's deprecated) to configure the linter and "Deployment Target" doesn't sound super specific to me.

jaimergp:
> Have we considered storing the linter docs as a json blob within the package like we store the config schema?

Done! Must confess it looks very nice.

conda_smithy/data/linter-messages.json
```diff
@@ -0,0 +1,813 @@
...
+      "name": "MacOSDeploymentTargetConflict",
+      "identifier": "CBC-000",
+      "category": "CBC",
+      "kind": "lint",
+      "added_in": "<3.56",
+      "deprecated_in": "3.56.0",
+      "documentation": "https://github.com/conda-forge/conda-forge.github.io/issues/2102",
+      "message": "Conflicting specification for minimum macOS deployment target!\nIf your conda_build_config.yaml sets `MACOSX_DEPLOYMENT_TARGET`, please change the name of that key to `c_stdlib_version`!\nContinuing with `max(c_stdlib_version, MACOSX_DEPLOYMENT_TARGET)`.",
+      "examples": []
+    },
```
h-vetinari, line 23: It's not just deprecated; that warning is simply not emitted anymore. I don't see the value of keeping it?

tests/test_linter_messages_registry.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def test_linter_docs_up_to_date():
+    """
+    If this test fails, run 'python -m conda_smithy.linter.messages' to regenerate.
+    """
+    repo_root = Path(__file__).parent.parent
+    linter_messages_path = repo_root / "conda_smithy" / "data" / "linter-messages.json"
+    on_disk = json.loads(linter_messages_path.read_text().strip())
+    assert generate_docs(write=False) == on_disk
```
h-vetinari: Can we order the entries by their `identifier` here?

conda_smithy/data/linter-messages.json
jaimergp: Might have been emitted in the past and is there in a log somewhere.

jaimergp:
Website render component at https://deploy-preview-2816--conda-forge-previews.netlify.app/docs/maintainer/linter/

tests/test_linter_messages_registry.py
jaimergp: Aren't they already?

conda_smithy/data/linter-messages.json
h-vetinari: So? I don't see where historical linter warnings would be relevant. We should be documenting what smithy is currently doing, no? If not, can you please elaborate on your intentions here?

tests/test_linter_messages_registry.py
h-vetinari: This is currently an implementation detail; it's not tested anywhere that I can see. If you enforce an order here, _then_ it becomes part of the contract, and any changes to the implementation that reshuffle the ordering will show up as red CI even _after_ regenerating the json.

I was also thinking about a lint check to avoid duplicate identifiers.

conda_smithy/data/linter-messages.json
jaimergp: I'm copying what the folks at cargo clippy are doing: https://rust-lang.github.io/rust-clippy/master/index.html?groups=deprecated

My intention is to have better documentation. I think having historical record of why we made some decisions is worth the extra bytes in a text file.

tests/test_linter_messages_registry.py
jaimergp: > I was also thinking about a lint check to avoid duplicate identifiers.

Have you checked `tests/test_linter_messages_registry.py`?

jaimergp: > If you enforce an order here, then it becomes part of the contract

Added in [265b835](https://github.com/conda-forge/conda-smithy/pull/2408/commits/265b8353ac55ffdc1ee3e3f92a450ceb16ac2be5)

conda_smithy/data/linter-messages.json
h-vetinari: OK, fine by me.

jaimergp: For the record, these will be hidden by default in the website render.

conda_smithy/linter/messages/recipe.py
```diff
@@ -0,0 +1,1224 @@
...
+            cls(
+                command=[
+                    "shellcheck",
+                    "--enable=all",
+                    "--shell=bash",
+                    "--exclude=SC2154",
+                ],
+                output_lines=[
+                    "In ./recipe/build.sh line 337:",
+                    "" "ln -sf $PREFIX/$f $PWD/$f",
```
h-vetinari, line 846: I don't see the point concatenating an empty string
```suggestion
                    "ln -sf $PREFIX/$f $PWD/$f",
```

h-vetinari (approved):
This is great, thanks a lot for this! 🙏 

LGTM

conda_smithy/linter/messages/recipe.py
jaimergp: Oops, multicursor artifact I guess. Thanks!

news/2478-linter-messages.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Refactor linter messages as self-documenting dataclasses with automated Markdown outputs. (#2478)
```
beckermr:
```suggestion
* Refactor linter messages as self-documenting dataclasses with automated JSON outputs. (#2478)
```

conda_smithy/linter/messages/__init__.py
```diff
@@ -0,0 +1,22 @@
+"""
+This subpackage contains all linter messages. To add new linter rules,
+find the right category in the accompanying modules and add a new dataclass
+definition. Identifier must be unique across the whole subpackage.
```
beckermr: Does this mean the `messages` subpackage or the one of subsubpackages imported below?

conda_smithy/linter/messages/base.py
```diff
@@ -0,0 +1,151 @@
...
+
+    #: Shorthand to identify a given error, using two parts: category-instance; e.g. E-100
+    identifier: ClassVar[str]
+    #: The templated message that will be rendered when converted to string. Subclass with
+    #: a property if dynamic behavior is required.
+    message: ClassVar[str]
+    #: Whether the problem is a lint (error) or a hint (warning)
+    kind: ClassVar[Literal["lint", "hint"]]
+    #: conda-smithy version where the lint introduced
+    added_in: ClassVar[str] = "<3.56"
```
beckermr, line 48: We're at 3.61 now? We should update this right before merging.

conda_smithy/linter/errors.py
```diff
@@ -1,6 +1,7 @@
 """
-Collection of errors that can be raised by the linter.
+DEPRECATED. Moved to messages.py.
 """
 
-HINT_NO_ARCH = """Whenever possible python packages should use noarch.
-                See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-builds"""
+from conda_smithy.linter.messages.recipe import SuggestNoarch
+
+HINT_NO_ARCH = SuggestNoarch.message  # DEPRECATED
```
beckermr, line 7: We do not have a deprecation policy for smithy besides a major version bump. We can leave the "DEPRECATED" text here, but we won't be able to do anything about it.

beckermr (changes requested):
It appears this PR has changed the output types of the linting functions from strings to data classes?

This is an API change and will break code in other places. See for example: https://github.com/conda-forge/conda-forge-feedstock-ops/blob/main/conda_forge_feedstock_ops/lint.py

IDK what to do about this, but as is we cannot merge.

conda_smithy/linter/messages/base.py
jaimergp: The messages themselves (as strings) were introduced at least before 3.56. Only some were introduced afterwards and those are properly annotated.

jaimergp:
> It appears this PR has changed the output types of the linting functions from strings to data classes?

Well teeeechnically the return types only changed in two private functions (they start with `_`). It's true that we have changed the type of the containers we pass around. The function used in your reference above is `lint_recipe.main`, which still returns strings; see this `map()`: https://github.com/jaimergp/conda-smithy/blob/4fd683253a455bb73724c1b714dae15fb1603447/conda_smithy/lint_recipe.py#L438

conda_smithy/linter/messages/__init__.py
jaimergp: This is the subpackage; the other files are single files, i.e. modules. So identifiers need to be unique across all modules in this subpackage. There's a test checking it anyway.

jaimergp: Added [5262f51](https://github.com/conda-forge/conda-smithy/pull/2408/commits/5262f51f45572d67503c2380312d319a8b5c5260) just in case.

conda_smithy/linter/lints.py
```diff
@@ -48,13 +54,7 @@ def lint_section_order(
...
 
     if major_sections != section_order_sorted:
-        section_order_sorted_str = map(lambda s: f"'{s}'", section_order_sorted)
-        section_order_sorted_str = ", ".join(section_order_sorted_str)
-        section_order_sorted_str = "[" + section_order_sorted_str + "]"
-        lints.append(
-            "The top level meta keys are in an unexpected order. "
-            f"Expecting {section_order_sorted_str}."
-        )
+        lints.append(msg.r.SectionOrder(order=section_order_sorted))
```
beckermr: This is an API change in a public function, as are the rest of these across the various files?

jaimergp: Yes, that's what I meant with "It's true that we have changed the type of the containers we pass around".

jaimergp: I see two ways forward:

* Add `str()` everywhere in the append() functions, remove once we break API.
* Introduce a kwarg that controls whether we call `str()` or not, equally noisy, or even noisier with all those ifs.
* Bump major with the next release that includes this. But as discussed, this can also be the switch to calver.

beckermr: I think we add `str` in this PR. Those append lines are changed anyways, so that is the minimal amount of noise. A keyword would be noisier. 

We could alias `str` to a single global function at the top of the module that we can easily switch later, but that abstraction might not be worth the trouble. It might look like

```python
LINTER_CAST = str  # now
LINTER_CAST = lambda x: x  # later
```

Linters won't like us doing an assignment of a lambda. 🤷

jaimergp: Got it! Let me give it a try! I'll probably use an alias for `str()` so it's easy to remove later with simple search/replace.

jaimergp: Done. We will need to revert when we want to implement fancier outputs with structured data, and ignore-hints settings, but for now it's ok.

beckermr (approved):
LGTM! Thank you!

isuruf (changes requested):
Can you add some docs on how to add a new linter message?

conda_smithy/linter/conda_recipe_v1_linter.py
```diff
@@ -52,7 +55,7 @@ def lint_recipe_tests(
 
     if not test_section:
         if not outputs_section:
-            lints.append("The recipe must have some tests.")
+            lints.append(msg.r.RequiredTests().as_string())
```
isuruf, line 58: Not sure if this was discussed previously, but why are we using strings here?

jaimergp: See thread at https://github.com/conda-forge/conda-smithy/pull/2408#discussion_r3130333051. It's an API breakage otherwise.

jaimergp:
> Can you add some docs on how to add a new linter message?

Good point! See [272ce8d](https://github.com/conda-forge/conda-smithy/pull/2408/commits/272ce8dcd3aa4fdd7db806830c63db85d596950c).

conda_smithy/linter/conda_recipe_v1_linter.py
isuruf: Isn't it only `lint_recipe.main` that is used by external users? If so, we are free to change the API in these functions and have a new `lint_recipe.main_v2` that returns data classes, while `lint_recipe.main` will use the output from `main_v2` and keep the API from before. Then we can slowly switch downstreams to use `main_v2`.

beckermr: That's not how semantic versioning works. Any API interface is covered. We can break semver, but we should be clear that we are doing so.

isuruf: I'm not sure I understand. Do you consider any method that is not underscored to be part of the public API?

jaimergp: So far that's been our contract, yes.

isuruf: What's the plan then?

beckermr: merge this PR without API changes and put them in later when we have a formal deprecation policy?

jaimergp:
> Can you add some docs on how to add a new linter message?

@isuruf, can you take another look and reconsider your "changes requested" review? 🙏

isuruf (approved):

jaimergp:
Thanks everyone for your reviews!

Issue #2349, opened 2025-07-08
Unified interface for common `conda-forge.yml` tasks that should be independent of CI provider


h-vetinari:
In #2347, @jaimergp commented about
https://github.com/conda-forge/conda-smithy/blob/090c3f7b0244cfb949086bdc904dd3be0fd21aac/conda_smithy/templates/github-actions.yml.tmpl#L180-L183

as follows:
> This is going to undo the performance improvements on MS-hosted Windows runners, not that we use those currently, but making a note here for the future.

My response was:
> In what way does it "undo" them? AFAIU we don't have them yet on the GHA runners, i.e. we were defaulting to `C:\bld` when `CONDA_BLD_DIR` wasn't being set.
> 
> We currently only have an interface to do this on azure
> 
> ```yaml
> azure:
>   settings_win:
>     variables:
>       CONDA_BLD_PATH: C:\\bld\\
> ```
> 
> It would be nice if we could generalize some of that into something that's independent of the provider. Setting environment variables should be a no-brainer, swapfiles maybe (#1971), and artefact generation would also be a candidate (while it clearly has to be implemented per provider, the demand itself that artefacts be created shouldn't have to be provider-specific). [...]

So here's an issue to design a better interface. I think we want to be able to
* set environment variables
* create swap files (#1971)
* opt-in to artefact creation
* [potentially] set the runner image
* [potentially] free up disk space

All of these may be (and often are) dependent on the platform, but shouldn't _necessarily_ have to depend on the provider (except of course in the sense that each provider might have different - or sometimes even missing - implementations for the same interface).

So rather than the yaml above, we could have something like
```yaml
set_env_vars:
  all:
    - FOO=1
  linux:       # values from `all:` are merged with platform-specific ones
    - BAR=2
  osx:
    - BAZ=3
  win:
    - QUX=4
```
and then similarly for
```yaml
store_build_artifacts:
  all: true
  linux: true  # error if `all:` conflicts with platform-specific value
  osx: true
  win: true
```
etc.

h-vetinari:
> Do we actually need per-provider settings, rather than per-platform? I.e. is there ever a case we'd use two different providers per platform and want different setup?

I agree that the concerns we're trying to address are generally per platform, even though things like setting timeouts can be provider-specific. In many ways its more "obvious" to start from the provider-specific approach, because that's where each feature must be implemented separately, but I'd be very much in favour of starting over from a generic interface that might have different implementation status per provider.

[...]

PR #2410, opened 2025-10-30
GH-hosted runners: automatically resize Windows partitions to max

### LINKED ISSUE #2394: Improve disk space use for Windows jobs
### Comment:

xref https://github.com/conda-forge/pytorch-cpu-feedstock/pull/409

PyTorch started taking a lot more space to build recently. While trying to workaround it by increasing the CIrun runner size, two problems were noticed:
1. Builds are performed in `c:\bld` rather than (unless I'm mistaken) the much larger `d:\` partition. Effectively, choosing a larger builder had no effect, since it did only increase the `d:\` partition which was used for miniforge but not the `c:\` partition which was used for build.
2. The partitions don't take up all the available space, so increasing the builder disk size doesn't immediately benefit the builds.

I've tried modifying the files in the feedstock to force `d:\bld` instead but to no avail. Instead, for now, we've used the snippet in https://github.com/conda-forge/pytorch-cpu-feedstock/pull/409/commits/a6b03c9fffacbac1a9e43a01dd85845f4f6ded2d (as suggested by @aktech, thanks a lot!) to resize the `c:\` partition to use the additional space available.

Could we:

1. Include this (or a similar snippet) in the templates, to use all the space available in the runner?
2. Look into using `d:\` for build (and then resizing it instead), since it seems to be the larger partition?

mgorny:
Checklist
* [x] Added a ``news`` entry

Automatically resize all partitions on Windows running to use all the available space.  This is necessary to actually use the additional space provided on CIrun runners.  The code is copied verbatim from conda-forge/pytorch-cpu-feedstock where it was needed for a while now. Resizing is performed unconditionally, since there is no harm in doing it even if there is no extra space -- it will simply be a no-op then.

Fixes #2394

### DIFF
--- a/news/cirun-win-resize.rst
+++ b/news/cirun-win-resize.rst
@@ -1,7 +1,7 @@
 **Added:**
 
-* Windows GHA runners now automatically resize partitions to use the whole
-  available disk space. (#2394)
+* ``github_actions.resize_win_partitions`` option to resize partitions
+  on Windows GHA runners to use the whole available disk space. (#2394)
 
 **Changed:**
 


mgorny:
Okay, on Azure it took 2.5 minutes and decided that there's nothing to resize. Do we have any kind of Azure pipelines that might have larger disks? If not, I guess let's do it for GHA only.

h-vetinari:
FWIW, I think that ~1min would be acceptable to do unconditionally, given how few feedstocks we have that run on the windows server. Not against a variable either. Though we'll probably have to eventually redesign this more coherently in the context of #2349

news/cirun-win-resize.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* Windows GHA runners now automatically resize partitions to use the whole
```
beckermr: This needs to be updated.

mgorny: Updated.

Issue #2397, opened 2025-09-23
README contains non-existing output for single-item `outputs` recipes


jaimergp:
### Solution to issue cannot be found in the documentation.

- [x] I checked the documentation.

### Issue

Comes from [#general > conda-smithy rerender of v1 recipe with outputs names @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/conda-smithy.20rerender.20of.20v1.20recipe.20with.20outputs.20names/near/539930501)

Example:

- Given this [recipe](https://github.com/conda-forge/ninja-hep-ph-feedstock/blob/af33d00c5ecc651ba18c9a5e09c44fa0382cb689/recipe/recipe.yaml#L25-L27)
- [This line](https://github.com/conda-forge/ninja-hep-ph-feedstock/blob/1f8e5faabb3a8e65e4ab59cc238355ea9717d926/README.md?plain=1#L132) should not be there because `ninja-split` does not really exist.

I _think_ this might be rattler-build only, but needs investigation.

### Installed packages

```shell
N/A
```

### Environment info

```shell
N/A
```

PR #2418, opened 2025-11-10
Fix non-existing outputs in readme with v1 recipes

jsmolic:
- The issue appears due to the render info containing recipe name which gets mistakenly included in the recipe outputs even though it is not one. We can avoid this by simply omiting the 'name' value from the 'recipe' key

Closes: https://github.com/conda-forge/conda-smithy/issues/2397

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -247,7 +247,7 @@
   email: isuru.11@cse.mrt.ac.lk
   alternate_emails:
   - isuruf@gmail.com
-  num_commits: 684
+  num_commits: 685
   first_commit: 2017-01-08 22:03:03
   github: isuruf
 - name: Julian Rüth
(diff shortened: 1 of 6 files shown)


beckermr (changes requested):
I'm confused by this PR and am worried that it will break readmes for v0 recipes. Also the test looks off since the strings you check for in the tests don't appear to be in the recipe. We should be looking to ensure that "iregi-saffdsaf" is not in the recipe, right?

jsmolic:
Re last point, yes, that was PEBKAC, we should be checking for `iregi-split` presence, I've updated this now.

As far as v0 recipes go, it shouldn't break anything as they don't contain `recipe:` entry anyway. I tested this change on several v0 feedstocks and nothing changed in READMEs but I guess I could add another test for that?

PR #2430, opened 2025-11-29
macOS GHA: use dedicated `OSX_SDK_DIR`

mgorny:
Checklist
* [x] Added a ``news`` entry

Set a dedicated `OSX_SDK_DIR` for macOS GitHub Actions runners, so that SDKs can be downloaded without altering the main system install, as requested on https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/409.

### DIFF
--- a/conda_smithy/templates/run_osx_build.sh.tmpl
+++ b/conda_smithy/templates/run_osx_build.sh.tmpl
@@ -125,7 +125,8 @@ fi
 {% if build_setup -%}
 if [[ "${OSX_SDK_DIR:-}" == "" ]]; then
   if [[ "${CI:-}" == "" ]]; then
-    OSX_SDK_DIR=$(mktemp -d)
+    echo "Please set OSX_SDK_DIR to a directory where SDKs can be downloaded to. Aborting"
+    exit 1
   else
     # here we're definitely in the CI case
     export OSX_SDK_DIR=/opt/conda-sdks
--- a/conda_smithy/templates/run_osx_build.sh.tmpl
+++ b/conda_smithy/templates/run_osx_build.sh.tmpl
@@ -123,9 +123,24 @@ if [[ "${sha:-}" == "" ]]; then
 fi
 
 {% if build_setup -%}
-export OSX_SDK_DIR=/opt/conda-sdks
-/usr/bin/sudo mkdir -p "${OSX_SDK_DIR}"
-/usr/bin/sudo chown "${USER}" "${OSX_SDK_DIR}"
+if [[ "${OSX_SDK_DIR:-}" == "" ]]; then
+  if [[ "${CI:-}" == "" ]]; then
+    OSX_SDK_DIR=$(mktemp -d)
+  else
+    # here we're definitely in the CI case
+    export OSX_SDK_DIR=/opt/conda-sdks
+    /usr/bin/sudo mkdir -p "${OSX_SDK_DIR}"
+    /usr/bin/sudo chown "${USER}" "${OSX_SDK_DIR}"
+  fi
+else
+  if tmpf=$(mktemp -p "$OSX_SDK_DIR" tmp.XXXXXXXX 2>/dev/null); then
+      rm -f "$tmpf"
+      echo "OSX_SDK_DIR is writeable without sudo, continuing"
+  else
+      echo "User-provided OSX_SDK_DIR is not writeable for current user! Aborting"
+      exit 1
+  fi
+fi
 
 echo -e "\n\nRunning the build setup script."
 {{ build_setup }}
--- a/conda_smithy/templates/run_osx_build.sh.tmpl
+++ b/conda_smithy/templates/run_osx_build.sh.tmpl
@@ -128,7 +128,6 @@ if [[ "${OSX_SDK_DIR:-}" == "" ]]; then
     echo "Please set OSX_SDK_DIR to a directory where SDKs can be downloaded to. Aborting"
     exit 1
   else
-    # here we're definitely in the CI case
     export OSX_SDK_DIR=/opt/conda-sdks
     /usr/bin/sudo mkdir -p "${OSX_SDK_DIR}"
     /usr/bin/sudo chown "${USER}" "${OSX_SDK_DIR}"
--- a/conda_smithy/templates/run_osx_build.sh.tmpl
+++ b/conda_smithy/templates/run_osx_build.sh.tmpl
@@ -123,6 +123,10 @@ if [[ "${sha:-}" == "" ]]; then
 fi
 
 {% if build_setup -%}
+export OSX_SDK_DIR=/opt/conda-sdks
+/usr/bin/sudo mkdir -p "${OSX_SDK_DIR}"
+/usr/bin/sudo chown "${USER}" "${OSX_SDK_DIR}"
+
 echo -e "\n\nRunning the build setup script."
 {{ build_setup }}
 {%- endif %}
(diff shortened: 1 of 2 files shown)


isuruf:
Can you edit the github actions template only?

h-vetinari:
> Can you edit the github actions template only?

Why would it be desirable to _not_ do this consistently across the various different providers?

h-vetinari (approved):
Absent a strong technical reason not to, I want us to do this uniformly across providers. So this PR LGTM, thanks

isuruf (changes requested):
I mentioned in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/409#issuecomment-3572112047 that we don't want any scripts with sudo run in user systems. This is no different from that PR.

This PR is even worse than that PR since it ignores the user set `OSX_SDK_DIR` environment variable

h-vetinari:
> This PR is even worse than that PR since it ignores the user set `OSX_SDK_DIR` environment variable

You had pointed out in the same comment that leaving this user-configurable is a security concern:
> Imagine the user sets OSX_SDK_DIR to a system location and mindlessly types in the password prompt

So this is contradictory IMO. Either we take the user-provided `OSX_SDK_DIR` (and be exposed to the risk that this is in a system location), or we don't care about their variable, and just do `export OSX_SDK_DIR=/opt/conda-sdks` unconditionally (obviously we could save any pre-existing value of that variable and restore it after the build).

> I mentioned in [conda-forge/conda-forge-ci-setup-feedstock#409 (comment)](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/409#issuecomment-3572112047) that we don't want any scripts with sudo run in user systems. This is no different from that PR.

That was about installing SDKs in system locations. Here we're taking a path that we own unambiguously. I don't see the security concern angle on this, _especially_ because it's not configurable.

But OK, let's take a step back. Taking some [inspiration](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/f74ad06edec72f7e19a7320962d77f6947647425/recipe/run_conda_forge_build_setup_osx#L25-L32) from the current ci-setup, we can do some combination of conditionals like `[[ "${OSX_SDK_DIR:-}" == "" ]]`, `[[ "${CI:-}" == "" ]]`,  etc. The current ci-setup will write into whatever `$OSX_SDK_DIR` is set to, even if it's a system location. It sounded like you want to avoid that too. So this could look like

```bash
if [[ "${OSX_SDK_DIR:-}" == "" ]]; then
  if [[ "${CI:-}" == "" ]]; then
    OSX_SDK_DIR=$(mktemp -d)
  else
    # here we're definitely in the CI case
    export OSX_SDK_DIR=/opt/conda-sdks
    /usr/bin/sudo mkdir -p "${OSX_SDK_DIR}"
    /usr/bin/sudo chown "${USER}" "${OSX_SDK_DIR}"
  fi
else
  if tmpf=$(mktemp -p "$OSX_SDK_DIR" tmp.XXXXXXXX 2>/dev/null); then
      rm -f "$tmpf"
      echo "OSX_SDK_DIR is writeable without sudo, continuing"
  else
      echo "User-provided OSX_SDK_DIR is not writeable for current user! Aborting"
      exit 1
  fi
fi
```

I don't much care whether that code lives in smithy or in the ci-setup, but I do not want to make this provider-dependent, which would just be one more pointless detail that contributors need to be aware of (resp. hard to discover/debug).

h-vetinari (changes requested):
Requesting changes because "no provider-specific logic" is an eminently reasonable desire to avoid pointless divergences where there need be none. "no sudo on user calls" is _also_ a reasonable ask, that I'm happy to incorporate.[^1]

[^1]: within reason; if someone sets the `CI` environment variable locally, I have no pity.

> I'm open to exploring alternative solutions but please reach an agreement.

I've suggested a compromise above. Waiting for @isuruf to respond or make a counter-proposal.

isuruf:
For clarification, I meant here in `run_osx_build.sh` and not conda-forge-ci-setup.

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -123,6 +123,25 @@ if [[ "${sha:-}" == "" ]]; then
 fi
 
 {% if build_setup -%}
+if [[ "${OSX_SDK_DIR:-}" == "" ]]; then
+  if [[ "${CI:-}" == "" ]]; then
+    OSX_SDK_DIR=$(mktemp -d)
```
isuruf: Error here asking the user to provide a OSX_SDK_DIR value.

mgorny: Done. Lemme know if you think the message should be more/less detailed.

mgorny:
I'll do that tomorrow.

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -123,6 +123,26 @@ if [[ "${sha:-}" == "" ]]; then
 fi
 
 {% if build_setup -%}
+if [[ "${OSX_SDK_DIR:-}" == "" ]]; then
+  if [[ "${CI:-}" == "" ]]; then
+    echo "Please set OSX_SDK_DIR to a directory where SDKs can be downloaded to. Aborting"
+    exit 1
+  else
+    # here we're definitely in the CI case
```
h-vetinari: this comment was for exposition; not sure we need it in the final result.

h-vetinari (approved):
Thanks for the discussion, I'm glad we were able to find common ground. This LGTM

mgorny:
Filed https://github.com/conda-forge/conda-forge.github.io/pull/2684 for the news entry. Lemme know if that's what you had in mind.

isuruf:
> OTOH, assuming patching should not be necessary for the setup as of today, then the question arises why the examples given in that news entry ([1](https://github.com/conda-forge/openpmd-api-feedstock/blob/15f9b3648f087d3e06331d6ec9ddff0710300593/recipe/build.sh#L100-L107), [2](https://github.com/conda-forge/cartographer-feedstock/blame/1812f8c13bccbad20daf6ba079f7722cace93a15/recipe/conda_build_config.yaml#L19-L23)) would need to patch anything right now.

Depends on whether downstream packages are built on user systems or not.

h-vetinari:
> It's not just feedstocks that want to cater to user-installs. It's also downstreams built with this PR and upstream without this PR.

Right, and such a break is ~unavoidable if you want to move away from "using a system location by default", which was one of the side quests of this PR.

If you think this change as-is would be too disruptive now (and I mean primarily for our own feedstock landscape in terms of before/after), then please say so. There should be ways to mitigate it (e.g. add a symlink, and/or piggyback off something else, e.g. like when we switch to clang 21 as the default compiler version on osx), or we throw our hands up and keep this change GHA-only (while documenting the findings here until the next time we have to touch this aspect).

h-vetinari:
> such a break is ~unavoidable if you want to move away from "using a system location by default", which was one of the side quests of this PR.
> 
> If you think this change as-is would be too disruptive now (and I mean primarily for our own feedstock landscape in terms of before/after), then please say so.

Ping @isuruf

isuruf:
Let's see how much things break and revert if too disruptive.

wolfv:
In rattler-build we started to strongly isolate the environment. Should we let this env var pass in to the build environment?

isuruf:
No, this is an internal variable for CI. Not to be used by recipes.

PR #2438, opened 2025-12-17
determine `key_add:` involving `zip_keys:` based on full set of values, not just primary_key

h-vetinari:
The tegra migrator for CUDA 12.x wants to add a variant to the zip involving `cuda_compiler_version`; however, this is not possible due to the way the existing infrastructure works. We cannot set `arm_variant_type:` as `primary_key:`, because then it would not be injected into the right `zip_keys:`. On the other hand, if we use `primary_key: cuda_compiler_version`, the value matches another one that exists already, and thus gets discarded here
https://github.com/conda-forge/conda-smithy/blob/5b2575d3ce9a976a5866e4ddc2efc83d48283ef7/conda_smithy/variant_algebra.py#L194-L196

Fix this by considering whether the entire set of values belonging to the `zip_keys:` group containing `primary_key:` is already present in the existing set of variants before discarding it.

Despite adding additional capabilities, the logic in `variant_algebra.py` actually gets shorter, and that's despite additional comments and tightening when we raise errors. For example, we effectively re-enable
https://github.com/conda-forge/conda-smithy/blob/5b2575d3ce9a976a5866e4ddc2efc83d48283ef7/conda_smithy/variant_algebra.py#L190

This works because we're essentially vectorizing parts of the code and unifying quite a bit of redundant massaging.

### DIFF
--- a/news/determine_unique_variants_for_key_add.rst
+++ b/news/determine_unique_variants_for_key_add.rst
@@ -4,7 +4,7 @@
 
 **Changed:**
 
-* When running `operation: key_add` migrations, consider whether the variant being added exists already based on the full set of values in the `zip_keys:` involving `primary_key:`, not just based on the value of `primary_key:` itself.
+* When running `operation: key_add` migrations, consider whether the variant being added exists already based on the full set of values in the `zip_keys:` involving `primary_key:`, not just based on the value of `primary_key:` itself. (#2438)
 
 **Deprecated:**
 


h-vetinari:
pre-commit.ci autofix

news/determine_unique_variants_for_key_add.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* When running `operation: key_add` migrations, consider whether the variant being added exists already based on the full set of values in the `zip_keys:` involving `primary_key:`, not just based on the value of `primary_key:` itself.
```
beckermr:
```suggestion
* When running `operation: key_add` migrations, consider whether the variant being added exists already based on the full set of values in the `zip_keys:` involving `primary_key:`, not just based on the value of `primary_key:` itself. (#2438)
```

beckermr (approved):
LGTM! Nice PR!

Issue #2364, opened 2025-08-13
bug(lint): false positive for templating in version


lucascolley:
### Solution to issue cannot be found in the documentation.

- [x] I checked the documentation.

### Issue

https://github.com/conda-forge/numpy-typing-compat-feedstock/pull/2

```console
numpy-typing-compat-feedstock/recipe on 🎋 versions
❯ pixi run lint
✨ Pixi task (lint in smithy): conda-smithy lint --conda-forge recipe: (Lint the feedstock recipe)                                                                                                          recipe has some lint:
  Package version ${{ numpy_min }}.20250730 doesn't match conda spec: Invalid version '${{ numpy_min }}.20250730': invalid character(s)

numpy-typing-compat-feedstock/recipe on 🎋 versions 
❯ bat recipe.yaml
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: recipe.yaml
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ context:
   2   │   numpy_typing_compat_version: 20250730
   3   │   pkg_version: ${{ numpy_min }}.${{ numpy_typing_compat_version }}
   4   │ 
   5   │ package:
   6   │   name: numpy-typing-compat
   7   │   # e.g. '1.22.20250730' for numpy >=1.22,<1.23, numpy-typing-compat version 20250730
   8   │   version: ${{ pkg_version }}
{snip}
```

maybe this needs a fix in rattler-build itself? Anyway opening an issue here to track the conda-forge integration.

### Installed packages

```console
numpy-typing-compat-feedstock/recipe on 🎋 versions 
❯ pixi ls -e smithy -x
Environment: smithy
Package               Version  Build              Size       Kind   Source
conda-build           25.7.0   py312hb401068_0    750.3 KiB  conda  https://conda.anaconda.org/conda-forge/
conda-forge-ci-setup  4.21.2   py312hd35cd3b_100  78.7 KiB   conda  https://conda.anaconda.org/conda-forge/
conda-smithy          3.51.1   unix_pyhd81877a_0  182.2 KiB  conda  https://conda.anaconda.org/conda-forge/
rattler-build         0.45.0   hd3e8693_0         14.5 MiB   conda  https://conda.anaconda.org/conda-forge/
```

### Environment info

irrelevant I think!

PR #2443, opened 2025-12-31
fix(linter): ignore templated version more robustly

lucascolley:
gh-2402 covered the case that the version contains templating from the first character. This commit makes the check more robust to cover cases where the templating starts partway through the string, e.g. https://github.com/conda-forge/numpy-typing-compat-feedstock/pull/5.

Closes gh-2364.

ping @isuruf @beckermr from gh-2402

### DIFF
--- a/news/linter-version-templating
+++ b/news/linter-version-templating
@@ -1,0 +1,3 @@
+**Fixed:**
+
+* The linter no longer flags false positives for valid templating starting partway through the `version` field of v1 recipes


conda_smithy/linter/utils.py
```diff
@@ -212,7 +212,7 @@ def _lint_package_version(version: Optional[str]) -> Optional[str]:
 
     ver = str(version)
 
-    if ver.startswith("${{"):
+    if re.search(r"\$\{\{.*\}\}", ver):
```
isuruf:
```suggestion
    if "${{" in ver:
```

beckermr (approved):

lucascolley:
thanks :)

PR #2450, opened 2026-01-27
Enable native support for Apple Silicon runners on Azure

jaimergp:
Checklist
* [x] Added a ``news`` entry

- Part of https://github.com/conda-forge/conda-forge.github.io/issues/1781

This PR only adds opt-in support, via these conda-forge.yml changes (+ rerender):

```diff
- build_platform:
-   osx_arm64: osx_64
  conda_forge_output_validation: true
  github:
    branch_name: main
    tooling_branch_name: main
  provider:
    linux_aarch64: azure
    linux_ppc64le: azure
+   osx_arm64: azure
  test: native_and_emulated
  conda_build:
    pkg_format: '2'
```

In the future we may discuss whether we want to make `osx-arm64` a default plaform so there's no change in `provider` needed.

### DIFF
--- a/conda_smithy/data/conda-forge.json
+++ b/conda_smithy/data/conda-forge.json
@@ -1487,7 +1487,7 @@
               "type": "null"
             }
           ],
-          "default": "azure",
+          "default": null,
           "title": "Osx Arm64"
         },
         "win_32": {


conda_smithy/schema.py
```diff
@@ -455,7 +456,7 @@ class DefaultTestPlatforms(StrEnum):
         ]
         + [
             (str(plat), (Optional[ProviderType], Field(default="azure")))
-            for plat in ("linux_64", "osx_64", "win_64")
+            for plat in ("linux_64", "osx_64", "osx_arm64", "win_64")
```
jaimergp: I think this part should be done separately, later. Without this, users would need to opt-in manually via `providers`:

```diff
- build_platform:
-   osx_arm64: osx_64
  conda_forge_output_validation: true
  github:
    branch_name: main
    tooling_branch_name: main
  provider:
    linux_aarch64: azure
    linux_ppc64le: azure
+   osx_arm64: azure
  test: native_and_emulated
  conda_build:
    pkg_format: '2'

```

conda_smithy/configure_feedstock.py
```diff
@@ -1936,6 +1936,13 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
             config_rendered["DOCKER_IMAGE"] = data["config"]["docker_image"][-1]
         if forge_config["azure"]["store_build_artifacts"]:
             config_rendered["SHORT_CONFIG"] = data["short_config_name"]
+        if platform == "osx":
+            if data["build_platform"] == "osx-64":
+                config_rendered["VMIMAGE"] = "macOS-15"
+            elif data["build_platform"] == "osx-arm64":
+                config_rendered["VMIMAGE"] = "macOS-15-arm64"
+            else:
+                raise ValueError(f"Unknown build platform: '{data['build_platform']}'")
```
h-vetinari, line 1945: This needs to allow at least linux-64 as a build platform too. Aside from an example here
https://github.com/conda-forge/conda-smithy/blob/8568ea5950f93892bdec0ad5347b77a9487d042e/conda_smithy/schema.py#L597-L598

it's also used in the [wild](https://github.com/conda-forge/sse2neon-feedstock/blob/096e5a9d24685615937fb84db295b51bbaba0222/conda-forge.yml#L8) (found by [this](https://github.com/search?q=org%3Aconda-forge+%2F%28%3Fm%29build_platform%3A.*%5Cn%3F.*%5Cn%3F.*%5Cn%3F.*%5Cn%3F.*osx_%28arm%29%3F64%3A%5Cslinux%2F&type=code) search).

jaimergp: If the build platform for `osx-64` is `linux-64`, then `platform` here is `linux`, so this block doesn't match, does it?

h-vetinari: It's possible that I'm not getting something. Now that I look at it more closely still, I'm certainly confused that `data["build_platform"]` is a single value, rather than a mapping. Presumably the key (the target_platform) has already been evaluated. And `platform` here has nothing to do with that mapping either (except that it's grouping various possible build platforms into linux/osx/win buckets), due to the way that azure config needs to be sliced, i.e
```
.azure-pipelines/azure-pipelines-linux.yml
.azure-pipelines/azure-pipelines-osx.yml
.azure-pipelines/azure-pipelines-win.yml
```
?

jaimergp: Something like that it's what I gathered, yep. I did try the example above and the osx-64 job went to `azure-pipelines-linux`, as expected.

jaimergp:
I'm fine with merging. Folks could still manually edit their workflows anyway, so it's not like we are avoiding much by holding this. Worst case, the queue is too long and they switch back to cross-compiling.

PR #2451, opened 2026-01-27
Optimize `get_most_recent_version` with sharded repodata

jaimergp:
Checklist
* [X] Added a ``news`` entry

Fetching the latest version of packages via Anaconda.org's API is really slow, especially if there are lots of releases (like it's the case of conda-forge-pinning). Using sharded repodata (via `rattler`'s binding) reduces the fetch time from 20-30s each call to <2s with a cold cache. Warm cache is almost instantaneous.

This change brings the rerender of `dav1d-feedstock` from ~80s to ~20s.

### DIFF
--- a/conda_smithy/templates/azure-pipelines.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines.yml.tmpl
@@ -15,7 +15,7 @@ stages:
   jobs:
     - job: Skip
       pool:
-        vmImage: 'ubuntu-22.04'
+        vmImage: 'ubuntu-latest'
       variables:
         DECODE_PERCENTS: 'false'
         RET: 'true'
--- a/news/2449-ubuntu-skip.rst
+++ b/news/2449-ubuntu-skip.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Use ``ubuntu-latest`` in Azure Pipelines skip-control job. (#2449)
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>


conda_smithy/configure_feedstock.py
```diff
@@ -2581,16 +2583,30 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
...
+NameVersionUrlRecord = namedtuple("PackageRecord", ["name", "version", "url"])
 
-    PackageRecord = namedtuple("PackageRecord", ["name", "version", "url"])
-    return PackageRecord(name, pkg["version"], "https:" + pkg["download_url"])
+
+@cache
+def get_most_recent_version(name, include_broken=False) -> NameVersionUrlRecord:
+    channels = ["conda-forge"]
+    if include_broken:
+        channels.append("conda-forge/label/broken")
```
h-vetinari, line 2593: silly question perhaps, but do we know for certain that labels work as expected with `py-rattler`?

jaimergp: It should behave just like a regular channel (anaconda.org does the heavy lifting of exposing a `repodata.json` for each subdir in a label), so I'd assume yes. Let me prove it with a couple of queries.

jaimergp:
```python
>>> import rattler, asyncio
>>> g = rattler.Gateway()
>>> asyncio.run(g.query(channels=["conda-canary/label/dev"], platforms=["osx-arm64", "noarch"], specs=["conda>=25.11"
]))[0][0]
RepoDataRecord(url="https://conda.anaconda.org/conda-canary/label/dev/osx-arm64/conda-25.11.0.7+ge308518b4-py310_0.conda")
```

Looks like it does.

jaimergp: What I'm not so sure about is whether we are still applying the `broken` label to broken artifacts (or we simply remove them from `main`). Regardless, that would also happen with the old approach. And for the record, we are not calling this function with `include_broken=True` in any spots either.

PR #2452, opened 2026-01-27
Require `conda-recipe-manager>=0.9` and expect duplicate key errors

jaimergp:
Checklist
* [x] Added a ``news`` entry

https://github.com/conda/conda-recipe-manager/pull/460 introduced a new behavior where conda-recipe-manager raises on duplicate keys, which breaks the tests. Adjust the tests accordingly for the new version requirement.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -42,4 +42,5 @@ dependencies:
   - jsonschema
   - backports.strenum
   - exceptiongroup
+  - py-rattler >=0.21
   - rattler-build-conda-compat >=1.4.5,<2.0.0a0
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1,3 +1,4 @@
+import asyncio
 import copy
 import glob
 import hashlib
@@ -34,6 +35,7 @@ import conda_build.api
 import conda_build.render
 import conda_build.utils
 import conda_build.variants
+import rattler
 from conda.exceptions import InvalidVersionSpec
 from conda.models.match_spec import MatchSpec
 from conda.models.version import VersionOrder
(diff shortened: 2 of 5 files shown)


conda_smithy/linter/lints.py
jaimergp, line 1: The diff in this file is unnecessarily involved. I only moved the `ruamel-yaml` block a bit above so it can be expressed as an `elif recipe_version ...`. The rest is untouched; simply GH trying to make sense of it.

.github/workflows/tests.yml
```diff
@@ -31,9 +31,6 @@ jobs:
             python=3.11
             coverage
             coveralls
-            conda-recipe-manager>=0.4.1,<0.9
-            conda-souschef
-            conda-forge-tick
```
beckermr, line 36: We should run the full test suite with CRM etc.

beckermr: Ohhhh nvm. I did not add these to the env file because they are optional. This is fine.

jaimergp: Yes, it's part of `environment.yml` now.

environment.yml
```diff
@@ -14,6 +14,10 @@ dependencies:
   - mock
   - pytest
   - pytest-cov
+  # # Part of some optional lint tests
+  - conda-recipe-manager >=0.9
+  - conda-souschef
+  - conda-forge-tick
```
h-vetinari, line 20: conda-forge-tick is not installable on windows
```
error    libmamba Could not solve for environment specs
    The following package could not be installed
    └─ conda-forge-tick =* * is not installable because it requires
       └─ tar =* *, which does not exist (perhaps a missing channel).
```
we should move this dependency (singular) back to to `tests.yml`

PR #2460, opened 2026-02-04
fix: ensure we can recover if feedstock team is not added to a repo on first try

### LINKED ISSUE #2457: team updates fail if feedstock repo doesn't get added to the team
Right now we use the feedstock repo to find the github team in order to update the members. However, if we fail to add the repo to the team when the team is first made, then subsequent updates fail. PyGithub now has an api endpoint to get a team by its slug. We should use that to fetch the team to bypass this edge case.

beckermr:
As it says on the top...

Checklist
* [x] Added a ``news`` entry

closes #2457

### DIFF
--- a/news/team-update-fix.rst
+++ b/news/team-update-fix.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):
LGTM :)

PR #2461, opened 2026-02-04
reflect upcoming macos 11.0 baseline

h-vetinari:
For https://github.com/conda-forge/conda-forge.github.io/issues/2467; we should try to sync the next smithy release with when we actually flip the switch to 11.0 (xref https://github.com/conda-forge/conda-forge.github.io/pull/2721)

### DIFF
--- a/conda_smithy/linter/lints.py
+++ b/conda_smithy/linter/lints.py
@@ -924,7 +924,7 @@ def lint_stdlib(
                 versions = versions[::-1]
         return versions
 
-    baseline_version = ["10.13", "11.0"]
+    baseline_version = ["11.0", "11.0"]
     v_stdlib = sort_osx(cbc_osx.get("c_stdlib_version", baseline_version))
     macdt = sort_osx(cbc_osx.get("MACOSX_DEPLOYMENT_TARGET", baseline_version))
     sdk = sort_osx(cbc_osx.get("MACOSX_SDK_VERSION", baseline_version))
@@ -968,7 +968,7 @@ def lint_stdlib(
         # only warn if version is below baseline
         outdated_lint = (
             "You are setting `c_stdlib_version` below the current global baseline "
-            "in conda-forge (10.13). If this is your intention, you also need to "
+            "in conda-forge (11.0). If this is your intention, you also need to "
             "override `MACOSX_DEPLOYMENT_TARGET` (with the same value) locally."
         )
         if len(v_stdlib) == len(macdt):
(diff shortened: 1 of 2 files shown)


beckermr (approved):

jakirkham (approved):
Thanks Axel! 🙏

h-vetinari:
In https://github.com/conda-forge/conda-forge.github.io/pull/2721 we promised (loosely, to keep some flexibility) to move to 11.0 sometime next week. This is a rare case where coordination with smithy is more or less necessary (for correct linter messages), and so I'd suggest we hold up on merging this, to keep the ability to do a smithy release before that (if necessary).

The alternative is to merge this, and then pull the trigger on https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8208 early (in case an expedited smithy release becomes necessary).

conda_smithy/linter/lints.py
```diff
@@ -968,7 +968,7 @@ def sort_osx(versions):
         # only warn if version is below baseline
         outdated_lint = (
             "You are setting `c_stdlib_version` below the current global baseline "
-            "in conda-forge (10.13). If this is your intention, you also need to "
+            "in conda-forge (11.0). If this is your intention, you also need to "
```
jaimergp, line 971: These magic numbers may deserve a constant in the module (or somewhere else) so future updates are easier. WDYT?

h-vetinari: Yeah, not a big fan of hardcoding these values either, but since it only changes every couple of years, it was OK so far. Looking at that code again (that I mostly wrote 🙈), it has a lot of complexity for manually parsing CBCs, and somehow extracting the osx-relevant bits. It's pretends to understand the `osx-64` vs. `osx-arm64` distinction in hilariously bad manner (`osx-arm64` is the higher value), and then juggles and merges the various values that update things (what if `CBC` only uses a single `# [osx]` with no architecture distinction, etc.).

In other words, this is all immensely crufty code that I'd love to just rip out and throw away, rather than slap a bandaid on it, e.g. by modularizing out the values (though that in and of itself is obviously useful).

But yeah, we'll be stuck with CBCs for a long time still. 😑 

One thing we could do to simplify the situation is to simply ignore any value of `MACOSX_DEPLOYMENT_TARGET` now (we've been linting against its use for 1.5yrs...), and now that we're moving to macOS 11, we could also enforce that `c_stdlib_version` for osx must be specified without architecture selectors. That way we'd get a single value we can compare, which would substantially simplify the logic here.

But all that should IMO be moved to an issue as follow-up improvements, and not block this PR.

PR #2465, opened 2026-02-06
ensure we strip illegal characters out of variant paths

h-vetinari:
I couldn't actually check out https://github.com/conda-forge/pytorch-cpu-feedstock/pull/484 on my main machine (which has long paths enabled) due to the `:` in the variant values, which got persisted into the paths of the variants:
```
error: invalid path '.ci_support/linux_aarch64_c_stdlib_version2.28channel_targetsconda-forge_maincuda_compiler_version13.0github_actions_labelsghcr.iocirruslabsubuntu-runner-arm64:24.04is_rcFalse.yaml'
```

Ensure we sanitize characters that aren't path-legal everywhere.

### DIFF
--- a/news/path-characters.rst
+++ b/news/path-characters.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* Ensure that characters that aren't legal for filepaths are stripped from file names for variant configs.
+* Ensure that characters that aren't legal for filepaths are stripped from file names for variant configs. (#2465)
 
 **Security:**
 


news/path-characters.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Ensure that characters that aren't legal for filepaths are stripped from file names for variant configs.
```
jaimergp:
```suggestion
* Ensure that characters that aren't legal for filepaths are stripped from file names for variant configs. (#2465)
```

PR #2469, opened 2026-02-16
Lint against custom GHA workflows

jaimergp:
Checklist
* [x] Added a ``news`` entry

This won't prevent people from actually trying (they'd need to override the gitignore filters anyway), but at least they will know why we don't want this.

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -700,7 +700,7 @@ def run_conda_forge_specific(
     ):
         lints.append(
             "conda-forge feedstocks cannot have custom Github Actions workflows. "
-            "See https://conda-forge.org/docs/maintainer/infrastructure/#third-party-use-of-our-ci-services "
+            "See https://github.com/conda-forge/conda-forge.github.io/issues/2750 "
             "for more information. If you didn't add any custom workflows, please "
             "consider rerendering your feedstock to remove deprecated workflows."
         )


conda_smithy/lint_recipe.py
```diff
@@ -688,6 +689,16 @@ def run_conda_forge_specific(
...
             )
 
+    # 14: Do not allow custom Github Actions workflows
+    gha_workflows_dir = Path(recipe_dir, "..", ".github", "workflows")
+    gha_workflows = [*gha_workflows_dir.glob("*.yml"), *gha_workflows_dir.glob("*.yaml")]
+    if gha_workflows and len(gha_workflows) > 1 or gha_workflows[0].name != "conda-build.yml":
+        lints.append(
+            "conda-forge feedstocks cannot have custom Github Actions workflows. "
+            "See https://conda-forge.org/docs/maintainer/infrastructure/#third-party-use-of-our-ci-services "
+            "for more information."
```
beckermr: We should instruct people to rerender the feedstock if they think this lint is in error. For sure we will hit a feedstock with old automerge workflows that needs a rerender for the linter to pass.

jaimergp: I was thinking of adding this, but then I thought: if they are on a PR, and the feedstock is old enough to have workflows around, wouldn't have they rerendered already? It's part of the checklist after all. But sure, it wouldn't hurt.

jaimergp: See [6a6c72d](https://github.com/conda-forge/conda-smithy/pull/2469/commits/6a6c72d7549cbb81ea4daf4094e13f950ad1151b)

conda_smithy/lint_recipe.py
```diff
@@ -688,6 +689,22 @@ def run_conda_forge_specific(
...
+    ]
+    if gha_workflows and (
+        len(gha_workflows) > 1 or gha_workflows[0].name != "conda-build.yml"
+    ):
+        lints.append(
+            "conda-forge feedstocks cannot have custom Github Actions workflows. "
+            "See https://conda-forge.org/docs/maintainer/infrastructure/#third-party-use-of-our-ci-services "
+            "for more information. If you didn't add any custom workflows, please "
+            "consider rerendering your feedstock to remove deprecated workflows."
+        )
```
h-vetinari, line 706: I think it'd be worth a thought to create a link to some issue on https://github.com/conda-forge/conda-forge.github.io, for the simple reason that it provides visibility to us in core about where this lint pops up.

We did this for things like the stdlib migration (where it created so many links that the issue got hard to navigate; not something I anticipate here), and I regularly used that to check "what's going on" around conda-forge.

jaimergp: Smart! So by adding a link here then the lint message itself will "notify" the mentioned issue. That's great. I'll add it.

jaimergp: Actually I wonder whether we should use this technique more broadly, and use those issues as the documentation itself for all those lint rules, perhaps in a separate repository just for that?

h-vetinari: > Actually I wonder whether we should use this technique more broadly, and use those issues as the documentation itself for all those lint rules, perhaps in a separate repository just for that?

Some lints are clearly too noisy for that, e.g. the parser stuff and probably also stuff like `python_min` is IMO not high-value enough. For comparison, the stdlib meta issue had like ~2500 "items" from the POV of GH (and it becomes hard to navigate, much less find a comment somewhere in the middle of all those cross-references).

But I very much like the idea to create anchor issues that are just for audit purposes. We could keep a link at the top of the that issue that points to further information, if necessary.

The stdlib case was a bit special because that meta issue was genuinely meant to be explanatory (and started with "if you got directed here by the linter..."), but we don't _need_ to have extensive explanation in that issue for the tracking to be worth it[^1]; we could just say, "for reasons of traceability, we're keeping track of this lint in X".

[^1]: in fact, the stdlib case IMO proves that we don't want to mix issues with non-trivial discussion with the issue that "receives" the cross-references.

jaimergp: See [586837e](https://github.com/conda-forge/conda-smithy/pull/2469/commits/586837e6afbaae11cbd667ba6b578c3aeda86bbf)

jaimergp: Issue is https://github.com/conda-forge/conda-forge.github.io/issues/2750

jaimergp: > But I very much like the idea to create anchor issues that are just for audit purposes. We could keep a link at the top of the that issue that points to further information, if necessary.

This would be trivial to implement as part of https://github.com/conda-forge/conda-smithy/pull/2408, too.

PR #2470, opened 2026-02-17
Add "feedstock-name must not end with -feedstock" lint

pavelzw:
Checklist
* [x] Added a ``news`` entry

this happened in https://github.com/conda-forge/staged-recipes/pull/32139 and lead to the creation of https://github.com/conda-forge/vouch-feedstock-feedstock.

### DIFF
--- a/news/feedstock-name-not-end-with-feedstock.rst
+++ b/news/feedstock-name-not-end-with-feedstock.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Add "feedstock-name must not end with -feedstock" linter (#2470)
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>


isuruf:
I'd rather we edit https://github.com/conda-forge/conda-smithy/blob/b014232b4650dc30c22d9935ee633835da243fcb/conda_smithy/utils.py#L64 and remove `-feedstock` at the end

h-vetinari:
> I'd rather we edit [...] and remove `-feedstock` at the end

What does that buy us? Do you have a feedstock in mind where this would be necessary/beneficial?

There was a recent discussion on this with @beckermr somewhere, but I cannot find it for the life of me[^1].

His point was that chopping off `-feedstock` still breaks injectivity if it gets applied in several places: `foo-feedstock-feedstock` would coincide with `foo-feedstock` (and so on), which gets complicated by the fact that the team of the former would be called `foo-feedstock` etc.

This is all pretty theoretical as a concern, but I'd rather forbid names ending with `-feedstock` outright, rather than waste brain cycles and infra complexity on this.


[^1]: the term`"-feedstock"` is unsearchable, even using GH's regex search, which apparently only works for code, not issues/PRs; I remember https://github.com/conda-forge/conda-forge-webservices/pull/1237, but that's not cross-referenced with anything.

isuruf:
> What does that buy us? Do you have a feedstock in mind where this would be necessary/beneficial?

See OP. What I'm saying is chopping off `-feedstock` is better than adding a lint asking the user to do it themselves.

> His point was that chopping off -feedstock still breaks injectivity if it gets applied in several places

That's why we have a common function that is used by the rest of the code.

>  I'd rather forbid names ending with -feedstock outright, rather than waste brain cycles and infra complexity on this.

Same here.

h-vetinari:
> That's why we have a common function that is used by the rest of the code.

It would need to call itself recursively to deal with `foo-feedstock-feedstock-feedstock` etc.; I could live with that, but not all infra-code actually shares these functions currently; `"-feedstock"` handling is open-coded in [many places](https://github.com/search?q=org%3Aconda-forge%20%2F%5C%22-feedstock%5C%22%2F&type=code).

> See OP. What I'm saying is chopping off `-feedstock` is better than adding a lint asking the user to do it themselves.

It doesn't just apply to the feedstock name, but also team names (c.f. https://github.com/conda-forge/conda-forge-webservices/pull/1237). Rather than try to make the feedstock & team names for a _package_ called `foo-feedstock` legal, it seems way easier and more effective to me to lint on package (and probably output) names ending in `-feedstock`. While we're at that, might as well add `feedstock-name` to the mix.

Doing this "behind the scenes" as you suggest means we'd be loading up tech debt both on the feedstock and in our infrastructure, for a benefit that is not apparent to me. It's trivial to fix a lint for a disallowed name. Why would we want to allow `feedstock-name: foo-feedstock` and then do fragile clean-ups after that ~everywhere and in perpetuity (as well as adding confusion in the recipe because the field becomes a lie)?

h-vetinari:
> We finally got to what you are suggesting. Could have saved us a lot of time, if it was suggested first.

No need for snark. Rather than shooting down your suggestion and claiming "mine is better", I tried to understand why you suggested something that to me seemed net-negative on the tech debt side. I fail to see how anything went "wrong" here; a little back-and-forth is IMO normal interaction for resolving differences of opinion...?

isuruf:
> No need for snark. 

Sorry about that. I'll remove that comment

PR #2473, opened 2026-02-26
Linter improvements

h-vetinari:
What started out as just wanting to rip out the logic for `MACOSX_DEPLOYMENT_TARGET` became a deeper and deeper rabbit hole as various obvious follow-up clean-ups were blocked by the state of the code.

This PR:
* changes the logic regarding OSX pins (`c_stdlib_version` / `MACOSX_SDK_VERSION`) to simply ignore `MACOSX_DEPLOYMENT_TARGET`; this showed that the merge logic contained some subtle bugs where our tests were misconfigured after #2461, but didn't trigger (I don't expect this to affect many recipes, but strictly speaking we were missing some lints)
* break out osx-specific lints from `lint_stdlib` to a separate lint that's conda-forge-specific
* in that new check, also lint against presence of `MACOSX_DEPLOYMENT_TARGET`
* separately, also lint the content of `conda_build_config.yaml` in v1 recipes (e.g. the state before https://github.com/conda-forge/aws-sdk-cpp-feedstock/pull/993 didn't actually trigger a linter warning despite the outdated pins)
* add lint if both `conda_build_config.yaml` as well as `variants.yaml` are in the recipe folder
* consolidate and clean up a bunch of redundant logic in the main linter driver.
* skip a linter test that fails on windows

All the commits have been individually tested and I highly recommend to review them one-by-one (consider each as a mini-PR; the overall diff is too messy).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,7 +70,7 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: coveralls
-        uses: coverallsapp/github-action@master
+        uses: coverallsapp/github-action@main
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           path-to-lcov: ./coverage.lcov


conda_smithy/lint_recipe.py
```diff
@@ -420,11 +425,11 @@ def lintify_meta_yaml(
             hints,
         )
 
-    # 8. check for obsolete os_version
+    # 7. check for obsolete os_version
```
xhochy, line 428: Unrelated nit: We're a bit inconsistent in the comments with "1." vs "1:" ;)

conda_smithy/lint_recipe.py
```diff
@@ -264,37 +284,12 @@ def lintify_meta_yaml(
...
-        if forge_yaml_filename:
-            with open(forge_yaml_filename, encoding="utf-8") as fh:
-                forge_yaml = get_yaml().load(fh)
-        else:
-            forge_yaml = {}
-    else:
-        conda_build_config_keys = set()
-        forge_yaml = {}
+    # Interlude: load feedstock and recipe config
+    feedstock_config_keys = _get_feedstock_config(recipe_dir)
```
xhochy, line 288: What's the reason you only load this here and not at the very top of the function?

xhochy (commented):
Content-wise, this looks good; two minor nits that don't need to be fixed.

conda_smithy/lint_recipe.py
h-vetinari: Personally, I'd be more inclined to just remove the numbering. It serves no purpose, and makes it harder to move lints (if it ever becomes desirable to group or otherwise reorder checks)

conda_smithy/lint_recipe.py
h-vetinari: This PR was already ballooning in size, so I tried to keep changes local where possible. Further cleanups are of course possible.

h-vetinari:
Thanks for the review @xhochy!

conda_smithy/linter/lints.py
```diff
@@ -879,153 +999,6 @@ def flatten_reqs(reqs):
...
-            # where one key is specified for less arches than the other and
-            # let the rerender deal with the details
-            pass
-        else:
-            mismatch_lint = (
-                "Conflicting specification for minimum macOS deployment target!\n"
-                "If your conda_build_config.yaml sets `MACOSX_DEPLOYMENT_TARGET`, "
-                "please change the name of that key to `c_stdlib_version`!\n"
-                "Continuing with `max(c_stdlib_version, MACOSX_DEPLOYMENT_TARGET)`."
-            )
```
jaimergp, line 955: @h-vetinari while rebasing https://github.com/conda-forge/conda-smithy/pull/2408, I noticed that this lint is gone, but the equivalent checks in `configure_feedstock` are still present. Is that intentional or accidental? In case we need to remove those too.

jaimergp: See https://github.com/conda-forge/conda-smithy/blob/1a47d6c1b31fcc2392d0428eef9320aa2393f7ca/conda_smithy/configure_feedstock.py#L568

h-vetinari: > I noticed that this lint is gone, but the equivalent checks in `configure_feedstock` are still present. Is that intentional or accidental?

Any remaining treatment (except deprecation warnings) of `MACOSX_DEPLOYMENT_TARGET` is accidental. Sorry if I missed some occurrences

PR #2477, opened 2026-02-27
fix too-long variant paths on windows on GHA

### LINKED ISSUE #2476: Path lengths issues on GHA
We already have code that shortens variant filenames (most recently touched in #2233)
https://github.com/conda-forge/conda-smithy/blob/5dd5767b2e984641cfc69a5b7351a9ad845014d2/conda_smithy/configure_feedstock.py#L904-L907
and is pretty robust overall (windows default limit is 260 characters for the full path, so 200 should be quite conservative). However, on GHA, we're still running into issues, e.g. https://github.com/conda-forge/pytorch-cpu-feedstock/pull/491:

> Regarding the windows file lengths
> 
> ```
>    Error: error: unable to create file .ci_support/linux_aarch64_c_compiler_version13c_stdlib_version2.17channel_targetsconda-forge_maincuda_compiler_version12.9cxx_compiler_version13github_actions_labelscirun-openstack-gpu-2xlargeis_rcFalse.yaml: Filename too long
> ```
> 
> I just checked, and with the file name itself we're actually a fair bit away from the windows maximum of 260 characters for the full path, but the checkout somehow doubles the feedstock name in the containing folder
> 
> ```
> Initialized empty Git repository in C:/Users/runnerx/_work/pytorch-cpu-feedstock/pytorch-cpu-feedstock/.git/
>                                                                                  ^^^^^^^^^^^^^^^^^^^^^
>                                                                                         WHY?
> ```
> 
> which brings us over the edge
> 
> ```python
> >>> q = ".ci_support/linux_aarch64_c_compiler_version13c_stdlib_version2.17channel_targetsconda-forge_maincuda_compiler_version12.9cxx_compiler_version13github_actions_labelscirun-openstack-gpu-2xlargeis_rcFalse.yaml"
> >>> len(q)
> 207
> >>> len("C:/Users/runnerx/_work/pytorch-cpu-feedstock/pytorch-cpu-feedstock/")
> 67
> ```

The path is not shortened here because we're not counting `.ci_support/`, so the path length of the variant yaml itself is less than 200 characters and doesn't trigger. However, the really annoying part is how wastefully GHA checkout will just waste `len("pytorch-cpu-feedstock/") == 22` characters for absolutely no reason, and there isn't any option to override the work directory, despite years of users demanding a fix for this:
* https://github.com/actions/checkout/issues/197
* https://github.com/actions/checkout/issues/955
* https://github.com/actions/runner/issues/1676

Given that the length of feedstock names is variable, we should do something smarter here. Once we have https://github.com/conda-forge/.cirun/issues/168 up and running, this will not strictly be necessary anymore (because we can enable long paths in our VM images), but as long as we don't remove too much relevant content from the variant files, I think shortening to account for this is the least bad choice for now.

h-vetinari:
Fixes #2476

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -885,6 +885,11 @@ def dump_subspace_config_files(metas, root_path, platform, arch, upload, forge_c
     yaml.add_representer(OrderedDict, _yaml_represent_ordereddict)
     yaml.add_representer(str, _yaml_represent_str)
 
+    # get feedstock name; prefer explicit config, otherwise fall back to default
+    fn = metas[0].meta.get("extra", {}).get("feedstock-name")
+    feedstock_name = fn if fn else metas[0].meta.get("package", {}).get("name")
+    repo_name = f"{feedstock_name}-feedstock"
+
     platform_arch = f"{platform}-{arch}"
 
     result = []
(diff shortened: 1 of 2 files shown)


h-vetinari:
I used this PR to create https://github.com/conda-forge/pytorch-cpu-feedstock/commit/575b7f8feb9f99ed32c11cc8e3904e8a7d62f377, and this [works](https://github.com/conda-forge/pytorch-cpu-feedstock/actions/runs/22476589969).

PR #2478, opened 2026-02-27
feat: enable GHA for all of conda-forge

beckermr:
Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/tests/test_configure_feedstock.py
+++ b/tests/test_configure_feedstock.py
@@ -16,6 +16,10 @@ from rattler_build_conda_compat.loader import parse_recipe_config_file
 
 from conda_smithy import configure_feedstock
 from conda_smithy.configure_feedstock import _read_forge_config
+
+# This import of conda_smithy.schema can only be done in the test suite.
+# We cannot do it in the main package in order to avoid the need for
+# pydantic at runtime.
 from conda_smithy.schema import DEFAULT_PROVIDER
 from conda_smithy.utils import ensure_standard_strings
 


conda_smithy/schema.py
```diff
@@ -18,12 +18,18 @@
     from backports.strenum import StrEnum
 
 
-# use relative imports to ensure that we don't pick up the data paths from
-# a non-development conda-smithy installed in site-packages
-from .validate_schema import (  # noqa: TID252
-    CONDA_FORGE_YAML_DEFAULTS_FILE,
-    CONDA_FORGE_YAML_SCHEMA_FILE,
-)
+if __name__ == "__main__":
```
beckermr: Even with smithy installed via pip in dev mode, I could not get the generation command to work with relative imports.

jaimergp: I think you need to use `python -m conda_smithy.schema`

beckermr: Ahhh thank you! I added this bit in a few more spots to make it easier to find.

tests/test_configure_feedstock.py
```diff
@@ -16,6 +16,7 @@
 
 from conda_smithy import configure_feedstock
 from conda_smithy.configure_feedstock import _read_forge_config
+from conda_smithy.schema import DEFAULT_PROVIDER
```
jaimergp, line 23: We had considered `.schema` its own little world so far so we don't require pydantic at runtime. I guess it's ok for tests, though.

tests/test_configure_feedstock.py
```diff
@@ -792,6 +819,7 @@ def test_files_skip_render(render_skipped_recipe, jinja_env):
 
 
 def test_choco_install(choco_recipe, jinja_env):
```
jaimergp: Hm, seeing this test, I wonder if we have support for `choco` on non-Azure. Let me check.

jaimergp: Yep, we do. Perhaps we should add a test for GHA now that it's default to ensure we don't have regressions in the future (but it's a relatively niche feature).

tests/test_configure_feedstock.py
```diff
@@ -588,6 +604,8 @@ def test_readme_has_terminating_newline(noarch_recipe, jinja_env):
 
 
 def test_secrets(py_recipe, jinja_env):
```
jaimergp, line 680: Same for this one, this is important enough to make sure we are checking it for GHA too.

tests/test_configure_feedstock.py
jaimergp, line 540: Let's add another `pytest.raises` block for GHA too.

tests/test_configure_feedstock.py
```diff
@@ -344,6 +355,10 @@ def test_abi3_bools(py_abi3_recipe, jinja_env, caplog, request):
 
 
 def test_upload_on_branch_azure(upload_on_branch_recipe, jinja_env):
```
jaimergp, line 361: There's nothing specific about Azure here, so maybe we can parametrize this one to test both Azure and GHA.

jaimergp: Or, actually after expanding the diff further, maybe it's better to have its own test aside.

tests/test_configure_feedstock.py
beckermr: done!

tests/test_configure_feedstock.py
beckermr: done! I added a separate test below.

tests/test_configure_feedstock.py
beckermr: done!

tests/test_configure_feedstock.py
beckermr: done!

tests/test_configure_feedstock.py
beckermr: I added a comment above the import to remind people of this.

beckermr: I added a comment above the import to remind people of this.

PR #2481, opened 2026-03-03
Use recipe metadata instead of forge_dir basename for feedstock name

flferretti:
Conda-smithy rerender currently derives feedstock_name from the local checkout directory basename, which breaks in git worktrees or custom checkout paths. This patch derives the default feedstock name from recipe metadata, consistent with existing helper logic and GitHub repo creation paths.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/news/feedstock-name-not-from-metadata.rst
+++ b/news/feedstock-name-not-from-metadata.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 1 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -2948,7 +2949,16 @@ def main(
...
-    config["feedstock_name"] = os.path.basename(forge_dir)
+    metadata = _get_metadata_from_feedstock_dir(
+        forge_dir,
+        config,
+        conda_forge_pinning_file=exclusive_config_file,
+    )
+    package_name = get_feedstock_name_from_meta(metadata)
+    config["package"] = package_name
+    if not config["github"]["repo_name"]:
+        config["github"]["repo_name"] = f"{package_name}-feedstock"
```
isuruf: You can delete all these lines and keep only Line 2961

flferretti: You're right, thanks! Fixed with rebase

PR #2483, opened 2026-03-06
NEW: Set RATTLER_CACHE_DIR to a directory outside the Docker container

### LINKED ISSUE #2482: RATTLER_CACHE_DIR should be set to same path as conda-build uses
In build_steps.sh, we have the following:

```bash
cat >~/.condarc <<CONDARC

conda-build:
  root-dir: ${FEEDSTOCK_ROOT}/build_artifacts
pkgs_dirs:
  - ${FEEDSTOCK_ROOT}/build_artifacts/pkg_cache
  - /opt/conda/pkgs
solver: libmamba

CONDARC
```
This sets the package cache for conda to be in a folder which is mounted from outside the container. The reason for this is such that when running the same local build multiple times, the cache will be reused.

We should also do the following:

```bash
export RATTLER_CACHE_DIR="${FEEDSTOCK_ROOT}/build_artifacts/pkg_cache"

cat >~/.condarc <<CONDARC

conda-build:
  root-dir: ${FEEDSTOCK_ROOT}/build_artifacts
pkgs_dirs:
  - ${FEEDSTOCK_ROOT}/build_artifacts/pkg_cache
  - /opt/conda/pkgs
solver: libmamba

CONDARC
```
so that behavior for rattler-build is comparable.

Currently we have:
```bash
    # The default cache location might not be writable using docker on macOS.
    if ns.config.startswith("linux") and platform.system() == "Darwin":
        os.environ["CONDA_FORGE_DOCKER_RUN_ARGS"] = (
            os.environ.get("CONDA_FORGE_DOCKER_RUN_ARGS", "")
            + " -e RATTLER_CACHE_DIR=/tmp/rattler_cache"
        )
```
Which only sets this variable when building for Linux on macOS.

I believe the default cache location is inside the container which is not good because then  there is *no* caching since each container is single-use.

carterbox:
This makes the cache actually peristent across multiple builds on the same machine (build-locally.py) instead of single use.

Checklist
* [x] Added a ``news`` entry

Closes #2482

### DIFF
--- a/conda_smithy/templates/build_steps.sh.tmpl
+++ b/conda_smithy/templates/build_steps.sh.tmpl
@@ -20,6 +20,7 @@ export PYTHONUNBUFFERED=1
 export RECIPE_ROOT="${RECIPE_ROOT:-/home/conda/recipe_root}"
 export CI_SUPPORT="${FEEDSTOCK_ROOT}/.ci_support"
 export CONFIG_FILE="${CI_SUPPORT}/${CONFIG}.yaml"
+export RATTLER_CACHE_DIR="${FEEDSTOCK_ROOT}/build_artifacts/pkg_cache"
 {#- We are not handling CONDA_BLD_PATH here because it's already set in the condarc below #}
 
 cat >~/.condarc <<CONDARC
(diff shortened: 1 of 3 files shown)


conda_smithy/templates/build_steps.sh.tmpl
carterbox, line 31: We already set the conda-build cache to the same directory.

PR #2499, opened 2026-03-18
Allow mixing Github-hosted and self-hosted runners in `github_actions_labels`

jaimergp:
Checklist
* [X] Added a ``news`` entry

Also deprecates `github_actions.self_hosted` switch in `conda-forge.yml`, since it's not necessary anymore.

Some examples:

```yaml
# In conda_build_config.yaml:
github_actions_labels:
- ghcr.io/cirruslabs/ubuntu-runner-amd64-gpu:24.04  # [linux64]
- default                                           # [linux and aarch64]
- ghcr.io/cirruslabs/macos-runner:tahoe-xl          # [osx]
- cirun-azure-windows-4xlarge                       # [win]
```

- Linux x64 runs on Cirrus with GPU. Since the name contains `gpu`, extra variables are passed to Docker for the GPU passthrough.
- Linux ARM runs on whatever conda-smithy has as default for `linux-aarch64` (e.g. ubuntu-24.04-arm). We could have also omitted this line and it would have the same effect.
- macOS runs on Cirrus XL runner
- Windows runs on custom VM hosted in Azure, via Cirun

Note that I also added a workaround for the `self-hosted` label requirements. Some providers need it, some don't care, and some fail to provision the runner if present. The compromise is to not add them by default but, if needed, prefix the label string with `self-hosted@` and conda-smithy will "expand" it.

Of course, if `github_actions_labels` is NOT provided at all, conda-smithy will pick the default Github-hosted values.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -143,8 +143,8 @@ DEFAULT_PROVIDERS = {
     "linux_s390x": "github_actions",  # emulated
     "osx_64": "azure",
     "osx_arm64": "azure",
-    "win_64": "github_actions",
-    "win_arm64": "github_actions",
+    "win_64": "azure",
+    "win_arm64": "azure",  # emulated
 }
 
 NATIVE_CI_PROVIDER = {
--- a/conda_smithy/data/conda-forge.json
+++ b/conda_smithy/data/conda-forge.json
@@ -1536,7 +1536,7 @@
               "type": "null"
             }
           ],
-          "default": "github_actions",
+          "default": "azure",
           "title": "Win 64"
         },
         "win_arm64": {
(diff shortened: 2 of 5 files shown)


jaimergp:
I'll open a couple PRs in some feedstocks to show this live.

conda_smithy/configure_feedstock.py
```diff
@@ -1786,83 +1819,52 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
-        labels = conda_build.utils.ensure_list(
-            data["config"].get("github_actions_labels", [labels_default])[0]
+        hosted_default = list(
+            GITHUB_ACTIONS_RUNS_ON[data["build_platform"]]["hosted_labels"]
         )
 
-        if len(labels) == 1 and labels[0] == "hosted":
+        # labels may be overridden in conda_build_config.yaml
+        labels = data["config"].get("github_actions_labels") or ["hosted"]
+        if labels in (["hosted"], ["default"]):
```
isuruf, line 1842: I'd rather not have two different ways to say the same thing. `hosted` should be enough and let's remove `default`.

PR #2500, opened 2026-03-19
Unified configuration format, `store_build_artifacts` part

mgorny:
Checklist
* [x] Added a ``news`` entry

Introduce a unified configuration tree for more fine-grained settings of the CI setup. Currently, `store_build_artifacts` is implemented in the new format. The settings are rooted at `workflow_settings` top key, with its subkeys representing specific settings. The values can either represent a common value for all workflows, e.g.:

```yaml
workflow_settings:
  store_build_artifacts: true
```

or a list of multiple values with conditions describing when they apply, for example:

```yaml
workflow_settings:
  store_build_artifacts:
    - provider: github_actions
      value: true
    - platform: [win-64, linux-64]  # matched as an OR
      value: true
```

in which case the last value matching the given workflow is used.

Part of issue #2349.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1822,9 +1822,19 @@ def render_appveyor(jinja_env, forge_config, forge_dir, return_metadata=False):
 
 
 def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform):
-    forge_config.setdefault("workflow_settings_processed", {})[
-        "store_build_artifacts"
-    ] = set()
+    platform_templates = {
+        "linux": [
+            ".scripts/run_docker_build.sh",
+            ".scripts/build_steps.sh",
+        ],
+        "osx": [
+            ".scripts/run_osx_build.sh",
+        ],
+        "win": [
+            ".scripts/run_win_build.bat",
+        ],
+    }
+    template_files = platform_templates.get(platform, [])
 
     # Handle GH-hosted and self-hosted runners runs-on config
     # Do it before the deepcopy below so these changes can be used by the
@@ -1882,8 +1892,9 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
             )
         )
         if data["store_build_artifacts"]:
-            forge_config["workflow_settings_processed"]["store_build_artifacts"].add(
-                data["platform"].split("-", 1)[0]
+            script_suffix = ".bat" if platform == "win" else ".sh"
+            template_files.append(
+                f".scripts/create_conda_build_artifacts{script_suffix}"
             )
 
     build_setup = _get_build_setup_line(forge_dir, platform, forge_config)
@@ -1895,24 +1906,6 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
 
     forge_config = deepcopy(forge_config)
     forge_config["build_setup"] = build_setup
-
-    platform_templates = {
-        "linux": [
-            ".scripts/run_docker_build.sh",
-            ".scripts/build_steps.sh",
-        ],
-        "osx": [
-            ".scripts/run_osx_build.sh",
-        ],
-        "win": [
-            ".scripts/run_win_build.bat",
-        ],
-    }
-
-    template_files = platform_templates.get(platform, [])
-    if platform in forge_config["workflow_settings_processed"]["store_build_artifacts"]:
-        suffix = ".bat" if platform == "win" else ".sh"
-        template_files.append(f".scripts/create_conda_build_artifacts{suffix}")
 
     _render_template_exe_files(
         forge_config=forge_config,
@@ -1965,6 +1958,23 @@ def render_github_actions(jinja_env, forge_config, forge_dir, return_metadata=Fa
 def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
     build_setup = _get_build_setup_line(forge_dir, platform, forge_config)
 
+    platform_templates = {
+        "linux": [
+            ".scripts/run_docker_build.sh",
+            ".scripts/build_steps.sh",
+            ".azure-pipelines/azure-pipelines-linux.yml",
+        ],
+        "osx": [
+            ".azure-pipelines/azure-pipelines-osx.yml",
+            ".scripts/run_osx_build.sh",
+        ],
+        "win": [
+            ".azure-pipelines/azure-pipelines-win.yml",
+            ".scripts/run_win_build.bat",
+        ],
+    }
+    template_files = platform_templates.get(platform, [])
+
     if platform == "linux":
         yum_build_setup = generate_yum_requirements(forge_config, forge_dir)
         if yum_build_setup:
@@ -1990,10 +2000,6 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
         ratio = platform_counts[platform.split("-")[0]] / n_configs
         azure_settings["strategy"]["maxParallel"] = max(1, round(max_parallel * ratio))
 
-    forge_config.setdefault("workflow_settings_processed", {})[
-        "store_build_artifacts"
-    ] = set()
-
     for data in forge_config["configs"]:
         if not data["build_platform"].startswith(platform):
             continue
@@ -2014,32 +2020,15 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
             else:
                 raise ValueError(f"Unknown build platform: '{data['build_platform']}'")
 
-        config_rendered.update(get_workflow_settings(forge_config["workflow_settings"], "azure", data["platform"]))
+        workflow_settings = get_workflow_settings(forge_config["workflow_settings"], "azure", data["platform"])
+        data.update(workflow_settings)
+        config_rendered.update(workflow_settings)
         if config_rendered["store_build_artifacts"]:
-            forge_config["workflow_settings_processed"]["store_build_artifacts"].add(data["platform"].split("-", 1)[0])
             config_rendered["SHORT_CONFIG"] = data["short_config_name"]
+            script_suffix = ".bat" if platform == "win" else ".sh"
+            template_files.append(f".scripts/create_conda_build_artifacts{script_suffix}")
         azure_settings["strategy"]["matrix"][data["config_name"]] = config_rendered
         # fmt: on
-
-    platform_templates = {
-        "linux": [
-            ".scripts/run_docker_build.sh",
-            ".scripts/build_steps.sh",
-            ".azure-pipelines/azure-pipelines-linux.yml",
-        ],
-        "osx": [
-            ".azure-pipelines/azure-pipelines-osx.yml",
-            ".scripts/run_osx_build.sh",
-        ],
-        "win": [
-            ".azure-pipelines/azure-pipelines-win.yml",
-            ".scripts/run_win_build.bat",
-        ],
-    }
-    template_files = platform_templates.get(platform, [])
-    if platform in forge_config["workflow_settings_processed"]["store_build_artifacts"]:
-        suffix = ".bat" if platform == "win" else ".sh"
-        template_files.append(f".scripts/create_conda_build_artifacts{suffix}")
 
     forge_config["azure_yaml"] = yaml.dump(azure_settings)
     _render_template_exe_files(
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1872,15 +1872,14 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
                 data["gha_with_gpu"] = True
             data["gha_runs_on"].append(label)
 
-        store_build_artifacts = filter_conditional_values(
-            forge_config["workflow_settings"]["store_build_artifacts"],
-            provider="github_actions",
-            platform=data["platform"],
-            os=data["platform"].split("-", 1)[0],
-        )
-        data["store_build_artifacts"] = (
-            store_build_artifacts[-1].value if store_build_artifacts else None
-        )
+        for setting_key, setting_value in forge_config["workflow_settings"].items():
+            filtered = filter_conditional_values(
+                setting_value,
+                provider="github_actions",
+                platform=data["platform"],
+                os=data["platform"].split("-", 1)[0],
+            )
+            data[setting_key] = filtered[-1].value if filtered else None
 
     build_setup = _get_build_setup_line(forge_dir, platform, forge_config)
 
@@ -2035,16 +2034,17 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
                 config_rendered["VMIMAGE"] = "macOS-15-arm64"
             else:
                 raise ValueError(f"Unknown build platform: '{data['build_platform']}'")
-        store_build_artifacts = filter_conditional_values(
-            forge_config["workflow_settings"]["store_build_artifacts"],
-            provider="azure",
-            platform=data["platform"],
-            os=data["platform"].split("-", 1)[0],
-        )
-        config_rendered["STORE_BUILD_ARTIFACTS"] = (
-            store_build_artifacts[-1].value if store_build_artifacts else None
-        )
-        if config_rendered["STORE_BUILD_ARTIFACTS"]:
+        for setting_key, setting_value in forge_config["workflow_settings"].items():
+            filtered = filter_conditional_values(
+                setting_value,
+                provider="azure",
+                platform=data["platform"],
+                os=data["platform"].split("-", 1)[0],
+            )
+            config_rendered[setting_key] = (
+                filtered[-1].value if filtered else None
+            )
+        if config_rendered["store_build_artifacts"]:
             config_rendered["SHORT_CONFIG"] = data["short_config_name"]
         azure_settings["strategy"]["matrix"][data["config_name"]] = config_rendered
         # fmt: on
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2460,6 +2460,19 @@ def _read_forge_config(forge_dir, forge_yml=None):
         )
         logger.debug("Relevant schema:\n%s", json.dumps(err.schema, indent=2))
 
+    # Check for conflicting values.
+    if "store_build_artifacts" in file_config.get("workflow_settings", {}):
+        if "store_build_artifacts" in file_config.get("azure", {}):
+            raise ValueError(
+                "store_build_artifacts both in workflow_settings and azure. "
+                "Please remove the latter."
+            )
+        if "store_build_artifacts" in file_config.get("github_actions", {}):
+            raise ValueError(
+                "store_build_artifacts both in workflow_settings and "
+                "github_actions. Please remove the latter."
+            )
+
     # The config is just the union of the defaults, and the overridden
     # values.
     config = _update_dict_within_dict(file_config.items(), default_config)
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1872,14 +1872,15 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
                 data["gha_with_gpu"] = True
             data["gha_runs_on"].append(label)
 
-        # TODO: filter per specific job
+        store_build_artifacts = filter_conditional_values(
+            forge_config["workflow_settings"]["store_build_artifacts"],
+            provider="github_actions",
+            platform=data["platform"],
+            os=data["platform"].split("-", 1)[0],
+        )
         data["store_build_artifacts"] = (
-            filter_conditional_values(
-                forge_config["workflow_settings"]["store_build_artifacts"],
-                provider="github_actions",
-            )
-            or []
-        )[-1].value
+            store_build_artifacts[-1].value if store_build_artifacts else False
+        )
 
     build_setup = _get_build_setup_line(forge_dir, platform, forge_config)
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1980,30 +1980,6 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
     forge_config = deepcopy(forge_config)
     forge_config["build_setup"] = build_setup
 
-    platform_templates = {
-        "linux": [
-            ".scripts/run_docker_build.sh",
-            ".scripts/build_steps.sh",
-            ".azure-pipelines/azure-pipelines-linux.yml",
-        ],
-        "osx": [
-            ".azure-pipelines/azure-pipelines-osx.yml",
-            ".scripts/run_osx_build.sh",
-        ],
-        "win": [
-            ".azure-pipelines/azure-pipelines-win.yml",
-            ".scripts/run_win_build.bat",
-        ],
-    }
-    azure_store = filter_conditional_values(
-        forge_config["workflow_settings"]["store_build_artifacts"], provider="azure"
-    )
-    if any(x.value for x in azure_store):
-        platform_templates["linux"].append(".scripts/create_conda_build_artifacts.sh")
-        platform_templates["osx"].append(".scripts/create_conda_build_artifacts.sh")
-        platform_templates["win"].append(".scripts/create_conda_build_artifacts.bat")
-    template_files = platform_templates.get(platform, [])
-
     azure_settings = deepcopy(forge_config["azure"][f"settings_{platform}"])
     azure_settings.pop("swapfile_size", None)
     azure_settings.pop("install_atl", None)
@@ -2020,6 +1996,8 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
         )
         ratio = platform_counts[platform.split("-")[0]] / n_configs
         azure_settings["strategy"]["maxParallel"] = max(1, round(max_parallel * ratio))
+
+    store_build_artifacts = {x: False for x in ("linux", "osx", "win")}
 
     for data in forge_config["configs"]:
         if not data["build_platform"].startswith(platform):
@@ -2040,20 +2018,46 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
                 config_rendered["VMIMAGE"] = "macOS-15-arm64"
             else:
                 raise ValueError(f"Unknown build platform: '{data['build_platform']}'")
+        os = data["platform"].split("-", 1)[0]
         for setting_key, setting_value in forge_config["workflow_settings"].items():
             filtered = filter_conditional_values(
                 setting_value,
                 provider="azure",
                 platform=data["platform"],
-                os=data["platform"].split("-", 1)[0],
+                os=os,
             )
             config_rendered[setting_key] = (
                 filtered[-1].value if filtered else None
             )
         if config_rendered["store_build_artifacts"]:
+            store_build_artifacts[os] = True
+            forge_config[f"store_{os}_build_artifacts"] = True
             config_rendered["SHORT_CONFIG"] = data["short_config_name"]
         azure_settings["strategy"]["matrix"][data["config_name"]] = config_rendered
         # fmt: on
+
+    platform_templates = {
+        "linux": [
+            ".scripts/run_docker_build.sh",
+            ".scripts/build_steps.sh",
+            ".azure-pipelines/azure-pipelines-linux.yml",
+        ],
+        "osx": [
+            ".azure-pipelines/azure-pipelines-osx.yml",
+            ".scripts/run_osx_build.sh",
+        ],
+        "win": [
+            ".azure-pipelines/azure-pipelines-win.yml",
+            ".scripts/run_win_build.bat",
+        ],
+    }
+    for key, enabled in store_build_artifacts.items():
+        if enabled:
+            suffix = ".bat" if key == "win" else ".sh"
+            platform_templates[key].append(
+                f".scripts/create_conda_build_artifacts{suffix}"
+            )
+    template_files = platform_templates.get(platform, [])
 
     forge_config["azure_yaml"] = yaml.dump(azure_settings)
     _render_template_exe_files(
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1872,6 +1872,15 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
                 data["gha_with_gpu"] = True
             data["gha_runs_on"].append(label)
 
+        # TODO: filter per specific job
+        data["store_build_artifacts"] = (
+            filter_conditional_values(
+                forge_config["workflow_settings"]["store_build_artifacts"],
+                provider="github_actions",
+            )
+            or []
+        )[-1].value
+
     build_setup = _get_build_setup_line(forge_dir, platform, forge_config)
 
     if platform == "linux":
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -59,9 +59,9 @@ from conda_smithy.utils import (
     RATTLER_BUILD,
     HashableDict,
     ensure_standard_strings,
-    filter_conditional_values,
     get_feedstock_about_from_meta,
     get_feedstock_name_from_meta,
+    get_workflow_settings,
 )
 from conda_smithy.validate_schema import (
     CONDA_FORGE_YAML_DEFAULTS_FILE,
@@ -1875,15 +1875,11 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
                 data["gha_with_gpu"] = True
             data["gha_runs_on"].append(label)
 
-        for setting_key, setting_value in forge_config["workflow_settings"].items():
-            filtered = filter_conditional_values(
-                setting_value,
-                provider="github_actions",
-                platform=data["platform"],
-                os=data["platform"].split("-", 1)[0],
+        data.update(
+            get_workflow_settings(
+                forge_config["workflow_settings"], "github_actions", data["platform"]
             )
-            data[setting_key] = filtered[-1].value if filtered else None
-
+        )
         if data["store_build_artifacts"]:
             forge_config["store_any_build_artifacts"] = True
             if platform.startswith("win-"):
@@ -2018,17 +2014,9 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
                 config_rendered["VMIMAGE"] = "macOS-15-arm64"
             else:
                 raise ValueError(f"Unknown build platform: '{data['build_platform']}'")
+
+        config_rendered.update(get_workflow_settings(forge_config["workflow_settings"], "azure", data["platform"]))
         os = data["platform"].split("-", 1)[0]
-        for setting_key, setting_value in forge_config["workflow_settings"].items():
-            filtered = filter_conditional_values(
-                setting_value,
-                provider="azure",
-                platform=data["platform"],
-                os=os,
-            )
-            config_rendered[setting_key] = (
-                filtered[-1].value if filtered else None
-            )
         if config_rendered["store_build_artifacts"]:
             store_build_artifacts[os] = True
             forge_config[f"store_{os}_build_artifacts"] = True
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2495,13 +2495,13 @@ def _read_forge_config(forge_dir, forge_yml=None):
         # Check for conflicting old keys.
         if "store_build_artifacts" in file_config.get("azure", {}):
             raise ValueError(
-                "store_build_artifacts both in workflow_settings and azure. "
-                "Please remove the latter."
+                "`store_build_artifacts` both in `workflow_settings` and `azure` "
+                "sections. Please remove the latter."
             )
         if "store_build_artifacts" in file_config.get("github_actions", {}):
             raise ValueError(
-                "store_build_artifacts both in workflow_settings and "
-                "github_actions. Please remove the latter."
+                "`store_build_artifacts` both in `workflow_settings` and "
+                "`github_actions` sections. Please remove the latter."
             )
     else:
         # Convert old keys to new settings.
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1822,6 +1822,9 @@ def render_appveyor(jinja_env, forge_config, forge_dir, return_metadata=False):
 
 
 def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform):
+    store_build_artifacts_unix = False
+    store_build_artifacts_win = False
+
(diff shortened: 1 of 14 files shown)


conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -69,7 +69,6 @@ jobs:
         {{ secret }}: $({{ secret }})
 {%- endfor %}
 
-{%- if azure.store_build_artifacts %}
```
jaimergp, line 72: I think it's better if we keep this logic with something like `if any(data.store_build_artifacts for data in configs)` (or something pre-processed in the Python layer) so we don't render these blocks unnecessarily.

conda_smithy/configure_feedstock.py
```diff
@@ -1866,6 +1867,16 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
 
+        store_build_artifacts = filter_conditional_values(
+            forge_config["workflow_settings"]["store_build_artifacts"],
+            provider="github_actions",
+            platform=data["platform"],
+            os=data["platform"].split("-", 1)[0],
+        )
+        data["store_build_artifacts"] = (
+            store_build_artifacts[-1].value if store_build_artifacts else False
+        )
```
jaimergp: Trying to think of how we generalize this, I think the loop would look like this, right?


```suggestion
    for setting_key, setting_value in forge_config["workflow_settings"].items():
        filtered = filter_conditional_values(
            setting_value,
            provider="github_actions",
            platform=data["platform"],
            os=data["platform"].split("-", 1)[0],
        )
        data[setting_key] = (
            store_build_artifacts[-1].value if store_build_artifacts else False
        )
        # maybe store the result of `if any(...)` here too
```

conda_smithy/configure_feedstock.py
```diff
@@ -1866,6 +1867,16 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
             data["gha_runs_on"].append(label)
 
+        store_build_artifacts = filter_conditional_values(
+            forge_config["workflow_settings"]["store_build_artifacts"],
+            provider="github_actions",
+            platform=data["platform"],
+            os=data["platform"].split("-", 1)[0],
+        )
+        data["store_build_artifacts"] = (
+            store_build_artifacts[-1].value if store_build_artifacts else False
```
jaimergp: Can you elaborate the role of `-1` here?

conda_smithy/configure_feedstock.py
```diff
@@ -2459,6 +2486,33 @@ def _read_forge_config(forge_dir, forge_yml=None):
     # values.
     config = _update_dict_within_dict(file_config.items(), default_config)
 
+    if "store_build_artifacts" in file_config.get("workflow_settings", {}):
+        # Check for conflicting old keys.
+        if "store_build_artifacts" in file_config.get("azure", {}):
+            raise ValueError(
+                "store_build_artifacts both in workflow_settings and azure. "
```
jaimergp: let's quote or backtick the variable names

conda_smithy/configure_feedstock.py
```diff
@@ -2459,6 +2486,33 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
+                "Please remove the latter."
+            )
+        if "store_build_artifacts" in file_config.get("github_actions", {}):
+            raise ValueError(
+                "store_build_artifacts both in workflow_settings and "
+                "github_actions. Please remove the latter."
+            )
+    else:
```
jaimergp, line 2487: Love this.

conda_smithy/schema.py
```diff
@@ -460,6 +463,59 @@ class DefaultTestPlatforms(StrEnum):
 )
 
 
+def conditional_value(typ: type, default: Any = None) -> BaseModel:
```
jaimergp, line 468: Fancy 😎

conda_smithy/schema.py
```diff
@@ -460,6 +463,59 @@ class DefaultTestPlatforms(StrEnum):
...
+    )
+
+
+class WorkflowSettings(BaseModel):
+    store_build_artifacts: Optional[
+        Union[bool, conditional_value(bool, False), Nullable]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Store the conda build_artifacts directory as artifacts.
```
jaimergp: `build_artifacts` is only for Linux, IIRC. We may need to generalize this to "outputs of the build process" or something like that.


```suggestion
        Store the outputs of the build process as uploaded CI artifacts.
```

conda_smithy/schema.py
```diff
@@ -991,6 +1047,12 @@ class ConfigModel(BaseModel):
         """),
     )
 
+    workflow_settings: Optional[WorkflowSettings] = Field(
+        default_factory=WorkflowSettings,
+        description=cleandoc("""
+        Per-workflow settings.
```
jaimergp, line 1053: This description could use some general examples because it will be rendered in the website documentation.

conda_smithy/utils.py
```diff
@@ -244,3 +245,52 @@ def ensure_standard_strings(cfg: Any) -> Any:
...
+    provider: Optional[list[str]] = None
+
+
+def filter_conditional_values(
+    value: Any,
+    os: Optional[str] = None,
+    platform: Optional[str] = None,
+    provider: Optional[str] = None,
+) -> list[ConditionalValue]:
+    """Filter conditional values from conda-smithy.yml by specified
```
jaimergp:
```suggestion
    """Filter conditional values from conda-forge.yml by specified
```

conda_smithy/utils.py
```diff
@@ -244,3 +245,52 @@ def ensure_standard_strings(cfg: Any) -> Any:
...
+    criteria, and return the matching values as a list of ConditionalValue.
+    The matched criteria are removed, leaving only the unmatched criteria
+    in ConditionalValue instances."""
+
+    # direct value
+    if value is None:
+        return []
+    if not isinstance(value, list):
```
jaimergp, line 303: Since `value` is `Any`,  you may get all types of non-list iterables that should be converted to a list instead of wrapped, right?

conda_smithy/utils.py
```diff
@@ -244,3 +245,52 @@ def ensure_standard_strings(cfg: Any) -> Any:
...
+
+    # a list of condition-values
+    ret = []
+    for cv_item in value:
+        new_cv_item = {"value": cv_item["value"]}
+        for cond_name, cond_expect in (
+            ("os", os),
+            ("platform", platform),
+            ("provider", provider),
+        ):
```
jaimergp: Can you define this as a dictionary, once, outside of the loop?

conda_smithy/utils.py
```diff
@@ -244,3 +245,52 @@ def ensure_standard_strings(cfg: Any) -> Any:
...
+            if cond_expect is not None:
+                # filter by specified condition
+                if cond_expect not in cond_value:
+                    break
+            elif cond_name in cv_item:
+                # preserve the condition from from input
+                new_cv_item[cond_name] = cond_value
+        else:
+            ret.append(ConditionalValue(**new_cv_item))
+    return ret
```
jaimergp, line 329: The variables name in this block are a bit tricky to follow. Prefixes like `cv` and `cond` are not adding much information in this function, and were distracting to me. Can we think of some other names?

conda_smithy/utils.py
```diff
@@ -244,3 +245,52 @@ def ensure_standard_strings(cfg: Any) -> Any:
...
+def filter_conditional_values(
+    value: Any,
+    os: Optional[str] = None,
+    platform: Optional[str] = None,
+    provider: Optional[str] = None,
+) -> list[ConditionalValue]:
+    """Filter conditional values from conda-smithy.yml by specified
+    criteria, and return the matching values as a list of ConditionalValue.
+    The matched criteria are removed, leaving only the unmatched criteria
+    in ConditionalValue instances."""
```
jaimergp: I'd add some examples of the expected behavior to make it easier to understand.

conda_smithy/configure_feedstock.py
mgorny: Roughly, except for the `if store_build_artifacts else False` default. We'd either need another dict of defaults, or just store `None` here (or skip the key?), and handle that in the template.

conda_smithy/configure_feedstock.py
mgorny: In "pathological" case, multiple list items may match the same criteria, in which case we take the last one that matched.

conda_smithy/schema.py
mgorny: I've roughly copied it from the previous attempt :-).

conda_smithy/utils.py
mgorny: I don't think so, no.

This code is handling three cases:
1. List of ConditionalValue-dicts (`isinstance(value, list)`):
    ```yaml
    foo:
      - provider: azure
        value: foo
      - provider: github_actions
        value: bar
    ```
2. `None` if there is no `foo:`.
3. The direct value case like `foo: foo` or `foo: true`.

I see it as a best-effort solution.

conda_smithy/utils.py
mgorny: Good idea!

conda_smithy/utils.py
mgorny: Okay, I've tried hard to make it more readable but the variable names still suck. I'd appreciate some suggestions here.

conda_smithy/configure_feedstock.py
```diff
@@ -1867,15 +1870,21 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
+                setting_value,
+                provider="github_actions",
+                platform=data["platform"],
+                os=data["platform"].split("-", 1)[0],
+            )
+            data[setting_key] = filtered[-1].value if filtered else None
+
+        if data["store_build_artifacts"]:
+            forge_config["store_any_build_artifacts"] = True
+            if platform.startswith("win-"):
```
jaimergp: Is this the target platform or the build platform? This should be the build one, right?

conda_smithy/configure_feedstock.py
```diff
@@ -2490,13 +2500,13 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
+                "`store_build_artifacts` both in `workflow_settings` and `azure` "
+                "sections. Please remove the latter."
             )
         if "store_build_artifacts" in file_config.get("github_actions", {}):
             raise ValueError(
-                "store_build_artifacts both in workflow_settings and "
-                "github_actions. Please remove the latter."
+                "`store_build_artifacts` both in `workflow_settings` and "
+                "`github_actions` sections. Please remove the latter."
             )
```
jaimergp, line 2486: Are these ValueErrors bubbled up to a github comment during bot rerenders, or does it simply fail? 🤔 If the message is not bubbled up, I wonder whether this should be part of the linter instead

conda_smithy/schema.py
```diff
@@ -1051,6 +1051,16 @@ class ConfigModel(BaseModel):
         default_factory=WorkflowSettings,
         description=cleandoc("""
         Per-workflow settings.
+
+        ```yaml
+        workflow_settings:
+          store_build_artifacts:
+            # the ultimate value matched is used
```
jaimergp: "ultimate" as in "last"? as in "last match wins"?

conda_smithy/utils.py
```diff
@@ -261,36 +261,66 @@ def filter_conditional_values(
...
+      workflow_settings:
+        store_build_artifacts:
+          - provider: github_actions
+            value: true
+          - platform: [win-64, linux-64]  # matched as OR
+            value: true
+      ```
+
+      All items that matched the criteria will be returned, normalized to
+      `ConditianalValue` instances. Normally, you'd want to use the ultimate
```
jaimergp: Typo: ConditionalValue

conda_smithy/utils.py
```diff
@@ -261,36 +261,66 @@ def filter_conditional_values(
...
+      value from the list. If no items matched, an empty list will be returned.
 
-    # direct value
+    - A direct value, as from `store_build_artifacts: true`. In that case, a
+      list with a single `ConditionValue` instance will be returned.
```
jaimergp: Using a different class name here

conda_smithy/configure_feedstock.py
mgorny: It's the build platform, see:

https://github.com/conda-forge/conda-smithy/blob/1a47d6c1b31fcc2392d0428eef9320aa2393f7ca/conda_smithy/configure_feedstock.py#L1461-L1467

conda_smithy/configure_feedstock.py
mgorny: It leaves a comment pointing at the run log: https://github.com/conda-forge/triton-feedstock/pull/68#issuecomment-4157261920

conda_smithy/schema.py
mgorny: Yes.

conda_smithy/configure_feedstock.py
jaimergp: Hm, a bit obscure for my taste. What if we do not raise here, but instead log that we are ignoring the legacy option in favor of `workflow_settings`, and then an also add a hint in the linter?

jaimergp: Since we are working in #2408 too, let's defer this to later though.

mgorny: Okay.

conda_smithy/configure_feedstock.py
```diff
@@ -1871,6 +1875,22 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
             data["gha_runs_on"].append(label)
 
+        for setting_key, setting_value in forge_config["workflow_settings"].items():
+            filtered = filter_conditional_values(
+                setting_value,
+                provider="github_actions",
+                platform=data["platform"],
+                os=data["platform"].split("-", 1)[0],
+            )
+            data[setting_key] = filtered[-1].value if filtered else None
```
jaimergp: Let's add a comment here saying "we take -1 because 'last item wins'"

conda_smithy/schema.py
```diff
@@ -460,6 +463,59 @@ class DefaultTestPlatforms(StrEnum):
...
+                CI providers to set environment variable on (default: all)
+                """),
+            ),
+        ),
+        value=(
+            typ,
+            Field(
+                description=cleandoc("""
+                Option value
+                """),
```
jaimergp:
```suggestion
                description="Option value",
```

news/2500-unified-configuration.rst
```diff
@@ -0,0 +1,25 @@
+**Added:**
+
+* Add a new top-level ``workflow_settings`` configuration key for
+  ``conda-forge.yml`` that provides more fine-grained configuration for
+  workflows. initially, it can be used to control ``store_build_artifacts``.
```
jaimergp:
```suggestion
  workflows. Initially, it can be used to control ``store_build_artifacts``. (#2500)
```

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -249,11 +250,10 @@ jobs:
       id: prepare-artifacts
       shell: bash
       # we do not want to trigger artefact creation if the build was cancelled
-      if: {% raw %}${{ always() && steps.determine-status.outputs.status != 'cancelled' }}{% endraw %}
+      if: {% raw %}${{ always() && steps.determine-status.outputs.status != 'cancelled' && matrix.STORE_BUILD_ARTIFACTS }}{% endraw %}
       env:
         CI: github_actions
         CONFIG: {% raw %}${{ matrix.CONFIG }}{% endraw %}
-        SHORT_CONFIG: {% raw %}${{ matrix.SHORT_CONFIG }}{% endraw %}
```
jaimergp, line 256: Don't we need this? 🤔

conda_smithy/data/conda-forge.json
```diff
@@ -678,6 +758,15 @@
       "title": "Platforms",
       "type": "string"
     },
+    "PlatformsAliases": {
```
h-vetinari, line 761: This is not really an alias but a grouping. Something like `OSes` would also work.

conda_smithy/configure_feedstock.py
h-vetinari: I'm pretty strongly opposed to the "last item wins" approach, but I'm willing to live with it if we at least hint on it. That is, under `workflow_settings:`, we should check for all target_platforms if any key has >1 settings that apply. See my last comment in the issue.

conda_smithy/configure_feedstock.py
```diff
@@ -1871,6 +1875,22 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
+
+        if data["store_build_artifacts"]:
+            forge_config["store_any_build_artifacts"] = True
```
h-vetinari: Assuming it's unavoidable to generate `linux`/`osx`/`win`/`any`/whatever variables, we need to keep the original key intact. You cannot break `store_build_artifacts` into `store_{...}_build_artifacts`, no matter how much you may think that makes it easier to read. This breaks searching/grepping for keys in the source, which is a non-starter.

conda_smithy/configure_feedstock.py
```diff
@@ -2010,18 +2011,54 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
...
+                provider="azure",
+                platform=data["platform"],
+                os=os,
+            )
+            config_rendered[setting_key] = (
+                filtered[-1].value if filtered else None
+            )
+        if config_rendered["store_build_artifacts"]:
+            store_build_artifacts[os] = True
+            forge_config[f"store_{os}_build_artifacts"] = True
```
h-vetinari: Could we not simply write `store_build_artifacts` into `forge_config`? That would also avoid messing with the extra variables.

```suggestion
            forge_config[f"store_build_artifacts"] = store_build_artifacts
```

conda_smithy/utils.py
```diff
@@ -244,3 +245,82 @@ def ensure_standard_strings(cfg: Any) -> Any:
...
+    - A list of "conditional value" dicts, such as the value of
+      `store_build_artifacts` in:
+
+      ```yaml
+      workflow_settings:
+        store_build_artifacts:
+          - provider: github_actions
+            value: true
+          - platform: [win-64, linux-64]  # matched as OR
+            value: true
```
h-vetinari, line 283: A natural way to write this so that we can also deal with and-conditions would IMO be
```yaml
      workflow_settings:
        store_build_artifacts:
          - value: true
            provider: github_actions
            platform: osx-64          # AND-condition
         # OR-condition; linted to be non-overlapping
          - value: true
            platform: [win-64, linux-64]  # matched as OR
```

h-vetinari (commented):
Thanks for tackling this! Overall direction looks good, but I have some comments

conda_smithy/utils.py
mgorny: Well, that's how it works. The example is pure OR-OR.

conda_smithy/data/conda-forge.json
mgorny: That's an existing enum.

mgorny: As in, it is already used in other code. It just wasn't included in the generated data before.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: Hmm, I think it was removed accidentally indeed.

conda_smithy/configure_feedstock.py
mgorny: A "hint" as in `conda smithy lint` or a message here? I can make it fail when more than one item ends up matching too, it just didn't seem really necessary to me to keep it that restrictive.

conda_smithy/configure_feedstock.py
mgorny: Hmm, something like:

```
forge_config["processed_workflow_settings"]["store_build_artifacts"] = {
  'any': ...,
  'linux': ...,
  'win': ...,
}
```

perhaps? That should mirror the original structure without actually overwriting it.

conda_smithy/schema.py
```diff
@@ -460,6 +463,57 @@ class DefaultTestPlatforms(StrEnum):
...
+    store_build_artifacts: Optional[
+        Union[bool, conditional_value(bool, False), Nullable]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Store the outputs of the build process as uploaded CI artifacts.
+        """),
+    )
+
+
```
pb01ka, line 518: I think we should mention list in the type annotation of `store_build_artificats` because of the `default=[]`,

```suggestion
class WorkflowSettings(BaseModel):
    store_build_artifacts: Optional[
        bool,
        Union[bool, conditional_value(bool, False), Nullable],
        list[Union[bool, conditional_value(bool, False), Nullable]],
        Nullable,
    ] = Field(
        default=[],
        description=cleandoc("""
        Store the outputs of the build process as uploaded CI artifacts.
        """),
    )


```

conda_smithy/schema.py
```diff
@@ -460,6 +463,57 @@ class DefaultTestPlatforms(StrEnum):
 )
 
 
+def conditional_value(typ: type, default: Any = None) -> BaseModel:
```
pb01ka, line 468: I think this is a factory function. One usage of this I can spot is in the below snippet (taken from this file itself at L506),

```python
class WorkflowSettings(BaseModel):
    store_build_artifacts: Optional[
        Union[bool, conditional_value(bool, False), Nullable]
    ] = Field(
        default=[],
        description=cleandoc("""
        Store the outputs of the build process as uploaded CI artifacts.
        """),
    )
```

`conditional_value(bool, False)` is trivial. However, it always calls `create_model` from Pydantic which I think is non-trivial because as far as I know Pydantic doesn't caches the output of `create_model` for a given `type` and `default_value` (correct me if I'm wrong). So every `conditional_value(bool, False)` would result in creation and registration of a new class object.

It is only called once when `WorkflowSettings` is defined, so not an immediate issue. However, there are two issues of non-caching of `conditional_value` call,

1. Multiple calls would produce distinct classes with same name,

```python
>>> from conda_smithy.schema import conditional_value
>>> A = conditional_value(bool, False)
>>> B = conditional_value(bool, False)
>>> A is B
False
>>> A.__name__
"ConditionalValue_<class 'bool'>"
>>> B.__name__
"ConditionalValue_<class 'bool'>"
```

2. Duplicate work to create models with same parameters,

Building a new Pydantic model class seems like non-trivial work. There is no reason to redo it when the inputs are identical.

So considering the above two pointers and to be future safe (when multiple models will be created using call to `conditional_value`), I would recommend to decorate this function with `@lru_cache` (if the fact that Pydantic doesn't automatically cache the output of `create_model` is true),

```suggestion
@lru_cache
def conditional_value(typ: type, default: Any = None) -> BaseModel:
```

conda_smithy/configure_feedstock.py
pb01ka: As per my code analysis I am stating two facts here,

1. `forge_config["store_any_build_artifacts"]` is set to `True` here.
2. `_github_actions_specific_setup` is called in a loop inside `_render_ci_provider`,

```python
for platform in build_platforms.keys():
            platform_specific_setup(
                jinja_env=jinja_env,
                forge_dir=forge_dir,
                forge_config=forge_config,
                platform=platform,
            )
```

3. `forge_config["store_any_build_artifacts"]` isn't updated anywhere else in the code base (set to `False`).

This means none of the platforms ever set `store_any_build_artifacts` in `forge_config` then this key is absent. In the file, `conda_smithy/templates/github-actions.yml.tmpl` we have,

```
{%- if store_any_build_artifacts %}

    - name: Determine build outcome
      # this is to merge the status of the linux/osx/win builds into
      # something we can easily reuse during artefact generation
      id: determine-status
      if: {% raw %}${{ always() }}{% endraw %}
      shell: bash
      env:
      .
      .
      .
              name: {% raw %}${{ steps.prepare-  artifacts.outputs.ENV_ARTIFACT_NAME }}{% endraw %}
        path: {% raw %}${{ steps.prepare-artifacts.outputs.ENV_ARTIFACT_PATH }}{% endraw %}
        retention-days: {{ github_actions.artifact_retention_days }}
      continue-on-error: true
{%- endif %}
```

This means we will be using an undefined, `store_any_build_artifacts`. I think we should set `store_any_build_artifacts` to `False` in `conda_smithy/data/conda-forge.yml` (not currently the case).

Please correct me if I'm wrong in my analysis.

pb01ka (commented):
Dropping some comments related to cosmetics and semantics of the code. Please correct me if I made a mistake in my analysis.

conda_smithy/schema.py
mgorny: You're right, the type definitions is leftover from the previous design. Though your change doesn't seem syntactically valid. I'll fix it.

conda_smithy/schema.py
mgorny: I don't know the guts enough to tell if this will have any undesirable side effects or not.

conda_smithy/configure_feedstock.py
mgorny: I don't think that's really a problem but sure, I can change that. However, @h-vetinari already requested changing this logic, so it'll be part of that change once we agree on how exactly it should go.

conda_smithy/schema.py
pb01ka: @jaimergp @h-vetinari Do you have any thoughts on this?

jaimergp: I'm equally unfamiliar with pydantic but I think it makes sense to cache, worst case we observe something weird and remove it.

conda_smithy/configure_feedstock.py
```diff
@@ -1821,6 +1822,10 @@ def render_appveyor(jinja_env, forge_config, forge_dir, return_metadata=False):
 
 
 def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform):
+    forge_config.setdefault("workflow_settings_processed", {})[
```
jaimergp:
```suggestion
    forge_config.setdefault("workflow_settings_per_provider", {})[
```

Is this more accurate? "processed" sounded a bit too generic to me.

mgorny: Well, it's more like per-os right now.

i.e. it's: `workflow_settings_processed.store_build_artifacts: set[os_name]`

mgorny: My assumption was that it will be just a generic key for variety of preprocessed data we need to pass. Perhaps that's a bad assumption, and something like global `store_build_artifacts_on_os` would be better.

conda_smithy/configure_feedstock.py
```diff
@@ -1998,6 +1990,10 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
         ratio = platform_counts[platform.split("-")[0]] / n_configs
         azure_settings["strategy"]["maxParallel"] = max(1, round(max_parallel * ratio))
 
+    forge_config.setdefault("workflow_settings_processed", {})[
+        "store_build_artifacts"
+    ] = set()
```
h-vetinari: set-of-dicts is not an obvious choice of structure to me. Could you explain why you chose that?

conda_smithy/schema.py
```diff
@@ -991,6 +1047,23 @@ class ConfigModel(BaseModel):
...
 
+    workflow_settings: Optional[WorkflowSettings] = Field(
+        default_factory=WorkflowSettings,
+        description=cleandoc("""
+        Per-workflow settings.
+
+        ```yaml
+        workflow_settings:
+          store_build_artifacts:
+            # there can be at most one value for each workflow
```
h-vetinari, line 1058: could you add a test for this please?

conda_smithy/configure_feedstock.py
h-vetinari: Could we do something like:
```
workflow_settings_granular[ized]
workflow_settings_per_ci
workflow_settings_...
```
This obviously depends (somewhat at least) on the choice of data structure

conda_smithy/configure_feedstock.py
mgorny: What? It's a set of strings, i.e. OS names. We could technically do `dict[str, bool]`, but that would require unnecessary redundancy (i.e. filing `false` values). I _suppose_ a `dict[str, Any]` would make sense if we wanted all workflow-settings-per-os, but it's not clear to me if we will ever actually need any other of the settings preprocessed like that.

conda_smithy/configure_feedstock.py
```diff
@@ -2464,6 +2483,33 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
+                "`github_actions` sections. Please remove the latter."
+            )
+    else:
+        # Convert old keys to new settings.
+        config["workflow_settings"]["store_build_artifacts"].append(
+            {
+                "provider": "azure",
+                "value": config["azure"]["store_build_artifacts"],
+            }
+        )
```
h-vetinari, line 2494: I was talking about this with set-of-dicts. Here you're appending a dict to a set, right?

mgorny: No, that's a list of dicts, i.e. the:

```
workflow_settings:
  store_build_artifacts:
    - provider: azure
      value: {old_value}
```

thing (sets use `.add()` and `dict`s are unhashable). It's the same list used by the user; we're appending the old value for backwards compatibility, so we wouldn't have to check for both new and old values separately.

conda_smithy/configure_feedstock.py
h-vetinari: ok, that was my bad, not least because I mixed up `config["workflow_settings"]["store_build_artifacts"]` with `forge_config["workflow_settings_processed"]["store_build_artifacts"]`. Still, a set is a bit of a strange choice to me, even though I can see how you ended up there.

> I _suppose_ a `dict[str, Any]` would make sense if we wanted all workflow-settings-per-os, but it's not clear to me if we will ever actually need any other of the settings preprocessed like that.

Isn't this a logical extension for other cases from #2349? E.g. when setting swapfiles, we need to provide a size per platform/provider, not just a boolean.

mgorny: No, actually I was thinking wrong. The only reason for these "preprocessed" booleans is to know if there's at least one value for the OS, so we'd know whether to output the template or not.

mgorny: Okay, here's an idea: to avoid all the confusion, I'll stop putting the "preprocessed" stuff back and just do it in pure Jinja. The downside is that the is-any-value-in-loop-true logic will be done twice now: once in Python to determine whether to include .sh/.bat script, and then again in Jinja to determine whether to include the extra template part. However, all of this will be kept in a single file, so there will be no confusion over what exactly the value is.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
```diff
@@ -2,6 +2,14 @@
 # update the conda-forge.yml and/or the recipe/meta.yaml.
 # -*- mode: yaml -*-
 
+{%- set vars = namespace(store_build_artifacts=False) %}
+{%- for data in configs %}
+  {%- set pfarchless = data.build_platform.split('-')[0] %}
```
h-vetinari:
```suggestion
  {%- set os = data.build_platform.split('-')[0] %}
```

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
```diff
@@ -2,6 +2,14 @@
 # update the conda-forge.yml and/or the recipe/meta.yaml.
 # -*- mode: yaml -*-
 
+{%- set vars = namespace(store_build_artifacts=False) %}
+{%- for data in configs %}
+  {%- set pfarchless = data.build_platform.split('-')[0] %}
```
h-vetinari: ditto

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -2,6 +2,14 @@
 # update the conda-forge.yml and/or the recipe/meta.yaml.
 # -*- mode: yaml -*-
 
+{%- set vars = namespace(store_build_artifacts=False) %}
+{%- for data in configs %}
+  {%- set pfarchless = data.build_platform.split('-')[0] %}
```
h-vetinari: ditto

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -2,11 +2,14 @@
 # update the conda-forge.yml and/or the recipe/meta.yaml.
 # -*- mode: yaml -*-
 
-{%- set platformset = [] %}
+{%- set vars = namespace(platformset=[], store_build_artifacts=False) %}
 {%- for data in configs %}
   {%- set pfarchless = data.build_platform.split('-')[0] %}
```
h-vetinari: I realize now that you're copying from here. But this is a horrible name. `plat_archless` might be acceptable, but `os` is both shorter and less convoluted.

conda_smithy/configure_feedstock.py
```diff
@@ -1821,6 +1822,20 @@ def render_appveyor(jinja_env, forge_config, forge_dir, return_metadata=False):
 
 
 def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform):
+    platform_templates = {
+        "linux": [
```
h-vetinari, line 1826: I understand that this is just C&P'd from existing code. You can leave this as-is, though I really wish we'd have more consistent usage of `platform` / `arch` / `os` / `target` etc.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: May as well change it all over the place.

tests/test_utils.py
```diff
@@ -78,3 +80,82 @@ def test_get_feedstock_name_from_rattler_metadata_multiple_outputs(
...
+        ],
+        os="linux",
+    ) == [
+        ConditionalValue(
+            True,
+            platform=["linux_64", "win_64"],
+            provider=["azure"],
+        ),
+        ConditionalValue(True, provider=["github_actions"]),
+    ]
```
h-vetinari: Can you modify this to be
```python
    # explanation
    result = filter_conditional_values(...)
    expected = ConditionalValue(...)
    assert result == expected 
```

conda_smithy/configure_feedstock.py
```diff
@@ -2464,6 +2472,33 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
+        if "store_build_artifacts" in file_config.get("azure", {}):
+            raise ValueError(
+                "`store_build_artifacts` both in `workflow_settings` and `azure` "
+                "sections. Please remove the latter."
+            )
+        if "store_build_artifacts" in file_config.get("github_actions", {}):
+            raise ValueError(
+                "`store_build_artifacts` both in `workflow_settings` and "
+                "`github_actions` sections. Please remove the latter."
+            )
```
h-vetinari, line 2486: I see no test that covers this; also no tests for platform handling/duplication, or even that stuff like
```yaml
      workflow_settings:
        store_build_artifacts:
          - value: true
            provider: github_actions
            platform: osx-64          # AND-condition
         # OR-condition; linted to be non-overlapping
          - value: true
            platform: [win-64, linux-64]  # matched as OR
```
would work at all

conda_smithy/configure_feedstock.py
```diff
@@ -2464,6 +2472,33 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
+                "provider": "azure",
+                "value": config["azure"]["store_build_artifacts"],
+            }
+        )
+        config["workflow_settings"]["store_build_artifacts"].append(
+            {
+                "provider": "github_actions",
+                "value": config["github_actions"]["store_build_artifacts"],
+            }
+        )
```
h-vetinari, line 2500: Also not sure if this is covered by any tests?

h-vetinari (commented):
I didn't look too closely at `utils.py` yet, but the interface looks good; mainly I'm interested in seeing more tests.

tests/test_utils.py
mgorny: Done. I've assumed you meant just for the long final test.

conda_smithy/configure_feedstock.py
mgorny: Well, okay. Lemme figure out how to test this. I see that there are some `configure_feedstock` tests, so hopefully won't be that hard.

mgorny:
Ok, this will take some time but I've decided to go for proper test coverage. So far I'm thinking:

- [x] GHA plain true/false tests for old/new setting
- [x] Azure plain true/false tests for old/new setting
- [x] GHA/Azure test for error when both are defined
- [x] GHA/Azure test for more complex conditions (per-platform / per-os?)
- [x] GHA/Azure test for error when overlapping conditions are defined

mgorny:
@h-vetinari, broad set of tests added :-).

tests/test_utils.py
```diff
@@ -78,3 +80,85 @@ def test_get_feedstock_name_from_rattler_metadata_multiple_outputs(
...
+            {
+                "provider": "github_actions",
+                "value": True,
+            },
+            {
+                "os": "win",
+                "value": True,
+            },
+        ],
+        os="linux",
```
h-vetinari, line 156: Can you add an explanatory comment why these values are filtered out?

conda_smithy/schema.py
```diff
@@ -991,6 +1047,23 @@ class ConfigModel(BaseModel):
...
+            - provider: github_actions
+              platform: linux_aarch64
+              value: true
```
h-vetinari, line 1061: I don't see a test that covers the AND-condition between two keys (like `provider:` & `platform:` here).

h-vetinari (commented):
Basically LGTM, just one missing test and perhaps a bit of explaining

conda_smithy/schema.py
mgorny: Used `os` + `provider`, so it actually tests that `os` on another `provider` doesn't get enabled.

h-vetinari (approved):
Thanks for the multiple iterations on this! 🙏

h-vetinari:
FWIW, existing
```yaml
azure:
  store_build_artifacts: true
```
now causes linter failures rather than a hint

mgorny:
Interesting. It seems to be caused by `deprecated=True`. Reverting that makes the lint pass. The diff says:

```diff
diff --git a/conda_smithy/data/conda-forge.json b/conda_smithy/data/conda-forge.json
index 81108f5f..fb3b5afa 100644
--- a/conda_smithy/data/conda-forge.json
+++ b/conda_smithy/data/conda-forge.json
@@ -147,7 +147,6 @@
             }
           ],
           "default": false,
-          "deprecated": true,
           "description": "Deprecated. Store the conda build_artifacts directory as an Azure pipeline artifact. Use `workflow_settings.store_build_artifacts` instead.",
           "title": "Store Build Artifacts"
         },
```

So I guess the `"deprecated"` property in schema is sufficient to break lint. I'll look if we're calling the schema linter wrong or sth.

mgorny:
Okay, this seems to be a generic problem. I'll file a separate bug for it.

mgorny:
Filed #2537. Will try bisecting it a bit.

PR #2517, opened 2026-04-13
Skip two tests that fail on Apple Silicon Macs

cdeil:
Pair-coding with @timomuc .

cc @jaimergp

### DIFF
--- a/tests/test_variant_algebra.py
+++ b/tests/test_variant_algebra.py
@@ -1,3 +1,4 @@
+import platform
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


cdeil:
pre-commit.ci autofix

jaimergp:
Thanks, I confirmed this locally. Tests do pass if `CONDA_SUBDIR` is set for platforms like `linux-64`, so the test is just noise for folks on Apple Silicon anyway and regressions would be eventually caught on CI anyway.

PR #2518, opened 2026-04-13
[PyCon.de sprints] Fix various pylint finds

snordhausen:
Remove re-imports, unused variables, implicit string concatenation, overly complex assignment.

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/azure_ci_utils.py
+++ b/conda_smithy/azure_ci_utils.py
@@ -133,11 +133,7 @@ def get_default_build_definition(
 ):
     import inspect
 
-    from vsts.build.v4_1.models import (
-        BuildDefinition,
-        BuildRepository,
-    )
-    from vsts.task_agent.v4_0.task_agent_client import TaskAgentClient
+    from vsts.build.v4_1.models import BuildRepository
 
     aclient = TaskAgentClient(config.instance_base_url, config.credentials)
 
@@ -213,8 +209,6 @@ def get_default_build_definition(
 
 
 def register_repo(github_org, repo_name, config: AzureConfig = default_config):
-    from vsts.build.v4_1.models import BuildDefinitionReference
-
     bclient = build_client()
     build_definition = get_default_build_definition(
         github_org,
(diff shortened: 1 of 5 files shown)


conda_smithy/azure_ci_utils.py
```diff
@@ -133,11 +133,7 @@ def get_default_build_definition(
 ):
     import inspect
 
-    from vsts.build.v4_1.models import (
-        BuildDefinition,
-        BuildRepository,
-    )
-    from vsts.task_agent.v4_0.task_agent_client import TaskAgentClient
+    from vsts.build.v4_1.models import BuildRepository
```
jaimergp: Let's move this import to the top-level too.

jaimergp (commented):
Thanks! Just one little change about those imports. We are already importing a few things from `vsts` so no need to keep any inlined.

conda_smithy/azure_ci_utils.py
snordhausen: Done.

news/2518-pylint-finds.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Small code improvements as pointed out by pylint.
```
jaimergp:
```suggestion
* Small code improvements as pointed out by pylint. (#2518)
```

jaimergp (approved):

dbast:
Can the pylint rules which found those things be activated via the rules list at https://github.com/conda-forge/conda-smithy/blob/main/pyproject.toml#L46?

snordhausen:
@dbast I'll look into that and include the rules for #2520, too.

snordhausen:
@dbast 
For unused variables that are created as part of a loop, there`s `B007`. But that rule will require some additional changes, so I'll put it into a separate PR once this one is merged.

For the re-imports, pylint's `W0404 (reimported)` seems to have no ruff equivalent. There's only PLC0415 for imports not at the top level. But this is actually used a lot in `cli.py` (I assume to emulate lazy loading).

For the `.format()` calls that should have been f-strings in #2520, I checked with some other sprint participants and ruff seems to have no rule to catch those specific cases.

So in summary, ruff seems to be missing some rules to catch these specific issues that were detected by pylint.

PR #2520, opened 2026-04-13
[PyCon.de sprints] simplify string formatting with f-strings

snordhausen:
Checklist
* [x] Added a ``news`` entry


conda_smithy/feedstock_tokens.py
```diff
@@ -88,19 +88,14 @@ def generate_and_write_feedstock_token(user, project, provider=None):
     # spill tokens
     failed = False
     err_msg = None
+    provider = "provider" if provider is None else "provider " + provider
```
jaimergp: This will override `provider`, let's find a different way.

jaimergp:
```suggestion
    provider_str = "provider" if provider is None else "provider " + provider
```

PR #2522, opened 2026-04-13
Replace black with ruff format [pyconde sprints]

cdeil:
cc @jaimergp 

This PR changes from `black` to `ruff format`.

What's not included is the result of running this which will create a large diff with `dedent()` calls:
```bash
ruff format .
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,15 +6,12 @@ repos:
         exclude: (tests/recipes|conda_smithy.recipe)
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
--   repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.3.1
-    hooks:
-    -   id: black
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.15.9
     hooks:
-        - id: ruff
+        - id: ruff-check
           args: [ --fix ]
+        - id: ruff-format
 
 ci:
     autofix_commit_msg: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,13 +34,13 @@ include-package-data = true
 write_to = "conda_smithy/_version.py"
 write_to_template = "__version__ = '{version}'"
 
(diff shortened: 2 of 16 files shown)


pyproject.toml
```diff
@@ -34,13 +34,9 @@ include-package-data = true
 write_to = "conda_smithy/_version.py"
 write_to_template = "__version__ = '{version}'"
 
-[tool.black]
-# matches black's default value
-line-length = 88
```
jaimergp, line 39: I confirmed we don't need this because it's the default in `ruff` and it's unlikely they will change that.

.pre-commit-config.yaml
```diff
@@ -6,13 +6,10 @@ repos:
...
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
--   repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 26.3.1
-    hooks:
-    -   id: black
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.15.9
     hooks:
+        - id: ruff-format
```
jaimergp: According to the [docs](https://docs.astral.sh/ruff/integrations/#pre-commit), this needs to go after the linter:

> When running with --fix, Ruff's lint hook should be placed before Ruff's formatter hook, and before Black, isort, and other formatting tools, as Ruff's fix behavior can output code changes that require reformatting.

cdeil: Fixed. Put `ruff check --fix` before `ruff format` in `.pre-commit-config.yaml`

jaimergp:
pre-commit.ci autofix

jaimergp:
Thanks @cdeil! Merging this will probably introduce conflicts in a few PRs that are in progress, so the most likely outcome is that we merge this once we have dealt with them. We'll probably need to add [66e2c24](https://github.com/conda-forge/conda-smithy/pull/2522/commits/66e2c242910db2b9318e07523100a2c367716666) to the ignored revs file once merged (and we can't squash this one).

PR #2523, opened 2026-04-13
[PyCon DE Sprints] Bump minimum Python version to 3.11

NMertsch:
Implicitly, this is already the case due to the [dependency](https://github.com/conda-forge/conda-recipe-manager-feedstock/blob/5f21af50abf9a72f79b0b79e3bb89a9f9919520c/recipe/meta.yaml#L3) on 'conda-recipe-manager>=0.9'. Also:

* Remove dependency `tomli` (Python 3.11 includes `tomllib`)
* Remove dependency `backports.strenum` (Python 3.11 includes `enum.StrEnum`)
* Disable ruff rules `UP007` and `UP045` (deferred to another PR to keep this one short and focused)
    * [UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/): `typing.Union[A, B]` -> `A | B`
    * [UP045](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/): `typing.Optional[A]` -> `A | None`

Checklist
* [x] Added a ``news`` entry

### DIFF
--- a/conda_smithy/linter/utils.py
+++ b/conda_smithy/linter/utils.py
@@ -1,8 +1,8 @@
 import copy
 import os
 import re
-import sys
 import time
+import tomllib
 from collections.abc import Mapping, Sequence
 from functools import lru_cache
 from glob import glob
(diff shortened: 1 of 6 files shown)


beckermr (approved):

beckermr:
@conda-forge/core are we ok bumping smithy past our minimum supported python version in our pinnings? That one is python 3.10.

beckermr:
Note that conda-recipe-manager is an optional dep.

jaimergp:
> Note that conda-recipe-manager is an optional dep.

Yea, I don't mind skipping a few months of support (3.10 goes EOL in October). We have been testing all PRs with Python 3.11 for a while (since October 2024), so if it works with Python 3.10... we don't really know beyond what we are testing in the feedstock.

At the same time we may be able to offer better granularity for optional dependencies with the Pixi PR at #2525.

PR #2525, opened 2026-04-13
Add pixi [pyconde sprints]

cdeil:
At the PyConDE sprints most of us trying to work on `conda-smithy` were running into `conda` issues in regular terminals or VSCode terminals.

This PR adds support for using `pixi` to work on `conda-smithy` keeping `conda` support unchanged.

The large diff in `environment.yml` is a one-time format change due to order and formatting of the pixi export.

The CI check to ensure consistency of dependencies in `environment.yml` and `pyproject.toml` between conda and pixi users is following the example [HERE](https://github.com/conda-forge/conda-forge.github.io/blob/f9dd23d8e804818a2579211ca4d93d6d2c2116f2/.github/workflows/deploy.yml#L81-L97) suggested by @jaimergp


.gitignore
```diff
@@ -23,7 +23,10 @@ nppBackup/
 .mypy_cache/
 
 conda_smithy/_version.py
+
+# pixi environments
 .pixi/*
+!.pixi/config.toml
```
jaimergp, line 29:
```suggestion
!.pixi/config.toml
pixi.lock
```

Let's not commit the lockfile for now.

pixi.lock
jaimergp, line 1: Remove this one, please 🙏

.gitattributes
```diff
@@ -9,3 +9,5 @@
 *.yaml.ci-skel.tmpl linguist-language=yaml
 *.yml.tmpl linguist-language=yaml
 *.yaml.tmpl linguist-language=yaml
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true -diff
```
jaimergp: We can remove this because we won't commit it.
```suggestion
```

environment.yml
```diff
@@ -1,46 +1,42 @@
 name: conda-smithy
 channels:
-  - conda-forge
-  - nodefaults
+- conda-forge
+- nodefaults
 dependencies:
```
jaimergp, line 5: It's a pity we lost the comments during the export. Instead, let's make sure that they are mentioned in the (now source of truth) pyproject.toml table.

pyproject.toml
```diff
@@ -78,3 +78,52 @@ markers = [
...
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+name = "conda-smithy"
+platforms = ["linux-64", "linux-aarch64", "osx-arm64", "win-64"]
+
+[tool.pixi.tasks]
+install = { cmd = "pip install --no-build-isolation -e .", description = "Install conda-smithy in editable mode" }
+test = { cmd = "pytest tests --cov conda_smithy", description = "Run tests with coverage", depends-on = ["install"] }
+lint = { cmd = "pre-commit run --all-files", description = "Run linters (pre-commit)" }
```
jaimergp: Not sure if these add any benefit over simply running `pixi run pytest` or `pixi run pre-commit`.
```suggestion
```

pyproject.toml
```diff
@@ -78,3 +78,52 @@ markers = [
...
     "cli: CLI tests outside of test/test_cli.py",
 ]
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+name = "conda-smithy"
+platforms = ["linux-64", "linux-aarch64", "osx-arm64", "win-64"]
+
+[tool.pixi.tasks]
+install = { cmd = "pip install --no-build-isolation -e .", description = "Install conda-smithy in editable mode" }
```
jaimergp: This needs to be replaced by a section like this

```suggestion
[tool.pixi.pypi-dependencies]
"conda-smithy" = { path  = ".", editable = true }
```

pyproject.toml
```diff
@@ -78,3 +78,52 @@ markers = [
...
+license-expression = "*"
+libarchive = "*"
+cirun = ">=0.30"
+pydantic = ">=2.11,<3"
+pixi = ">=0.59.0"
+jsonschema = "*"
+"backports.strenum" = "*"
+exceptiongroup = "*"
+py-rattler = ">=0.22,<0.23"
+rattler-build-conda-compat = ">=1.4.5,<2.0.0a0"
```
jaimergp, line 117: It may be interesting to split this long list in different "features" (as per Pixi parlance), and then combine all of them in the default environment. 

See https://pixi.prefix.dev/latest/workspace/multi_environment/#feature-environment-set-definitions for more details

pyproject.toml
```diff
@@ -78,3 +78,52 @@ markers = [
...
+name = "conda-smithy"
+platforms = ["linux-64", "linux-aarch64", "osx-arm64", "win-64"]
+
+[tool.pixi.tasks]
+install = { cmd = "pip install --no-build-isolation -e .", description = "Install conda-smithy in editable mode" }
+test = { cmd = "pytest tests --cov conda_smithy", description = "Run tests with coverage", depends-on = ["install"] }
+lint = { cmd = "pre-commit run --all-files", description = "Run linters (pre-commit)" }
+
+[tool.pixi.dependencies]
+python = ">=3.9"
```
jaimergp: This is already enforced because this is `pyproject.toml` and `python-requires` is defined above.
```suggestion
```

pyproject.toml
```diff
@@ -78,3 +78,52 @@ markers = [
...
+pip = "*"
+python-build = "*"
+setuptools = ">=45"
+setuptools_scm = ">=8.1"
```
jaimergp: These are just to install the package, so I guess we don't need it; Pixi will handle that as part of the `pypi-dependencies` comment above.

pyproject.toml
```diff
@@ -78,3 +78,52 @@ markers = [
...
+tomli = ">=1.0.0"
+pre-commit = "*"
+mock = "*"
+pytest = "*"
+pytest-cov = "*"
+conda-recipe-manager = ">=0.9"
+conda-souschef = "*"
```
jaimergp: These are testing dependencies, perhaps better placed in their own feature.

pyproject.toml
jaimergp, line 117: These are the actual runtime dependencies, and the ones that deserve to be in the default dependency list (where they are already 👍 )

pyproject.toml
jaimergp: Also tomli below, which I missed in the selection.

isuruf (dismissed):
This requires keeping `environment.yml` and `pyproject.toml` to be in sync. Please add a test that checks that these two lists are identical so that we don't accidentally change one list and not the other.

cdeil:
@jaimergp - could you please do a second review? I did a follow-up commit to try and address your feedback.

cdeil:
> This requires keeping `environment.yml` and `pyproject.toml` to be in sync. Please add a test that checks that these two lists are identical so that we don't accidentally change one list and not the other.

@isuruf - is the existing check OK?

environment.yml
```diff
@@ -1,46 +1,40 @@
...
+- pixi >=0.59.0
+- jsonschema *
+- backports.strenum *
+- exceptiongroup *
+- py-rattler >=0.22,<0.23
+- rattler-build-conda-compat >=1.4.5,<2.0.0a0
+- python >=3.9
+- pip
+- pip:
+  - -e .
```
isuruf, line 40: Remove this please

.github/workflows/tests.yml
```diff
@@ -78,3 +78,20 @@ jobs:
...
+      - uses: actions/checkout@v6
+      - uses: prefix-dev/setup-pixi@v0.9.5
+        with:
+          run-install: false
+      - name: Check pyproject.toml and environment.yml are in sync
+        run: |
+          pixi workspace export conda-environment --name conda-smithy | sed '/^$/d' > environment.yml
+          if ! git diff --exit-code environment.yml; then
+            echo "error: pyproject.toml [tool.pixi] and environment.yml are out of sync"
+            echo "Run: pixi workspace export conda-environment --name conda-smithy | sed '/^\$/d' > environment.yml"
```
isuruf, line 95: Can you please post the diff here? There's no need for a developer to force pixi to figure out this change.

cdeil: The diff is currently empty because the two files are consistent.
This CI check guarantees they will remain in sync moving forward.

In the future if a dev adjusts a dependency they have to do it in `pyproject.toml` and `environment.yml` consistently, or they just do it in `pyproject.toml` and run the `pixi workspace export` command which applies the correct consistent change to `environment.yml`.

As far as I know there is no better way to do it. But seems OK, no?

@isuruf  Does this answer your question?
(Apologies if I misunderstood/don't get what you are looking for)

isuruf: I mean report the failing diff in the CI test, so that a PR author can apply the diff without needing to have pixi.

jaimergp: AFAIK the line `if ! git diff --exit-code environment.yml` is not redirecting output anywhere, so it'll be printed by default 🤔 See this example with a single line change in README.md:

```
❯ git status
On branch linter-revamp
Your branch is up to date with 'origin/linter-revamp'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   README.md

no changes added to commit (use "git add" and/or "git commit -a")
❯ if ! git diff --exit-code; then
∙   echo "There are changes"
∙ fi
diff --git i/README.md w/README.md
index f10ed6ba..08723d03 100644
--- i/README.md
+++ w/README.md
@@ -15,6 +15,8 @@ Overview
 [![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
 
+A change
+
 Installation
 ------------
 
There are changes
```

environment.yml
jaimergp: I think this is added by `pixi workspace export conda-environment` because of the `pypi-dependencies` table, and I see no way to disable that. We may need to edit it out by hand with `sed` or something.

@cdeil, if you are into Rust, adding a `--no-pip` flag to the subcommand in Pixi would be straightforward. This the Rust module: https://github.com/prefix-dev/pixi/blob/802105d11f030187916ee812e66572a388c76c3d/crates/pixi_cli/src/workspace/export/conda_environment.rs#L166-L170

.github/workflows/tests.yml
isuruf, line 95:
```suggestion
            echo "Please apply the above diff or run: pixi workspace export conda-environment --name conda-smithy | sed '/^\$/d' > environment.yml"
```

news/pixi-dev-support.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* Add Pixi dev support in `pyproject.toml`, keeping `conda` support as-is.
+  This is for devs working on `conda-smithy`.
```
jaimergp, line 4:
```suggestion
* Add Pixi-driven development support in `pyproject.toml`, keeping `conda` support as-is. (#2525)
```

environment.yml
```diff
@@ -1,46 +1,40 @@
...
+- libarchive *
+- cirun >=0.30
+- pydantic >=2.11,<3
+- pixi >=0.59.0
+- jsonschema *
+- backports.strenum *
+- exceptiongroup *
+- py-rattler >=0.22,<0.23
+- rattler-build-conda-compat >=1.4.5,<2.0.0a0
+- python >=3.9
```
jaimergp, line 37:
```suggestion
- rattler-build-conda-compat >=1.4.12,<2.0.0a0
- python >=3.11
```

As per recently merged PRs.

pyproject.toml
```diff
@@ -78,3 +78,53 @@ markers = [
...
+libarchive = "*"
+cirun = ">=0.30"
+pydantic = ">=2.11,<3"
+pixi = ">=0.59.0"
+jsonschema = "*"
+"backports.strenum" = "*"
+exceptiongroup = "*"
+# py-rattler's API subject to change, pin to minor
+py-rattler = ">=0.22,<0.23"
+rattler-build-conda-compat = ">=1.4.5,<2.0.0a0"
```
jaimergp, line 117:
```suggestion
rattler-build-conda-compat = ">=1.4.12,<2.0.0a0"
```

As per recently merged PRs.

environment.yml
baszalmstra: Just out of curiosity, why do you want to remove this? Isnt it just as nice that you dont have to run `pip install -e` anymore?

isuruf: I don't like tools doing sneaky things. It's a hack anyway that `-e .` works at all.

pavelzw: pixi-build might solve this here? [#general > Add a pixi-build section to conda-smithy @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Add.20a.20pixi-build.20section.20to.20conda-smithy/near/579794920)

baszalmstra:
@cdeil Are you planning on continuing with this?

cdeil:
> @cdeil Are you planning on continuing with this?

No. Would be great if someone from the team could finish this up.

baszalmstra:
@jaimergp Do you think instead of using a dynamic version, it would be ok to use rever to update the version in the pyproject.toml instead?

environment.yml
baszalmstra: I implemented the `--no-pypi` https://github.com/prefix-dev/pixi/pull/6380

jaimergp:
> Do you think instead of using a dynamic version, it would be ok to use rever to update the version in the pyproject.toml instead?

Can you elaborate? 🤔 The dynamic version of what?

baszalmstra:
The pyproject defines the version of the package as dynamic:

https://github.com/conda-forge/conda-smithy/blob/ef2133572859fa570e78eaf2029c486416f730c8/pyproject.toml#L11

jaimergp:
Ah, that's for the `scm` integrations. We can change that sure if we want (or other parts of the release process), but why here? I'm not following the connection of Pixi <> scm versioning. Any documented gotchas I'm not aware? 🤔

baszalmstra:
Ah sorry I completely missed adding the context 😅 

Im asking because I was trying to turn conda-smithy into (optionally) a pixi build package. That makes it easy to (amongst others) use it directly as a git dependency.

However, currently pixi doesnt support `dynamic = ["version"]`.

jaimergp:
I see. I guess we can discuss it, but it'll have to be separate from this PR.

PR #2528, opened 2026-04-15
Use `CONDA_BLD_PATH` consistently in templates

mgorny:
Checklist
* [x] Added a ``news`` entry
* [ ] Regenerated schema JSON if schema altered (`python -m conda_smithy.schema`)

Replace the `CONDA_BLD_DIR` directory used when storing build artifacts with `CONDA_BLD_PATH` that is used when building, to ensure consistency and avoid unnecessary confusion.

PR #2529, opened 2026-04-15
Setting build directories via `workflow_settings`

### LINKED ISSUE #2527: MINIFORGE_HOME incorrectly set for Blacksmith runners
The [onnxruntime-feedstock](https://github.com/conda-forge/onnxruntime-feedstock) has the following section in `conda-forge.yaml`:
```yaml
azure:
  ...
  settings_win:
    variables:
      CONDA_BLD_PATH: C:\bld\
      MINIFORGE_HOME: C:\Miniforge
```
I'm not entirely sure how these variables are (or were, given that they are targeting "azure") supposed to be used, but `github-actions.yml.tmpl` at least does not appear to allow configuration of this kind when rendering the `conda-build.yml` workflow. The rendered Windows job looks as follows:

```yaml
...
    - name: Build on windows
      ...
      env:
        # default value; make it explicit, as it needs to match with artefact
        # generation below. Not configurable for now, can be revisited later
        CONDA_BLD_DIR: C:\bld
        MINIFORGE_HOME: ${{ contains(runner.arch, 'ARM') && 'C' || 'D' }}:\Miniforge
```

At least on the Blacksmith CI [runners](https://github.com/conda-forge/admin-requests?tab=readme-ov-file#larger-runners-for-github-actions), it seems to be causing the build to fail during setup of the build environment ([cf](https://github.com/conda-forge/onnxruntime-feedstock/actions/runs/24410961614/job/71307773850)). I assume that may be due to `D:\` not being available on those runners? Manually editing the file to point to `C:\` "solved" the immediate issue and the build environment was [set up correctly](https://github.com/conda-forge/onnxruntime-feedstock/actions/runs/24413346048/job/71316572899?pr=177). 

I am unsure if the correct fix for this is on the conda-smithy side or with the Blacksmith runners. 

I suppose this issue may also be relevant: https://github.com/conda-forge/conda-smithy/issues/2349

mgorny:
Checklist
* [x] Added a ``news`` entry

Per the discussion in #2349, adds two `workflow_settings`:
- `tools_install_dir` that replaces `MINIFORGE_HOME`
- `build_workspace_dir` that replaces `CONDA_BLD_PATH`

The defaults are mostly preserved, except that `D:\` paths are only used for Azure and hosted GHA win-64 runners, fixing #2527.

Includes a fix for Linux Docker builds to respect `build_workspace_dir` (but not `tools_install_dir`).


beckermr:
I suspect the linux issue is because this variable is also set in CI support?

See this line here: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/main/recipe/run_conda_forge_build_setup_linux#L16

conda_smithy/utils.py
```diff
@@ -352,4 +352,26 @@ def get_workflow_settings(
...
+        workflow_settings["tools_install_dir"] = {
+            "win-32": r"D:\Miniforge",
+            "win-64": r"D:\Miniforge",
+            "win-arm64": r"C:\Miniforge",
+        }.get(platform, "~/miniforge3")
+    if workflow_settings.get("build_workspace_dir") is None:
+        workflow_settings["build_workspace_dir"] = {
+            "linux": "build_artifacts",
+            "osx": f"{workflow_settings['tools_install_dir']}/conda-bld",
+            "win": r"C:\bld",
```
h-vetinari, line 376: not a criticism of this PR, but there's a very annoying footgun in the ci-setup for swapfiles on windows [where](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/29b3d39d4c21cd96c6274231f295bacd5c860611/recipe/run_conda_forge_build_setup_win.bat#L32-L42) the setup is sensitive to the spelling w.r.t. the otherwise redundant escaping of backslashes. In other words,
```
CONDA_BLD_PATH="C:\\bld\\"  # works
CONDA_BLD_PATH="C:\\bld"    # doesn't work
CONDA_BLD_PATH="C:\bld\"    # doesn't work
CONDA_BLD_PATH="C:\bld"     # doesn't work
```

I think it would be good to normalize the spelling in smithy (e.g. remove duplicate backslashes, decide on a style with or without a trailing backslash), and then we only need to handle that single case in the ci-setup.

h-vetinari (commented):
I don't have a clear idea whether this is easily possible or worth the hassle, but I was thinking that it'd be nice to introduce something like `unix_xor_win: {True, False}` (using the established "xor"="exclusive or") for values that _cannot_ be shared between unix and windows, like the paths here. If active, that should not prohibit fine-grained definition of values under `workspace_settings`, only supposed "universal" values where that's simply not possible.

mgorny:
> I suspect the linux issue is because this variable is also set in CI support?

It's more because we're using Docker, so even if the paths were passed into the container (I haven't checked if they are), they would be interpreted inside the container rather than on the host system. However, I think the easy way to resolve this is to mount them as volumes (so "translating" the specified host paths into fixed in-container paths).

conda_smithy/utils.py
mgorny: Sorry, could you clarify what you're asking me to do? Are you saying we should normalize user-provided paths here?

mgorny:
I've added Azure part now. Few notes (I'll need to write them down somewhere later):
- Azure does not need the `~` substitution — it handles that just fine
- I needed to `pop()` the old variables from `variables` dict, as they would interfere with the overrides
- I've just noticed that #2528 broke `azure.settings_win.variables.CONDA_BLD_DIR`; however, this PR fixes it (by converting to the new format)

I'll probably handle Linux/docker and tests for this next week.

conda_smithy/utils.py
h-vetinari: > Are you saying we should normalize user-provided paths here?

Yeah, that would be my preference; can be a separate PR if necessary.

h-vetinari:
> it'd be nice to introduce something like `unix_xor_win: {True, False}`

any comment on this @mgorny?

mgorny:
> > it'd be nice to introduce something like `unix_xor_win: {True, False}`
> 
> any comment on this @mgorny?

As in explicit error out if you specify it both for Unix and Windows? Are we talking about "effective" values (i.e. fail only if you end up using the same path in actual Unix and Windows workflows), or always (i.e. even if other criteria effectively make the error not possible; say, when you set `provider: azure` but not `os: win`, and Azure is only used for Windows)?

h-vetinari:
> As in explicit error out if you specify it both for Unix and Windows?

That's the idea. I'd like to prohibit
```yaml
workflow_settings:
  tools_install_dir:
    value: foo
```
i.e. those without a restriction to either `platform: win` or unix. We could also consider naming `win_xor_unix` as `requires_platform` or somesuch (and either way, note that this would be smithy-internal, not part of the feedstock config!). The idea is that classes of settings like `tools_install_dir` (and other with similar constraints in the future) can be specified to require a distinction between unix/win.

> Are we talking about "effective" values [...], or always [...]?

Obviously "effective" would be nicer from a UX POV (there's also the case to consider that a feedstock doesn't build for windows at all, whence one could consider re-allowing "universal" values for the paths), but all in all, this whole thing is more of a nice-to-have footgun avoidance, rather than a critical problem, so I think even an incomplete solution (or indeed none at all) would be fine.

conda_smithy/utils.py
```diff
@@ -352,4 +352,26 @@ def get_workflow_settings(
...
+    """
+    Fill the missing entries from `workflow_settings` with defaults for
+    the given provider-platform combination.
+    """
+    if workflow_settings.get("tools_install_dir") is None:
+        workflow_settings["tools_install_dir"] = {
+            "win-32": r"D:\Miniforge",
+            "win-64": r"D:\Miniforge",
+            "win-arm64": r"C:\Miniforge",
+        }.get(platform, "~/miniforge3")
```
jaimergp, line 371: We are going to need to special case when we use `D:`, since it's only on Github-hosted `win64` runners where it makes sense. Other providers don't use it (see https://github.com/conda-forge/conda-smithy/issues/2527)

jaimergp: So I say we always return `C:\Miniforge`, but then override in the provider-specific functions when we have `windows-latest` on GHA/Azure.

conda_smithy/utils.py
jaimergp, line 376: The drive of the bld directory should match the installation path too so there are no cross-filesystem overheads (hardlinks aren't supported across FS boundaries) when copying from the pkgs cache (which is embedded in the installation root).

conda_smithy/schema.py
```diff
@@ -509,6 +508,24 @@ class WorkflowSettings(BaseModel):
         """),
     )
 
+    tools_install_dir: Optional[
+        Union[str, list[conditional_value(str, None)], Nullable]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Directory to store build-time tools in.
```
jaimergp, line 516:
```suggestion
        Directory to install build-time tools in.
```

conda_smithy/schema.py
```diff
@@ -204,9 +202,10 @@ class AzureConfig(BaseModel):
 
             Finally, under `variables`, some important things you can set are:
 
-            - `CONDA_BLD_PATH`: Location of the conda-build workspace. Defaults to `D:\\bld`
-            - `MINIFORGE_HOME`: Location of the base environment installation. Defaults to
-              `D:\\Miniforge`.
+            - `CONDA_BLD_PATH`: Location of the conda-build workspace. Deprecated, use
+              `workflow_settings.build_workspace_dir` instead.
+            - `MINIFORGE_HOME`: Location of the base environment installation. Deprecated,
```
jaimergp, line 207:
```suggestion
            - ~~`CONDA_BLD_PATH`~~: Location of the conda-build workspace. Deprecated, use
              `workflow_settings.build_workspace_dir` instead.
            - ~~`MINIFORGE_HOME`~~: Location of the base environment installation. Deprecated,
```

For a better visual strikethrough?

conda_smithy/utils.py
mgorny: Should I adjust the default to `C:\\bld\\` then, in the meantime?

mgorny: Sorry, could you clarify what are you suggesting? Merely adjusting the defaults to match, or perhaps having default `build_workspace_dir` use the same drive as user-specified `tools_install_dir`?

conda_smithy/utils.py
mgorny: Ah, I thought the plan was to override it for Blacksmith in `conda-forge.yml`. So we should basically have the default dependent on `runs_on` value, is that correct?

mgorny:
Related question: do we want to adjust pixi to respect `tools_install_dir` / `MINIFORGE_HOME`?

conda_smithy/utils.py
jaimergp: The reason we are using `D:` is because in Github Actions and Azure, for `win-64`, this drive is significantly faster (although smaller too, so some heavy feedstocks go back to `C:`)

Considering that GHA/Azure's `win-arm64` does _not_ have `D:`, nor do Blacksmith and Namespace, I think it's fair to assume the `windows-latest` was the special case all along, not the default. 

So what I'm suggesting is that we think of `C:\xxx` as the default, but add a special case in the rerender pipeline to use `D:\xxx` when we are using Microsoft-hosted `win-64` runners.

conda_smithy/utils.py
jaimergp: At the very least the defaults should match. The 2nd suggestion is also nice but not required (we can document the filesystem boundaries limitations for hardlinks/copies and that should be ok).

jaimergp:
> Related question: do we want to adjust pixi to respect tools_install_dir / MINIFORGE_HOME?

I don't recall the details, but:

- I recall _trying_ because I wanted the local development workflow to be the same for conda, micromamba and pixi.
- However I also recall that there was no easy way to tell Pixi to install an environment to an arbitrary path, so maybe I gave up?
- These days we could also add @pavelzw's https://github.com/pavelzw/pixi-install-to-prefix to put the environment wherever.

mgorny:
Okay, I don't think overriding miniforge install dir for Linux is going to work, for the simple reason that it's already part of the Docker image. However, I suppose we might be able to override the directory where Docker stores data.

conda_smithy/utils.py
mgorny: By "default", I actually meant "user value should override it", but I get the point. Just thinking how to integrate it without it becoming awfully hacky.

jaimergp:
> Okay, I don't think overriding miniforge install dir for Linux is going to work, for the simple reason that it's already part of the Docker image. 

Makes sense

> However, I suppose we might be able to override the directory where Docker stores data.

You mean `build_artifacts`?

conda_smithy/utils.py
mgorny: Ok, tested against `SET_PAGEFILE` and indeed changed the default.

That said, if I were to normalize stuff, I'd rather go for normalizing on `C:\bld` as the canonical path. Do you think `\\` is actually safer or something?

mgorny:
> You mean `build_artifacts`?

No, `build_artifacts` I've covered already. What I'm thinking is what we could have Docker write the whole image into `tools_install_dir`, effectively placing miniforge and all the other bits there.

mgorny:
Okay, I think I've addressed all the concerns raised. Could you give it another look before I start adding tests?

mgorny:
(Note that I've went for "minimal effort" solution where I didn't have a better idea. Please treat that as a starting point for better ideas.)

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -79,12 +79,12 @@ jobs:
...
     - script: |
-        set MINIFORGE_HOME=$(MINIFORGE_HOME)
+        set MINIFORGE_HOME=$(tools_install_dir)
+        set CONDA_BLD_PATH=$(build_workspace_dir)
         set CI=azure
         set CI_RUN_ID=$(build.BuildNumber).$(system.JobAttempt)
         set FEEDSTOCK_NAME=$(build.Repository.Name)
         set ARTIFACT_STAGING_DIR=$(Build.ArtifactStagingDirectory)
-        set CONDA_BLD_PATH=$(CONDA_BLD_PATH)
         set BLD_ARTIFACT_PREFIX=conda_artifacts
```
jaimergp, line 88: `BLD_ARTIFACT_PREFIX` is a "prefix" in the filename sense (`./{prefix}somethingelse.txt`), not in the directory sense (`{prefix}/more/dirs`), right?

conda_smithy/utils.py
```diff
@@ -352,4 +353,53 @@ def get_workflow_settings(
                 f"platform={platform}: {filtered[0]} vs. {filtered[1]}"
             )
         data[setting_key] = filtered[-1].value if filtered else None
+
+    for path_var in ("tools_install_dir", "build_workspace_dir"):
+        if data[path_var] is None:
+            continue
+        win_path = PureWindowsPath(data[path_var])
+        print((os, win_path, win_path.drive))
```
jaimergp: Debug leftover?

conda_smithy/utils.py
```diff
@@ -352,4 +353,53 @@ def get_workflow_settings(
...
+    provider: str,
+    platform: str,
+    gha_runs_on: list[str],
+) -> None:
+    """
+    Fill the missing entries from `workflow_settings` with defaults for
+    the given provider-platform combination.
+    """
+    os = platform.split("-", 1)[0]
+    if workflow_settings.get("tools_install_dir") is None:
```
jaimergp, line 394: Do we need to account for `""` too?

conda_smithy/utils.py
```diff
@@ -352,4 +353,53 @@ def get_workflow_settings(
...
+        win_default = (
+            r"D:\Miniforge" if "windows-latest" in gha_runs_on else r"C:\Miniforge"
+        )
+        workflow_settings["tools_install_dir"] = (
+            win_default if os == "win" else "~/miniforge3"
```
jaimergp: Were we using `Miniforge3` before? I think the title-cased variant is the default in the installer 🤔

conda_smithy/utils.py
```diff
@@ -352,4 +353,53 @@ def get_workflow_settings(
...
+            )
+
     return data
+
+
+def fill_workflow_settings_defaults(
+    workflow_settings: dict[str, Any],
+    provider: str,
+    platform: str,
+    gha_runs_on: list[str],
```
jaimergp: Why is this `gha` only? The `C:` vs `D:` distinction also applies to Azure, IIRC.

conda_smithy/utils.py
```diff
@@ -352,4 +353,53 @@ def get_workflow_settings(
...
+    if workflow_settings.get("build_workspace_dir") is None:
+        tools_drive = (
+            PureWindowsPath(workflow_settings["tools_install_dir"]).drive
+            if os == "win"
+            else ""
+        )
+        workflow_settings["build_workspace_dir"] = {
+            "linux": "build_artifacts",
+            "osx": f"{workflow_settings['tools_install_dir']}/conda-bld",
+            "win": rf"{tools_drive}\\bld\\",
```
jaimergp, line 410:
```suggestion
            "win": rf"{tools_drive}\bld\",
```

If it's a raw string, do you need double backslashes, or is this because it will go through several layers before reaching the target script? (Python -> YAML -> Batch?)

conda_smithy/utils.py
h-vetinari: I have no strong preference for how we spell it, just that we normalize on one of the variants in smithy. I think ultimately the double backslashes come from being perceived as marginally "safer" in the sense that they are more universal

| | language requires escaping `\` | language does not require escaping `\` |
|-|--------|--------|
| `C:\\bld` | works ✅ | works ✅ |
| `C:\bld` | fails ❌ | works ✅ |

Python is obviously one of the many places where unescaped backslashes in strings (that aren't explicitly treated as raw) causes issues. That said, if we normalize here and ensure we have a clean ci-setup and no other places where this goes wrong, I'd slightly prefer the simpler `C:\bld`

conda_smithy/templates/azure-pipelines-win.yml.tmpl
mgorny: Yes.

conda_smithy/utils.py
mgorny: That would only happen if you explicitly passed:

```
workflow_settings:
  tools_install_dir: ""
```

conda_smithy/utils.py
mgorny: https://github.com/conda-forge/conda-smithy/blob/3d0a31f2003f45130420a3174846cab256b84c80/conda_smithy/templates/run_osx_build.sh.tmpl#L9

It's the value on osx. On Linux, we're using Docker with `/opt/conda` instead.

conda_smithy/utils.py
mgorny: See https://github.com/conda-forge/conda-smithy/pull/2529#discussion_r3090063136. Long story short, this is what `conda-forge-ci-setup` expects right now. Long term, I'd love to change it, but we need to adjust `conda-forge-ci-setup` first.

h-vetinari: @jaimergp, see this review [thread](https://github.com/conda-forge/conda-smithy/pull/2529#discussion_r3090063136)

conda_smithy/utils.py
mgorny: Yeah, that would be my preference too. That said, I don't really have any clue how likely it is for this to end up being unescaped somewhere. We could try changing it, and revert if people report issues. After all, it would only change things after a rerender.

We could also keep the default as-is for now, and try changing it locally in a few feedstocks first. Or perhaps have a mixed defaults, `D:\\bld\\` and `C:\bld`, so that feedstocks switching to Blacksmith would start testing the new value.

conda_smithy/utils.py
mgorny: Right, will fix that.

mgorny:
Okay, so I guess we'd want tests that:

- [x] correct defaults are applied to GHA default runners, custom GHA runners and Azure
- [x] path overrides are applied (including drive inference on Windows)
- [x] errors are thrown in error cases

conda_smithy/utils.py
jaimergp: Let's add a comment then, with a link if necessary.

conda_smithy/utils.py
h-vetinari: ci-setup only does a straightforward string equality comparison, so IMO we don't need so much testing here (in any case I'd like to avoid diverging defaults across drives).

For the purposes of transitioning more easily, we could complicate the if-branches in ci-setup to also check the single-backslash variant. Or we need to sync the eventual smithy release containing these changes with the corresponding update to the ci-setup

conda_smithy/configure_feedstock.py
```diff
@@ -1886,11 +1887,16 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
-                forge_config["workflow_settings"], "github_actions", data["platform"]
-            )
+        workflow_settings = get_workflow_settings(
+            forge_config["workflow_settings"], "github_actions", data["platform"]
+        )
+        fill_workflow_settings_defaults(
+            workflow_settings,
+            "github_actions",
+            data["platform"],
+            "D:" if "windows-latest" in data["gha_runs_on"] else "C:",
```
jaimergp:
```suggestion
            "D:" if {"windows-latest", "windows-2022", "windows-2025"}.intersection(data["gha_runs_on"]) else "C:",
```

I think? The point is that [there are more GH-hosted x64 Windows runners](https://github.com/actions/runner-images/blob/main/README.md#available-images).

mgorny:
Okay, I think I've addressed all the major points and added tests. I think we'll update the Windows defaults in a followup, after updating `conda-forge-ci-setup`.

news/2529-path-settings.rst
```diff
@@ -0,0 +1,26 @@
+**Added:**
+
+* Add ``workflow_settings.tools_install_dir`` and
+  ``workflow_settings.build_workspace_dir`` settings to ``conda-forge.yml`` that
+  provide per-platform and per-provider ability to override the paths used by
+  workflows. (#2500)
```
h-vetinari:
```suggestion
  workflows. (#2529)
```

conda_smithy/configure_feedstock.py
```diff
@@ -1886,11 +1887,22 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
+        fill_workflow_settings_defaults(
+            workflow_settings,
+            "github_actions",
+            data["platform"],
+            (
+                "D:"
+                if {"windows-latest", "windows-2022", "windows-2025"}.intersection(
+                    data["gha_runs_on"]
+                )
+                else "C:"
```
h-vetinari: I dislike this formatting quite a bit. Could we introduce a variable à la
```python
        on_gha = {"windows-latest", "windows-2022", "windows-2025"} & data["gha_runs_on"]
        # ...
        "D:" if on_gha else "C:"
```
?

h-vetinari (commented):
basically LGTM, just some nits; we should also consider the order between this PR and #2408 (since there are a bunch of new warnings here)

mgorny:
Done.

h-vetinari (approved):
This LGTM, thanks a lot

h-vetinari:
Now that https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/417 is in, I think we can merge this. Any last comments @jaimergp?

mgorny:
> Now that [conda-forge/conda-forge-ci-setup-feedstock#417](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/417) is in, I think we can merge this. Any last comments @jaimergp?

Now that it's in, I'll update the defaults, actually :-). And maybe add normalization, or that can be done in a subsequent PR.

mgorny:
The conda-forge-ci-setup change caused a regression, so let's stay with old paths for the time being.

jaimergp (approved):

conda_smithy/configure_feedstock.py
```diff
@@ -2499,6 +2511,46 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
 
+    if "MINIFORGE_HOME" in file_config.get("azure", {}).get("settings_win", {}).get(
+        "variables", {}
+    ):
+        if "tools_install_dir" in file_config.get("workflow_settings", {}):
+            raise ValueError(
+                "`workflow_settings.tools_install_dir` and "
+                "`azure.settings_win.variables.MINIFORGE_HOME` are mutually exclusive. "
+                "Please remove the latter."
+            )
```
jaimergp: We need to consider the interactions of these exceptions and the regro automation. This will break rerendering if a feedstock has both feedstocks. @beckermr do you recall if we have a rule like "rerendering must never raise an exception intentionally" or similar? Should this be a hint? We can choose to respect the modern version in `workflow_settings` and ignore the other for the purposes of rerendering, after all.

jaimergp: If we choose to do so, then #2408 should go first.

mgorny:
Added `D:` for namespace.

conda_smithy/configure_feedstock.py
beckermr: Rerendering can fail, but it should be for a very good reason. I do NOT think it failing because we changing how the conda-forge.yml is formatted is a good idea. We need to ensure things are backwards compatible. 

This same comment applies to the linter failures we see with the new store_build_artifacts setting.

mgorny:
Hmm, something wrong is happening with Linux here (but not in my limited testing):

https://github.com/conda-forge/pytorch-cpu-feedstock/actions/runs/24895021962/job/72897325574?pr=500

```
PermissionError: [Errno 13] Permission denied: '/home/conda/feedstock_root/build_artifacts/work'
```

The docker command there is:

```
  + docker run -v /home/runner/_work/pytorch-cpu-feedstock/pytorch-cpu-feedstock/recipe:/home/conda/recipe_root:rw,z,delegated -v /home/runner/_work/pytorch-cpu-feedstock/pytorch-cpu-feedstock:/home/conda/feedstock_root:rw,z,delegated -v build_artifacts:/home/conda/feedstock_root/build_artifacts:rw,z,delegated -e CONFIG -e HOST_USER_ID -e UPLOAD_PACKAGES -e IS_PR_BUILD -e GIT_BRANCH -e UPLOAD_ON_BRANCH -e CI -e FEEDSTOCK_NAME -e CPU_COUNT -e BUILD_WITH_CONDA_DEBUG -e BUILD_OUTPUT_ID -e flow_run_id -e remote_url -e sha -e BINSTAR_TOKEN -e FEEDSTOCK_TOKEN -e STAGING_BINSTAR_TOKEN quay.io/condaforge/linux-anvil-x86_64:alma9 bash /home/conda/feedstock_root/.scripts/build_steps.sh
```

I'll revert the Docker changes for now, since they're not strictly required here. We can look into making it respect custom paths separately later.

conda_smithy/configure_feedstock.py
beckermr: bump @mgorny @jaimergp for viz

mgorny: For the record, this is not a backwards compatibility problem. It will fail only if you explicitly add the new setting without removing the old setting.

beckermr: imho it should be a hint, not a failure.

mgorny: Well, I'm fine with changing it.

mgorny: Uh, actually, it's going to be hard.

If you do:

```yaml
azure:
  settings_win:
    variables:
      MINIFORGE_HOME: ...
workflow_settings:
  tools_install_dir:
    - os: [win]
      value: ...
```

It's going to fail anyway because of the overlap of values on Windows (which was explicitly requested). We could aim for "old value is ignored if new one is present", but since scopes are different, it's going to be a mess because e.g. `os: [win]` value applies both to Azure and GHA, while the Azure setting applies only to Azure.

jaimergp: If the `workflow_settings` option has precedence, it shouldn't matter what `azure` is set to. Even if it's imperfect; we can be blunt. "If `workflow_settings.tools_install_dir` is set, `azure.settings_*.variables.MINIFORGE_HOME` has no effect". Then hint against it, and folks will fix it.

mgorny: Ah, good idea! Done that now for the path variables. I'll update `store_build_artifacts` in a separate PR later, once this one is settled.

Should I also comment out `deprecated=True` from these vars until #2537 is fixed?

jaimergp: I don't think we ever had per-variable schemas there, so there's no `deprecated=True` to remove. The current docstring seems to be enough.

jaimergp: [d013ff1](https://github.com/conda-forge/conda-smithy/pull/2529/commits/d013ff15319d3f279a463eeda06a4d33fa31b2a5) is amazing, thanks.

jaimergp:
Cool, so with [d013ff1](https://github.com/conda-forge/conda-smithy/pull/2529/commits/d013ff15319d3f279a463eeda06a4d33fa31b2a5) we let the linter continue. We need to add the hints too, but we need https://github.com/conda-forge/conda-smithy/pull/2408 to do so. I suggest we merge here, and add the hints for the directories and the artifacts in a follow up PR.

PR #2533, opened 2026-04-22
fix: do not ignore .recipe_maintainers.json file in .gitignore

beckermr:
Checklist
* [x] Added a ``news`` entry

As it says at the top.

### DIFF
--- a/conda_smithy/feedstock_content/.gitignore
+++ b/conda_smithy/feedstock_content/.gitignore
@@ -6,6 +6,7 @@
 # Ignore all files and folders in root
 *
 !/conda-forge.yml
+!.recipe_maintainers.json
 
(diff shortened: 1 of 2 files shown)


beckermr:
@jaimergp @conda-forge/core this one is ready for review!

h-vetinari:
Why would we want to ignore these files? Aren't we relying on that mapping (e.g. user: ID) staying constant per feedstock?

If we do want to do this, it will not change the fact that these files have already been committed to a large number of feedstocks, and adding a matching `.gitignore` rule will not remove those files (rather, you'd have to then also commit a deletion of that file)

beckermr:
This PR ensures git does NOT ignore the files. See the diff

```
!.recipe_maintainers.json
```

h-vetinari (approved):
I had misunderstood the previous title, apologies.

PR #2538, opened 2026-04-27
Convert duplicate `store_build_artifacts` error to warning, and use `logger.warning()` to warn

mgorny:
Checklist
* [x] Added a ``news`` entry

- convert the error for duplicate `store_build_artifacts` error into a warning that the old value will be ignored
- convert the `DeprecationWarning`s from configuring feedstock into logging warnings, for more consistent display


news/2538-store_build_artifacts-warning.rst
```diff
@@ -0,0 +1,25 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Old ``{provider}.store_build_artifacts`` is now ignored with a warning if
```
jaimergp: You can simply add the PR number to the other news item because this is really a small detail in the overall feature that doesn't deserve its own item in the changelog.

h-vetinari (commented):
I'm not in favour of using logging for signalling, as none of that is visible to the regular `@bot, please rerender` workflow; it'd only be visible for people rerendering locally _and_ keeping an eye out for log messages. IMO we shouldn't fail the linter on this duplication, but raise a "hint"

beckermr:
The plan is to add a lint after the refactor in #2408 is done.

h-vetinari:
> The plan is to add a lint after the refactor in #2408 is done.

That's fine of course, by why introduce logging that we intend to remove again immediately after? Seems like pointless churn. 🤷

beckermr:
I don't have an opinion on the logging.

mgorny:
I don't have a really strong argument. My thinking is that logging a warning is less confusing than the setting being silently ignored, though obviously most of the people won't ever notice the warning in midst of all the output.

news/2538-store_build_artifacts-warning.rst
mgorny: Done.

jaimergp:
I just merged https://github.com/conda-forge/conda-smithy/pull/2408, so we can proceed with the lints. The logging is fine for local development workflows; it can stay imo.

h-vetinari:
As long as this is producing hints in `conda-smithy lint --conda-forge`, I have nothing against extra logging messages in the regular output during rerendering

tests/test_configure_feedstock.py
```diff
@@ -2445,10 +2459,23 @@ def test_store_build_artifacts_gha(py_recipe, jinja_env, path: str, value: bool)
...
+        config = configure_feedstock._load_forge_config(
+            forge_dir, "recipe/default_config.yaml"
+        )
+        assert (
+            any(
+                "`github_actions.store_build_artifacts` is ignored" in record.message
+                for record in caplog.records
+            )
+            == add_old
+        )
```
h-vetinari: comparing to a boolean is a bit weird. I'd prefer the (IMO) more straight-forward:
```suggestion
        if add_old:
            assert any(
                "`github_actions.store_build_artifacts` is ignored" in record.message
                for record in caplog.records
            )
```

tests/test_configure_feedstock.py
```diff
@@ -2499,10 +2531,24 @@ def test_store_build_artifacts_azure(py_recipe, jinja_env, path: str, value: boo
...
+        config = configure_feedstock._load_forge_config(
+            forge_dir, "recipe/default_config.yaml"
+        )
+        assert (
+            any(
+                "`azure.store_build_artifacts` is ignored" in record.message
+                for record in caplog.records
+            )
+            == add_old
+        )
```
h-vetinari: same here

tests/test_configure_feedstock.py
h-vetinari: I realize that this doesn't cover the absence of warnings in the `add_old is False` case; but we don't usually do absence-of-warnings tests.

mgorny: Very well.

jaimergp:
Can we add the hints here too, or would you rather cover those separately, @mgorny?

mgorny:
> Can we add the hints here too, or would you rather cover those separately, @mgorny?

That's basically the question posed by https://github.com/conda-forge/conda-smithy/pull/2540 : do we want special hints for this, or do we just assume that "deprecated field" hints are sufficient?

beckermr:
Deprecated field hints should be sufficient as long as they tell the user what to use instead.

mgorny:
With #2540 on top, it yields:

```
/home/mgorny/git/conda/libgit2-feedstock has some suggestions:
  'Store Build Artifacts' is deprecated.
Deprecated. Store the conda build_artifacts directory as an Azure pipeline artifact. Use `workflow_settings.store_build_artifacts` instead.
```

We could look into improving the exact message in #2540.

beckermr (approved):

PR #2546, opened 2026-05-05
Hint when dependency specifiers override pins

mgorny:
Checklist
* [ ] Added a ``news`` entry

Emit a hint when a dependency specifier in the recipe overrides a pinned dependency version.  This seems to be a common pitfall, for example when you copy dependency specifiers from upstream without realizing that a dependency such as:

    host:
    - libhwloc >=2.5

will actually cause the pinned version to be ignored, and force the newest `libhwloc`, leading to conflicting dependency constraints across packages.

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -97,6 +97,18 @@
       "path": ".github/workflows/*.y*ml"
     },
     {
+      "name": "PinnedDependencyOverridden",
+      "identifier": "CF-006",
+      "category": "CF",
+      "kind": "hint",
+      "added_in": "<3.56",
+      "deprecated_in": "",
+      "documentation": "Hint when dependency specification overrides a pin.",
+      "message": "${output} output overrides versions pinned in the feedstock:\n${bad_specs_list}\nRequirement spec should not list version specifiers to respect conda-forge-pinning. If you need to force another version, please override the pin via `conda_build_config.yaml`.",
+      "examples": [],
+      "path": ""
+    },
+    {
       "name": "NoDuplicateKeys",
       "identifier": "FC-001",
       "category": "FC",
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -27,6 +27,7 @@ from conda_smithy.linter import conda_recipe_v1_linter
 from conda_smithy.linter import messages as msg
 from conda_smithy.linter.hints import (
     hint_check_spdx,
+    hint_dependency_pins,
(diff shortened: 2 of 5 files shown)


mgorny:
I need to think about it more, and update tests, but I'm too tired today, so publishing what I have so far.

mgorny:
Okay, it's working. However, I'm having a second thoughts.

For example, right now we actually override `python` with `{{ python_min }}`. The check doesn't complain about that since it skips depspecs using variables, but the wider question remains.

jaimergp:
`python` is the one edge case, I'd say, so maybe we can skip that one. Maybe there are other packages in this one box, but generally we don't to override packages where ABI pinning is important.

@conda-forge/core, thoughts?

h-vetinari (commented):
How do we want to deal with cases like
```yaml
host:
  - foo
  - foo >2.3
```
This works from the POV of the bot, and allows recipe authors to still express constraints without interfering with the pinning mechanism. Arguably those constraints are basically redundant, but it is a pattern that exists with non-negligible frequency

mgorny:
> How do we want to deal with cases like
> 
> ```yaml
> host:
>   - foo
>   - foo >2.3
> ```
> 
> This works from the POV of the bot, and allows recipe authors to still express constraints without interfering with the pinning mechanism. Arguably those constraints are basically redundant, but it is a pattern that exists with non-negligible frequency

I suppose I could extend it to check for this pattern (i.e. emit a hint only if all depspecs for a package have constraints), but then I truly hate this pattern. I mean, if the behavior with depspecs is confusing in the first place, this pattern is even more confusing and I'd probably accidentally remove the duplicate, thinking it's a mistake.

beckermr:
I think we only hint when:

- the recipe is NOT using our noarch python pattern with python_min
- the item in host is in our pinnings
- there is only one instance of the item 
- the one instance of the item is NOT using jinja2 to specify a version 

The last condition is needed to exclude false positives like

```
host:
  - blah {{ blah }}
```

which are correct and simply explicitly specify what conda recipes do internally.

I understand folks don't like certain patterns, but that is not something to be debated here. The pattern above is a result of how conda recipes work right now.

mgorny:
> the recipe is NOT using our noarch python pattern with python_min

Isn't that already covered by the Jinja condition? Are you suggesting we also exclude `python` with explicit version number in that case?

mgorny:
Please let me know if that's what you had in mind.

beckermr:
> the recipe is NOT using our noarch python pattern with python_min

Yes jinja2 would already cover this.

beckermr (approved):

PR #2549, opened 2026-05-14
Fix `mktemp` in `run_osx_build.sh.tmpl`

holymonson:
On some old macos, mktemp doesn't support -p option.

Checklist


beckermr (changes requested):
This might break on newer macs that need this option.

holymonson:
> This might break on newer macs that need this option.

I don't have the latest macos on hand, so I can't comfirm that, but I test the latest mktemp and it still supports the old `mktemp "$OSX_SDK_DIR"/tmp.XXXXXXXX` format. IMHO, `-p` just was added for convenient, won't be a mandatory.

h-vetinari:
> This might break on newer macs that need this option.

By adding the slash, this should work AFAICT.

beckermr:
The `-p` option does this:

>  -p tmpdir, --tmpdir[=tmpdir]
>              Use tmpdir for the -t flag if the TMPDIR environment variable is not set.  Additionally, any provided template arguments will be interpreted relative to
>              the path specified as tmpdir.  If tmpdir is either empty or omitted, then the TMPDIR environment variable will be used.
> 

I don't think that is the same as the path.

holymonson:
> I don't think that is the same as the path.

It's not exactly the same. But the statement here is to test if inside `$OSX_SDK_DIR` writable, and several lines above has ensured `$OSX_SDK_DIR` is not empty, in this circumstance, they are the same.

jaimergp:
>  But the statement here is to test if inside $OSX_SDK_DIR writable, and several lines above has ensured $OSX_SDK_DIR is not empty, in this circumstance, they are the same.

This sounds good to me. What do you think @beckermr?

jaimergp:
@holymonson, can you add a changelog item under `news/` by copying the template? Please include the PR number, e.g. "Make sure writable location tests are compatible with older macOS versions (#2549)".

beckermr:
I'd like to see this PR tests in a feedstock on a recent osx machine before we merge.

jaimergp:
@beckermr see https://github.com/conda-forge/dav1d-feedstock/pull/34 (Windows error unrelated, obv.)

beckermr (approved):

jaimergp (approved):

PR #2551, opened 2026-05-14
Add `docker.run_args`

mgorny:
Checklist

Add a new `docker.run_args` setting that can be used to pass extra arguments to `docker run` for all CI providers.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1847,7 +1847,6 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
         # so it can be used in the general "render_github_actions" function
         # This avoid potential collisions with other CI providers :crossed_fingers:
         data["gha_os"] = GITHUB_ACTIONS_RUNS_ON[data["build_platform"]]["os"]
-        data["gha_with_gpu"] = False
 
         self_hosted_default = list(
             GITHUB_ACTIONS_RUNS_ON[data["build_platform"]]["self_hosted_labels"]
@@ -1876,6 +1875,7 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
             )
 
         data["gha_runs_on"] = []
+        with_gpu = False
         for label in labels:
             if label.startswith("cirun-"):
                 # Patch Cirun runners to add some extra debug info
@@ -1883,7 +1883,7 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
             if "gpu" in label.lower():
                 # Having 'gpu' in one label name is enough to trigger
                 # the extra docker args needed on Linux
-                data["gha_with_gpu"] = True
+                with_gpu = True
             data["gha_runs_on"].append(label)
 
         workflow_settings = get_workflow_settings(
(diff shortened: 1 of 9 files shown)


conda_smithy/schema.py
```diff
@@ -526,6 +526,13 @@ class WorkflowSettings(BaseModel):
         """),
     )
 
+    docker_run_args: Optional[
+        Union[str, list[conditional_value(str, None)], Nullable]
+    ] = Field(
+        default=[],
+        description="Additional arguments to pass to `docker run`.",
+    )
```
isuruf: This should be part of `CondaForgeDocker` and not in `WorflowSettings`.

h-vetinari: Or we absorb CondaForgeDocker into WorkflowSettings in some way (in the spirit of #2349)?

mgorny: I don't have a strong opinion. I figured out we want to go with `workflow-settings` because it provides finer granularity.

jaimergp: I'm on the "por qué no los dos" boat here.

I think it makes sense to embed `CondaForgeDocker` in workflow_settings so we can configure it per provider, and then add `run_args` in there. [There's no one using the `docker:` key](https://github.com/search?type=code&q=org%3Aconda-forge+path%3Aconda-forge.yml+docker%3A), anyway, so the risk is low.

jaimergp: @isuruf, any concerns with my proposal? Too much change for the little use this key gets? I can see that argument going both ways, fwiw: no one uses it, why change it; no one uses it, who cares if we change it.

isuruf: Yes, too much change for something that has very little usage.

conda_smithy/configure_feedstock.py
```diff
@@ -1902,6 +1902,7 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
             "github_actions",
             data["platform"],
             "D:" if on_hosted_runner or on_namespace else "C:",
+            "--gpus all" if with_gpu and platform == "linux" else "",
```
isuruf: This is a breaking change. Previously, if `docker.run_args` was set `--gpus all` was appended. Now if `workflow_settings.docker_run_args` is set, `--gpus all` is dropped.

mgorny: Yes. However, the one package using it is not affected.

isuruf: Right, but `--gpus all` is not really optional right? If a user has to override run args for some other reason, they have to remember to add `--gpus all` which is a pattern that I don't like.

mgorny: Well, I personally prefer giving the user the ability to override that but I can change that back.

conda_smithy/schema.py
mgorny: Sorry, so what is the preferred approach here?

isuruf: Let's keep the `docker` high level entry.

conda_smithy/configure_feedstock.py
isuruf: Me too, but this is just a fraction of arguments that we send to docker, so there's no point giving the user an ability to override just this small part.

conda_smithy/schema.py
mgorny: Sorry, does that mean "keep `docker` as-is and add to `workflow_settings`" or "add to `docker`"?

isuruf: "add to docker"

mgorny: Okay, will do that next.

mgorny: Done.

isuruf (approved):

PR #2553, opened 2026-05-15
Fix a crash when checking for overridden pins in v1 if:/else:

mgorny:
Checklist

It looks like the conditions are inlined already when in top-level `requirements` but not when in `outputs`.

### DIFF
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -476,7 +476,7 @@ def hint_rattler_build_bld_bat(
 def _check_pin_overridden(
     requirements_section: dict[str, list[str]],
     pins: set[str],
-    recipe_version: int = 1,
+    recipe_version: int,
 ) -> Generator[str]:
     from conda import CondaError
     from conda.models.match_spec import MatchSpec
@@ -521,7 +521,7 @@ def hint_dependency_pins(
     outputs_section,
     ci_support_files,
     hints,
-    recipe_version: int = 1,
+    recipe_version: int,
 ):
     """Hint for dependencies that override pinning"""
 
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -474,7 +474,9 @@ def hint_rattler_build_bld_bat(
 
 
 def _check_pin_overridden(
-    requirements_section: dict[str, list[str]], pins: set[str]
+    requirements_section: dict[str, list[str]],
+    pins: set[str],
+    recipe_version: int,
 ) -> Generator[str]:
     from conda import CondaError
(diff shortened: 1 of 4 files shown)


jaimergp:
> I suppose no news item, since it's a fix for the feature that wasn't yet released.

You can add the PR number to the news item of the feature.

conda_smithy/linter/hints.py
```diff
@@ -517,6 +521,7 @@ def hint_dependency_pins(
     outputs_section,
     ci_support_files,
     hints,
+    recipe_version: int,
```
jaimergp, line 524:
```suggestion
    recipe_version: int = 1,
```

Add a default value, otherwise we are technically breaking API.

jaimergp: `0` or `1`, either is fine.

conda_smithy/linter/hints.py
```diff
@@ -474,14 +474,18 @@ def hint_rattler_build_bld_bat(
 
 
 def _check_pin_overridden(
-    requirements_section: dict[str, list[str]], pins: set[str]
+    requirements_section: dict[str, list[str]],
+    pins: set[str],
+    recipe_version: int,
```
jaimergp, line 479:
```suggestion
    recipe_version: int = 1,
```

Same here.

mgorny:
Merged the suggestions. A bit surprised we're concerned about API stability when it hasn't been released yet.

jaimergp:
> A bit surprised we're concerned about API stability when it hasn't been released yet.

Ah, sorry, I didn't realize that. I had assumed these symbols existed before that, but yea, you are right. You can revert if you feel like it's better as a positional argument.

tests/test_lint_recipe.py
```diff
@@ -4718,21 +4718,12 @@ def test_lint_v1_unquoted_floats(
...
-build:
-  number: 1
-
+@pytest.mark.parametrize("outputs", (False, True))
+def test_hint_pinned_dependency_override(outputs: bool) -> None:
+    midpart = """\
 requirements:
   host:
-    - blah ${{ blah }}
+    - blah ${ blah }
```
jaimergp: Hm, what's this change for? 🤔

mgorny:
WTF, GitHub added your suggestions with an extra CR.

tests/test_lint_recipe.py
mgorny: Good catch. I must've accidentally replaced it when I was switching between f-strings and regular strings.

mgorny:
Reverted the API change and fixed the braces in the test.

jaimergp (approved):

PR #2562, opened 2026-05-26
Add a `workflow_settings.pagefile_size` key

### LINKED ISSUE #1971: ENH: Unified interface for creating swap files?

mgorny:
Checklist

Add a new `workflow_settings.pagefile_size` key that can be used to control page/swap file size regardless of CI provider and platform. It replicates the existing swapfile logic for Azure/Linux to GHA. For Windows, it defines `SET_PAGEFILE_SIZE` per
https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/160, along with compatibility code setting `SET_PAGEFILE` for compatibility with the current verison of `conda-forge-ci-setup`.

Closes #1971

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2567,14 +2567,25 @@ def _read_forge_config(forge_dir, forge_yml=None):
             )
             del config["azure"]["settings_linux"]["swapfile_size"]
         else:
-            # Convert old keys to new settings.
-            config["workflow_settings"]["pagefile_size"].append(
-                {
-                    "provider": "azure",
-                    "os": "linux",
-                    "value": config["azure"]["settings_linux"]["swapfile_size"],
-                }
-            )
+            value_re = re.compile(r"(\d+)GiB")
+            if (
+                value_match := value_re.fullmatch(
+                    config["azure"]["settings_linux"]["swapfile_size"]
+                )
+            ) is not None:
+                config["workflow_settings"]["pagefile_size"].append(
+                    {
+                        "provider": "azure",
+                        "os": "linux",
+                        "value": int(value_match.group(1)),
+                    }
+                )
+            else:
+                logger.warning(
+                    "`azure.settings_linux.swapfile_size` has invalid value `%s`, "
+                    "should be `{number}GiB`",
+                    config["azure"]["settings_linux"]["swapfile_size"],
+                )
 
     if "SET_PAGEFILE" in file_config.get("azure", {}).get("settings_win", {}).get(
         "variables", {}
@@ -2592,9 +2603,9 @@ def _read_forge_config(forge_dir, forge_yml=None):
                     "provider": "azure",
                     "os": "win",
                     "value": (
-                        "16GiB"
+                        16
                         if config["azure"]["settings_win"]["variables"]["SET_PAGEFILE"]
-                        else "0"
+                        else 0
                     ),
                 }
             )
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1904,11 +1904,15 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
             "D:" if on_hosted_runner or on_namespace else "C:",
         )
         data.update(workflow_settings)
+        script_suffix = ".bat" if platform == "win" else ".sh"
         if data["store_build_artifacts"]:
-            script_suffix = ".bat" if platform == "win" else ".sh"
             template_files.append(
                 f".scripts/create_conda_build_artifacts{script_suffix}"
             )
+        if data["pagefile_size"] != 0 and platform in ("linux", "win"):
+            template_files.append(f".scripts/create_pagefile{script_suffix}")
+            if platform == "win":
+                template_files.append(".scripts/SetPageFileSize.ps1")
 
         if platform == "linux":
             data["docker_run_args"] = forge_config["docker"]["run_args"]
@@ -2046,10 +2050,14 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
             config_rendered["docker_run_args"] = forge_config["docker"]["run_args"]
 
         config_rendered.update(workflow_settings)
+        script_suffix = ".bat" if platform == "win" else ".sh"
         if config_rendered["store_build_artifacts"]:
             config_rendered["SHORT_CONFIG"] = data["short_config_name"]
-            script_suffix = ".bat" if platform == "win" else ".sh"
             template_files.append(f".scripts/create_conda_build_artifacts{script_suffix}")
+        if config_rendered["pagefile_size"] != 0 and platform in ("linux", "win"):
+            template_files.append(f".scripts/create_pagefile{script_suffix}")
+            if platform == "win":
+                template_files.append(".scripts/SetPageFileSize.ps1")
         azure_settings["strategy"]["matrix"][data["config_name"]] = config_rendered
         # fmt: on
 
@@ -2972,6 +2980,9 @@ def get_common_scripts(forge_dir):
         "run_osx_build.sh",
         "create_conda_build_artifacts.bat",
         "create_conda_build_artifacts.sh",
+        "create_pagefile.bat",
+        "create_pagefile.sh",
+        "SetPageFileSize.ps1",
     ]:
         yield os.path.join(forge_dir, ".scripts", old_file)
 
@@ -2991,6 +3002,9 @@ def clear_scripts(forge_dir):
             "run_win_build.bat",
             "create_conda_build_artifacts.bat",
             "create_conda_build_artifacts.sh",
+            "create_pagefile.bat",
+            "create_pagefile.sh",
+            "SetPageFileSize.ps1",
         ]:
             remove_file(os.path.join(forge_dir, folder, old_file))
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2592,7 +2592,7 @@ def _read_forge_config(forge_dir, forge_yml=None):
                     "provider": "azure",
                     "os": "win",
                     "value": (
-                        "8GiB"
+                        "16GiB"
                         if config["azure"]["settings_win"]["variables"]["SET_PAGEFILE"]
                         else "0"
                     ),
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2559,6 +2559,46 @@ def _read_forge_config(forge_dir, forge_yml=None):
                 "variables"
             ].pop("CONDA_FORGE_DOCKER_RUN_ARGS")
 
+    if "swapfile_size" in file_config.get("azure", {}).get("settings_linux", {}):
+        if "pagefile_size" in file_config.get("workflow_settings", {}):
+            logger.warning(
+                "`azure.settings_linux.swapfile_size` is ignored when "
+                "`workflow_settings.pagefile_size` is set",
+            )
+            del config["azure"]["settings_linux"]["swapfile_size"]
+        else:
+            # Convert old keys to new settings.
+            config["workflow_settings"]["pagefile_size"].append(
+                {
+                    "provider": "azure",
+                    "os": "linux",
+                    "value": config["azure"]["settings_linux"]["swapfile_size"],
+                }
+            )
+
+    if "SET_PAGEFILE" in file_config.get("azure", {}).get("settings_win", {}).get(
+        "variables", {}
+    ):
+        if "pagefile_size" in file_config.get("workflow_settings", {}):
+            logger.warning(
+                "`azure.settings_win.variables.SET_PAGEFILE` is ignored when "
+                "`workflow_settings.pagefile_size` is set",
+            )
+            del config["azure"]["settings_win"]["variables"]["SET_PAGEFILE"]
+        else:
+            # Convert old keys to new settings.
+            config["workflow_settings"]["pagefile_size"].append(
+                {
+                    "provider": "azure",
+                    "os": "win",
+                    "value": (
+                        "8GiB"
+                        if config["azure"]["settings_win"]["variables"]["SET_PAGEFILE"]
+                        else "0"
+                    ),
+                }
+            )
+
     # check for conda-smithy 2.x matrix which we can't auto-migrate
     # to conda_build_config
     if file_config.get("matrix") and not os.path.exists(
--- a/conda_smithy/schema.py
+++ b/conda_smithy/schema.py
@@ -90,7 +90,11 @@ class AzureRunnerSettings(BaseModel):
     )
 
     swapfile_size: Optional[Union[str, Nullable]] = Field(
-        default=None, description="Swapfile size in GiB"
+        default=None,
+        description=cleandoc("""
+            Deprecated. Swapfile size in GiB.
+            Use `workflow_settings.pagefile_size` instead.
+        """),
     )
 
     timeout_in_minutes: Optional[int] = Field(
@@ -206,7 +210,8 @@ class AzureConfig(BaseModel):
               `workflow_settings.build_workspace_dir` instead.
             - ~~`MINIFORGE_HOME`~~: Location of the base environment installation. Deprecated,
               use `workflow_settings.tools_install_dir` instead.
-            - `SET_PAGEFILE`: `"True"` to increase the pagefile size via conda-forge-ci-setup.
+            - ~~`SET_PAGEFILE`~~: `"True"` to increase the pagefile size via conda-forge-ci-setup.
+              Deprecated, use `workflow_settings.pagefile_size` instead.
 
             If you are running out of space in `D:`, consider changing to `C:`.
             It's a slower drive but has more space available. We recommend you keep
@@ -528,6 +533,16 @@ class WorkflowSettings(BaseModel):
         default=[],
         description=cleandoc("""
         Directory to build in.
+        """),
+    )
+
+    pagefile_size: Optional[
+        Union[str, list[conditional_value(str, None)], Nullable]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Override the default paging (swap) file size. Specify along with
+        the unit, e.g. "8GiB".
         """),
     )
 
--- a/conda_smithy/schema.py
+++ b/conda_smithy/schema.py
@@ -537,12 +537,12 @@ class WorkflowSettings(BaseModel):
     )
 
     pagefile_size: Optional[
-        Union[str, list[conditional_value(str, None)], Nullable]
+        Union[int, list[conditional_value(int, None)], Nullable]
     ] = Field(
         default=[],
         description=cleandoc("""
-        Override the default paging (swap) file size. Specify along with
-        the unit, e.g. "8GiB".
+        Override the default paging (swap) file size, in GiB.
+        For example, 8 means 8 GiB.
         """),
     )
 
--- a/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
@@ -48,8 +48,9 @@ jobs:
          df -h
     displayName: Manage disk space
 {%- endif %}
-{%- if not (azure.settings_linux.swapfile_size|string).startswith("0") %}
   - script: |
+        PAGEFILE_SIZE=$(pagefile_size)
+        [[ ${PAGEFILE_SIZE} == 0* ]] && exit
         SWAPFILE=/swapfile
         # If there is already a swapfile, disable it and remove it
         if swapon --show | grep -q $SWAPFILE; then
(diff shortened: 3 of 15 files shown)


conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -133,6 +135,21 @@ jobs:
...
+                echo "Disabling existing swapfile..."
+                sudo swapoff $SWAPFILE
+            fi
+            [ -f "$SWAPFILE" ] && sudo rm -f $SWAPFILE
+            sudo fallocate -l "${SET_PAGEFILE_SIZE}" $SWAPFILE
+            sudo chmod 600 $SWAPFILE
+            sudo mkswap $SWAPFILE
+            sudo swapon $SWAPFILE
+            echo "::endgroup::"
+        fi
```
mgorny: Should we move that logic somewhere?

conda_smithy/schema.py
```diff
@@ -531,6 +536,16 @@ class WorkflowSettings(BaseModel):
         """),
     )
 
+    pagefile_size: Optional[
+        Union[str, list[conditional_value(str, None)], Nullable]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Override the default paging (swap) file size. Specify along with
+        the unit, e.g. "8GiB".
```
mgorny: There's an open question what values we're going to accept. For now, I went with the existing logic which passes it to `fallocate` and accepts whatever `fallocate` does. We may want to add some kind of normalization and validation here too.

conda_smithy/templates/github-actions.yml.tmpl
jaimergp: I assume this this is shared with Azure, and we want to avoid duplication in the conda-smithy source code. We have three options:

1. Template this in Jinja so we can define it in a single place in the Python source
2. Add a helper script that is shared among providers
3. Port the whole thing to conda-forge-ci-setup


I think I'm partial to (2), but I don't like that ci-setup governs the pagefile logic for Windows, but smithy does it for Unix :/

conda_smithy/schema.py
jaimergp: `fallocate` [allows](https://man7.org/linux/man-pages/man1/fallocate.1.html):

> The length and offset arguments may be followed by the
       multiplicative suffixes KiB (=1024), MiB (=1024\*1024), and so on
       for GiB, TiB, PiB, EiB, ZiB, and YiB (the "iB" is optional, e.g.,
       "K" has the same meaning as "KiB") or the suffixes KB (=1000), MB
       (=1000\*1000), and so on for GB, TB, PB, EB, ZB, and YB.

Our Windows logic [allows](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/d80886015602907e368d98580772d7359c71bfc6/recipe/SetPageFileSize.ps1#L7) GB,KB -like suffixes.

So I'd say we stick to only that type (1000x) of suffixes (KB, MB, GB, TB).

jaimergp: Alternatively, why would we want anything else than GB or MB? I'd be ok with hardcoding that choice and only letting users specify a positive integer.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: That said, part of this logic is unclear to me. I mean why are accounting for disabling a swapfile first whose path is precisely `/swapfile`?

conda_smithy/schema.py
mgorny: Currently, `swapfile_size` seems to be consistently used with `nnnGiB`, so we probably need to support at least that for backwards compatibility. We may need to support everything `fallocate` allows for 100% backwards compatibility.

Plain integers are cleaner indeed (and most importantly, we could implement them consistently across platforms). Perhaps allow `nnnGiB` for the old `swapfile_size`, convert it to `nnn` in the new value and ignore any other value in `swapfile_size` with a warning?

mgorny: Done that now.

conda_smithy/templates/github-actions.yml.tmpl
h-vetinari: I'm not against moving out the pagefile logic from [ci-setup](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/d80886015602907e368d98580772d7359c71bfc6/recipe/run_conda_forge_build_setup_win.bat#L24-L47) into smithy templates; it'd need modification on the default AZP/GHA job templates though, and we'd need to copy https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/main/recipe/SetPageFileSize.ps1 too. Nothing impossible though.

conda_smithy/schema.py
```diff
@@ -531,6 +536,16 @@ class WorkflowSettings(BaseModel):
         """),
     )
 
+    pagefile_size: Optional[
```
h-vetinari, line 539: I'm not sure we want to call this pagefile_size; IMO the nomenclature of "swapfile" is much more common. Probably it makes more sense to homogenize on the unix naming, rather than the windows one.

h-vetinari (commented):
Thanks for this. This is going in the right direction; I like @jaimergp's idead to move the windows logic here from ci-setup, and I think it'd make sense to wrap this in a scriptlet that can be called from various providers (which addresses the deduplication part). We may even want to introduce a separate template for this that gets copied/populated under `.scripts`, like we already do for `create_conda_build_artifacts.{sh,bat}.tmpl`.

conda_smithy/schema.py
mgorny: I went for "paging" because that's what Wikipedia called it ("swap file" redirects to "memory paging").

h-vetinari: OK, this is not a hill I care much about; feel free to stick to `pagefile_size`.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: Oh, I was just thinking about Linux, but indeed moving the Windows logic too would make things easier to update (at the cost of duplicating them in every feedstock but I guess that's not a significant concern, especially given how few feedstocks actually use that).

mgorny:
Okay, now the scripts are part of templates. I've integrated @h-vetinari 's PR to allow changeable pagefile size on Windows, and @jakirkham 's work to select drive better, plus my own idea of checking if `D:` is actually available (untested).

conda_smithy/configure_feedstock.py
```diff
@@ -1904,11 +1904,15 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
...
+        script_suffix = ".bat" if platform == "win" else ".sh"
         if data["store_build_artifacts"]:
-            script_suffix = ".bat" if platform == "win" else ".sh"
             template_files.append(
                 f".scripts/create_conda_build_artifacts{script_suffix}"
             )
+        if data["pagefile_size"] != 0 and platform in ("linux", "win"):
+            template_files.append(f".scripts/create_pagefile{script_suffix}")
+            if platform == "win":
+                template_files.append(".scripts/SetPageFileSize.ps1")
```
h-vetinari, line 1915: the fact that we do the same thing both in `_github_actions_specific_setup`...

conda_smithy/configure_feedstock.py
```diff
@@ -2046,10 +2050,14 @@ def _azure_specific_setup(jinja_env, forge_config, forge_dir, platform):
...
         config_rendered.update(workflow_settings)
+        script_suffix = ".bat" if platform == "win" else ".sh"
         if config_rendered["store_build_artifacts"]:
             config_rendered["SHORT_CONFIG"] = data["short_config_name"]
-            script_suffix = ".bat" if platform == "win" else ".sh"
             template_files.append(f".scripts/create_conda_build_artifacts{script_suffix}")
+        if config_rendered["pagefile_size"] != 0 and platform in ("linux", "win"):
+            template_files.append(f".scripts/create_pagefile{script_suffix}")
+            if platform == "win":
+                template_files.append(".scripts/SetPageFileSize.ps1")
```
h-vetinari, line 2060: ... and `_azure_specific_setup` would seem to imply that this is not actually that provider-specific. should we move this into a common setup step somewhere? As it's pre-existing, it doesn't have to be in this PR.

h-vetinari (approved):
Love this! 🤩 

Not sure why it's still a draft, but basically LGTM already!

conda_smithy/templates/create_pagefile.bat.tmpl
```diff
@@ -0,0 +1,24 @@
+setlocal enableextensions enabledelayedexpansion
+
+set SET_PAGEFILE_SIZE=%1%
```
jaimergp:
```suggestion
set SET_PAGEFILE_SIZE=%1
```

I would swear the syntax for args uses a single % 🤔

conda_smithy/templates/create_pagefile.bat.tmpl
```diff
@@ -0,0 +1,24 @@
...
+
+:: Increase pagefile size, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
+:: Both in the recipe and in the final package, this script is co-located with SetPageFileSize.ps1, see meta.yaml
```
jaimergp: I don't understand this comment about `meta.yaml`

conda_smithy/templates/create_pagefile.bat.tmpl
```diff
@@ -0,0 +1,24 @@
...
+set ThisScriptsDirectory=%~dp0
+set EntryPointPath=%ThisScriptsDirectory%SetPageFileSize.ps1
+REM use different drive than CONDA_BLD_PATH-location for pagefile
```
jaimergp:
```suggestion
REM Try to use different drive than CONDA_BLD_PATH-location for pagefile, if available
```

conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -0,0 +1,19 @@
...
+set -ex
+
+SET_PAGEFILE_SIZE=${1}
+
+[[ ${SET_PAGEFILE_SIZE} -le 0 ]] && exit
+
+SWAPFILE=/swapfile
+# If there is already a swapfile, disable it and remove it
+if swapon --show | grep -q "^${SWAPFILE}"; then
+	sudo swapoff "${SWAPFILE}" || true
```
jaimergp:
```suggestion
    sudo swapoff "${SWAPFILE}" || true
```

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -73,6 +74,7 @@ jobs:
 {%- if upload_on_branch %}
         UPLOAD_ON_BRANCH: {{ upload_on_branch }}
 {%- endif %}
+        SET_PAGEFILE_SIZE: $(pagefile_size)
```
jaimergp: Why the indirection with the env var here but not on Linux?

conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -ex
+
+SET_PAGEFILE_SIZE=${1}
+
+[[ ${SET_PAGEFILE_SIZE} -le 0 ]] && exit
```
jaimergp: `exit 0`? Shouldn't this be an error?

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -54,6 +54,7 @@ jobs:
             tools_install_dir: {{ data.tools_install_dir }}
             build_workspace_dir: {{ data.build_workspace_dir }}
             docker_run_args: {{ data.docker_run_args }}
+            pagefile_size: {{ data.pagefile_size }}
```
jaimergp, line 57: Should we only populate this if `>0`? We know that at rerender time, so no need to delegate to runtime, do we?

conda_smithy/utils.py
```diff
@@ -409,3 +409,6 @@ def fill_workflow_settings_defaults(
             # TODO: switch to normalizing all paths once we're ready
             "win": rf"{tools_drive}\\bld\\",
         }[os]
+
+    if workflow_settings.get("pagefile_size") is None:
+        workflow_settings["pagefile_size"] = "0"
```
jaimergp: I would have assumed this needs to be an `int`

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
```diff
@@ -48,21 +48,9 @@ jobs:
...
-        if swapon --show | grep -q $SWAPFILE; then
-            echo "Disabling existing swapfile..."
-            sudo swapoff $SWAPFILE || true
-        fi
-        [ -f "$SWAPFILE" ] && sudo rm -f $SWAPFILE
-        sudo fallocate -l {{ azure.settings_linux.swapfile_size }} $SWAPFILE
-        sudo chmod 600 $SWAPFILE
-        sudo mkswap $SWAPFILE
-        sudo swapon $SWAPFILE
+        .scripts/create_pagefile.sh "$(pagefile_size)"
```
jaimergp, line 56: Is this interpolated by AZP before running or will we run into a "command not found" error due to the `$()` syntax?

conda_smithy/configure_feedstock.py
```diff
@@ -2559,6 +2567,57 @@ def _read_forge_config(forge_dir, forge_yml=None):
...
+                        "os": "linux",
+                        "value": int(value_match.group(1)),
+                    }
+                )
+            else:
+                logger.warning(
+                    "`azure.settings_linux.swapfile_size` has invalid value `%s`, "
+                    "should be `{number}GiB`",
+                    config["azure"]["settings_linux"]["swapfile_size"],
+                )
```
jaimergp, line 2596: Not sure if a warning is enough here. Should it be an error?

jaimergp:
> plus my own idea of checking if D: is actually available (untested).

I think the one without a D: drive was `windows-11-arm`, if you want to check.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
mgorny: Yeah, `$(...)` is some AZP magic. See e.g. https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1527819&view=logs&j=d818e662-7181-5ce9-b265-50d03742b7af&t=367317f3-9b77-59ac-9e12-e56370c98acd

conda_smithy/templates/azure-pipelines-win.yml.tmpl
mgorny: Historical reasons :-). Will change.

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -57,6 +57,7 @@ jobs:
 {%- endif %}
 
     - script: |
+        call ".scripts\create_pagefile.bat" %SET_PAGEFILE_SIZE%
```
mgorny:
```suggestion
        call ".scripts\create_pagefile.bat" $(pagefile_size)
```

conda_smithy/templates/azure-pipelines-win.yml.tmpl
mgorny:
```suggestion
```

conda_smithy/templates/create_pagefile.bat.tmpl
mgorny: Leftover from ci-setup. I'll remove it, it's not really necessary anyway, given that the file is generated by smithy now.

mgorny:
```suggestion
```

conda_smithy/templates/create_pagefile.sh.tmpl
mgorny: Zero means it's disabled, covering negative values "just in case". I did the check here not to have to duplicate the condition. (It's kinda duplicated in GHA because we're doing fancy output there, but not in AZP.)

conda_smithy/templates/github-actions.yml.tmpl
mgorny: It gets messy in AZP where the variables are automatic. I think `0` is a cleaner default for `int` than `None` :-).

conda_smithy/configure_feedstock.py
mgorny: Yeah, we definitely should look into improving this at some point. But too complex for this PR.

conda_smithy/configure_feedstock.py
mgorny: I think we didn't want to kill the bot.

mgorny: (We may want to add a lint for valid values, but I'm not sure if we care that much.)

conda_smithy/utils.py
mgorny: Indeed. Forgot to update it.

mgorny:
```suggestion
        workflow_settings["pagefile_size"] = 0
```

conda_smithy/configure_feedstock.py
jaimergp: I checked and all instances in conda-forge are using legal values, so no need to do anything else here, since it's deprecated and they will be hinted towards moving to the new setting.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
jaimergp: Ah, I thought it was only available in some fields, not in scripts too. That's odd, that syntax can be breaking on Bash and others. Anyway, good.

jaimergp:
Cool, cool, cool! Do you have a draft PR somewhere where you have rerendered a feedstock with this PR and shows it's working live?

mgorny:
https://github.com/conda-forge/lancedb-feedstock/pull/44 is what I've been testing it on. Note that it's got a long ugly history of all my iterations :-).

mgorny:
> > plus my own idea of checking if D: is actually available (untested).
> 
> I think the one without a D: drive was `windows-11-arm`, if you want to check.

Thanks! Confirmed working as intended: https://github.com/mgorny/lancedb-feedstock/actions/runs/26651910998/job/78552201530

news/2562-pagefile-size.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* ``workflow_settings.pagefile_size`` to set page file (swap file) size consistently across CI providers, on Linux and Windows. (#2562)
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
```
jaimergp: Let's add the notes here for the variables under the legacy `azure_settings`

jaimergp:
Now that the pagefile PS1 script lives here we should mark it as deprecated in ci-setup. It will still need to stay there for a bit for feedstocks that have not rerendered recently, though.

mgorny:
> Now that the pagefile PS1 script lives here we should mark it as deprecated in ci-setup. It will still need to stay there for a bit for feedstocks that have not rerendered recently, though.

Given the use is limited, I suppose we can even track its removal.

jaimergp (approved):

h-vetinari:
Very nice. Hope you take this out of draft soon. ;)

The one thing I'd still ask for is to remove the `SET_` from `SET_PAGEFILE_SIZE`. It's redundant and potentially confusing.

The historical reason is that `SET_PAGEFILE` was a boolean, where it made sense to not just call it `PAGEFILE`. If that variable had had to coexist with the `_SIZE` variant then of course we'd have kept them consistent (my old ci-setup PR), but now that we're cleaning this up and don't need the boolean anymore, I'd much prefer to just call this `PAGEFILE_SIZE`

jakirkham:
Thank you for the cleanup Michał! 🙏

This generally seems like a good idea

Though haven't dug in closely yet

mgorny:
> The one thing I'd still ask for is to remove the `SET_` from `SET_PAGEFILE_SIZE`. It's redundant and potentially confusing.

Done. TBH, I just followed the convention from your draft PR, but yeah, `PAGEFILE_SIZE` is better.

h-vetinari (approved):
🥳

PR #2568, opened 2026-06-02
Add `workflow_settings.free_disk_space`

### LINKED ISSUE #1949: ENH: extend free_disk_space to windows?

mgorny:
Checklist

Part of issue #2349
Fixes #1949
Closes #1966

Add a new `workflow_settings.free_disk_space` setting as a replacement for the legacy `azure.free_disk_space` and
`github_actions.free_disk_space`.  The old values are converted automatically.  All values are also normalized to arrays to simplify templates.  The values can be conditional to individual jobs.

Progress:
- [x] GHA template support
- [x] Azure template support
- [x] Tests
- [x] osx/win support


conda_smithy/schema.py
```diff
@@ -546,6 +546,27 @@ class WorkflowSettings(BaseModel):
...
+                conditional_value(
+                    Union[bool, list[Literal["apt", "cache", "docker"]]], False
+                )
+            ],
+            Nullable,
+        ]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Free up disk space before running the Docker container for building on Linux.
```
jaimergp: Is it Linux only? 🤔 I thought we were going to do it OS-agnostic too.

conda_smithy/templates/github-actions.yml.tmpl
jaimergp, line 1: How reasonable is to attempt a CI provider agnostic script for this, so we can share it with Azure?

conda_smithy/schema.py
mgorny: > I suppose extending to other systems could be handled as a separate PR.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: The code looks roughly equivalent, so definitely doable.

conda_smithy/schema.py
h-vetinari: See #1949; John had a draft to extend this to windows in #1966

h-vetinari: I can live with deferring other platforms to a separate PR, but then we need some thought how os-specific `workflow_settings` will work. IMO there should be an error in smithy if someone does
```
workflow_settings:
  free_disk_space: true
```
because it gives the false impression of being cross-platform. 

`build_workspace_dir` already has a platform-specific warning (e.g. for setting windows paths on unix), so hopefully this should be a ~solved problem already.

mgorny: Okay, I'll start looking into other platforms.

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,36 @@
...
+           /usr/local/lib/android \
+           /usr/local/share/powershell \
+           /usr/share/dotnet \
+           /usr/share/swift \
+           ; do
+    sudo rsync --stats -a --delete /opt/empty_dir/ $d || true
+  done
+fi
+
+if [[ ${FREE_DISK_SPACE} == *,apt,* ]]; then
```
h-vetinari: not sure if this is a bash-ism that I'm not aware of, but I don't see how you can assume every entry (especially first/last) of the list is surrounded by commas; why not?
```suggestion
if [[ ${FREE_DISK_SPACE} == *apt* ]]; then
```

mgorny: Two extra commas are added when calling the script via the template. I don't mind making it simpler but @jaimergp preferred this version, so we don't accidentally match if keys overlap in the future.

conda_smithy/schema.py
mgorny: Ok, added macOS and Windows; with Windows supporting only `cache` and macOS not doing anything right now. Does anyone have a suggestion what we should be cleaning there?

mgorny:
Demo feedstock: https://github.com/conda-forge/polars-feedstock/pull/380

mgorny:
I can't reproduce the test failure locally. I suspect the schema generation is not reproducible.

h-vetinari:
If you squint a bit, the following could also be considered ~freeing~ making more disk space usable.
https://github.com/conda-forge/conda-smithy/blob/7ed6300656d706b62cd98734494401cf1e7577cc/conda_smithy/templates/github-actions.yml.tmpl#L182-L186

Not sure if/how we want to unify these things, but the named-arg-style approach already used linux might also be applicable to windows?

mgorny:
Yeah, I suppose that'd work, and it would definitely make sense to avoid inlining that script.

mgorny:
> Not sure if/how we want to unify these things, but the named-arg-style approach already used linux might also be applicable to windows?

To be honest, I'd keep it as a separate key. It's not like keys are expensive, and the backwards compatibility logic would end up being a bit messy here. And in the end, `resize_partitions: true` is more obvious than `free_disk_space: resize-partitions` or likewise. Or perhaps even:

```yaml
workflow_settings:
  free_disk_space:
    - os: win
      value: resize-partitions
    - os: linux
      value: true
```

mgorny:
Either way, I'd like to avoid adding more into this PR. We can add new functions as a followup, but I'd like to get what we have now reviewed.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
```diff
@@ -21,34 +24,11 @@ jobs:
...
-                                  google-chrome-stable \
-                                  microsoft-edge-stable
-         sudo apt-get autoremove -y >& /dev/null
-         sudo apt-get autoclean -y >& /dev/null
-{%- endif %}
-{%- if (azure.free_disk_space is iterable) and ('docker' in azure.free_disk_space) %}
-         sudo docker image prune --all --force
-{%- endif %}
-         df -h
+        .scripts/free_disk_space.sh ",$(free_disk_space),"
```
h-vetinari: can we add a comment about the role of the commas here?

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -ex
+
+FREE_DISK_SPACE=${1}
+
+if [[ ${FREE_DISK_SPACE} == *,cache,* ]]; then
+  case ${OS} in
```
h-vetinari: so "cache" becomes the universal key that every platform would want to use...

conda_smithy/schema.py
```diff
@@ -546,6 +550,27 @@ class WorkflowSettings(BaseModel):
...
 
+    free_disk_space: Optional[
+        Union[
+            bool,
+            list[Literal["apt", "cache", "docker"]],
+            list[
+                conditional_value(
+                    Union[bool, list[Literal["apt", "cache", "docker"]]], False
+                )
+            ],
```
h-vetinari: I'm unsure about the multi-type approach here, and mixing in platform-specific options. I think it may be cleaner/simpler to either do
```python
free_disk_space: Optional[Union[Literal[None, "quick", "extensive"]]]
```
(or however the bikeshed is coloured), which would trade granularity for an easier interface[^1]; or alternatively, we could make `free_disk_space: Optional[bool]` to just mean the `cache` portion, and introduce separate (but linux-specific-by-definition) switches for cleaning up docker & apt.

[^1]: because, let's face it, for recipe authors it's a simple trade-off of "make my recipe pass" vs. "how much extra time does my CI take"

conda_smithy/schema.py
h-vetinari: The description here needs an update, in one way or another

h-vetinari (commented):
Thinking about the interface here, and how to deal with platform-specific stuff in an ostensibly cross-platform `value:`

conda_smithy/schema.py
jaimergp: I mean, it could be both:

```python
Union[
	bool,  # true -> quick
	Literal["all"], # opt-in to everything
	list[Literal["apt", "cache", "docker"]],  # as granular as it gets
]
```

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -ex
+
+FREE_DISK_SPACE=${1}
+
```
jaimergp, line 6: Can we report the space available with `df -h` at the beginning too?

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+    macos)
+      # TODO
+      ;;
+    windows)
+      DIRS_TO_REMOVE=(
+        C:/hostedtoolcache/windows
+      )
+
+      # rm is one of the fastest methods to remove files on Windows
+      rm -rf "${DIRS_TO_REMOVE[@]}"
```
jaimergp: I think we should allow this step to fail, because on some GHA providers (Blacksmith, Namespace, whatever Cirun is connecting us to) these directories may not exist. With the current `set -ex` it would fail.

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+      ;;
+  esac
+fi
+
+if [[ ${FREE_DISK_SPACE} == *,apt,* && ${OS} == ubuntu ]]; then
```
jaimergp: Let's check that `apt-get` indeed exists as a command too.

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+  BROWSERS="firefox google-chrome-stable microsoft-edge-stable"
+  BROWSERS_TO_REMOVE=$(dpkg --get-selections $BROWSERS 2>/dev/null | awk '{print $1}')
+  sudo apt-get remove --purge -y $BROWSERS_TO_REMOVE
+
+  sudo apt-get autoremove -y >& /dev/null
+  sudo apt-get autoclean -y >& /dev/null
+fi
+
+if [[ ${FREE_DISK_SPACE} == *,docker,* && ${OS} == ubuntu ]]; then
```
jaimergp: Let's check whether `docker` is a known command as well, just in case.

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+        /opt/hostedtoolcache
+        /usr/lib/jvm
+        /usr/local/.ghcup
+        /usr/local/lib/android
+        /usr/local/share/powershell
+        /usr/share/dotnet
+        /usr/share/swift
+      )
+
+      sudo rm -rf "${DIRS_TO_REMOVE[@]}"
```
jaimergp:
```suggestion
      sudo rm -rf "${DIRS_TO_REMOVE[@]}" || true
```

Same as in Windows; these may not exist.

conda_smithy/templates/free_disk_space.sh.tmpl
jaimergp:
```suggestion
      rm -rf "${DIRS_TO_REMOVE[@]}" || true
```

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+      # rm is one of the fastest methods to remove files on Windows
+      rm -rf "${DIRS_TO_REMOVE[@]}"
+      ;;
+  esac
+fi
+
+if [[ ${FREE_DISK_SPACE} == *,apt,* && ${OS} == ubuntu ]]; then
+  BROWSERS="firefox google-chrome-stable microsoft-edge-stable"
+  BROWSERS_TO_REMOVE=$(dpkg --get-selections $BROWSERS 2>/dev/null | awk '{print $1}')
+  sudo apt-get remove --purge -y $BROWSERS_TO_REMOVE
```
jaimergp:
```suggestion
  sudo apt-get remove --purge -y $BROWSERS_TO_REMOVE || true
```

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+        /usr/local/lib/android
+        /usr/local/share/powershell
+        /usr/share/dotnet
+        /usr/share/swift
+      )
+
+      sudo rm -rf "${DIRS_TO_REMOVE[@]}"
+      ;;
+    macos)
+      # TODO
```
jaimergp: I checked https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md and there are a few things we can remove.

- `$ANDROID_HOME`
- `brew`'s cache: `brew cleanup --prune-all && rm -rf "$(brew --cache)"`
- a few browsers

In practice, folks are fine with [removing a few dirs only](https://github.com/search?q=%22rm+-rf+%2FUsers%2Frunner%2Fhostedtoolcache%22+macos-+language%3AYAML+%22df+-h%22+NOT+is%3Afork&ref=opensearch&type=code):

```suggestion
      DIRS_TO_REMOVE=(
        /Users/runner/Library/Android
        /Users/runner/.dotnet
        /Users/runner/hostedtoolcache
      )
      sudo rm -rf "${DIRS_TO_REMOVE[@]}" || true
```

conda_smithy/templates/free_disk_space.sh.tmpl
```diff
@@ -0,0 +1,50 @@
...
+      ;;
+    windows)
+      DIRS_TO_REMOVE=(
+        C:/hostedtoolcache/windows
```
jaimergp:
```suggestion
        C:/hostedtoolcache/windows
        C:/Android
```

conda_smithy/schema.py
```diff
@@ -546,6 +550,27 @@ class WorkflowSettings(BaseModel):
...
+            Nullable,
+        ]
+    ] = Field(
+        default=[],
+        description=cleandoc("""
+        Free up disk space before running the Docker container for building on Linux.
+        The following components can be cleaned up: `apt`, `cache`, `docker`.
+        When set to `true`, only `apt` and `cache` are cleaned up.
+        Set it to the full list to clean up all components.
```
jaimergp: My suggested description:

```suggestion
        Free up disk space before running the builds.

        - On Linux, the following components can be cleaned up:
          `apt`, `cache`, `docker`. When set to `true`, only `apt`
          and `cache` are used. Set it to the full list to
          clean up all components.
        - On macOS, only `cache` is supported (same as `true`).
        - On Windows, only `cache` is supported (same as `true`).
          Note that only `C:/` is cleaned up, so this is only useful
          if you are targeting this drive via `build_workspace_dir`,
          `tools_install_dir`, or `pagefile_size`.
```

conda_smithy/schema.py
jaimergp: Proposed something in https://github.com/conda-forge/conda-smithy/pull/2568/files#r3361885388

conda_smithy/schema.py
h-vetinari: What I'm not super keen on is implying that `apt` or `docker` are legitimate values anywhere but on linux. But perhaps I'm over-designing this? Could be that it's not that important.

My comment is based on the fact that I've only ever seen `true` or the full set of `apt, cache, docker` for this, so I'm wondering if this granularity really buys us anything? I kinda like `quick` vs. `extensive` (or `all`), that should IMO provide enough for what recipe authors use this for.

BTW, I've used this PR today in https://github.com/conda-forge/pyg-lib-feedstock/commit/6274e016a15ef9c5f9ab190c8111d9a9ce2f7d4b, and I think that
```yaml
workflow_settings:
  free_disk_space:
    - value: ["apt", "cache", "docker"]
      os: linux
```
looks weird? I mean having `value:` (itself already a list item) be again a list itself.

mgorny: If we're changing the format, then perhaps let's avoid lists indeed (it would also make the code cleaner). I'm fine with strings or string/bool, or even CSV for granularity.

One thing to consider is backwards compatibility, but I don't think there would be any harm even if we said that, say, `apt` alone is not supported anymore and either fall to not cleaning at all or cleaning more.

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
mgorny: Done. Also it occurred to me that we can add them directly in the script now, so we don't have to repeat the same pattern in all templates.

conda_smithy/templates/free_disk_space.sh.tmpl
mgorny: `rm -f` doesn't fail if the argument doesn't exist. It only fails if actual removal failed (e.g. because of I/O error).

conda_smithy/templates/free_disk_space.sh.tmpl
mgorny: Hmm, given that we're filtering the list already, perhaps we should instead check if `BROWSERS_TO_REMOVE` didn't end up empty?

conda_smithy/schema.py
mgorny: Uh, sorry, changed it prior to getting down to your comment. However, let's discuss the actual values first perhaps :-).

conda_smithy/templates/free_disk_space.sh.tmpl
mgorny: There's no `sudo` on macos, I'll fix that as a followup.

conda_smithy/templates/free_disk_space.sh.tmpl
jaimergp: That works too.

conda_smithy/schema.py
h-vetinari: Jaime 👍'd your [comment](https://github.com/conda-forge/conda-smithy/pull/2568#discussion_r3362194788) about changing away from list-as-value. So it appears that the next step would be to make a proposal how that would look like.

conda_smithy/schema.py
mgorny: Well, how about either:

- if we want real booleans, `false / true / "all"`
- if we just strings for fewer types: `"no" / "quick" / "all"`

?

h-vetinari: > "no" / "quick" / "all"

I like this 👍

h-vetinari: I'd maybe change "all" to "max". We're not deleting all things in the image after all, just the most we can. And "max" describes that more accurately, both in the sense of what can be deleted, as well as what we have actually implemented (or deemed worth implementing).

mgorny: `zero | min | max`?

/me hides.

h-vetinari: @jaimergp @beckermr @jakirkham opinions on this? My current preference is
```
free_disk_space: "no" | "quick" | "max"
```

jaimergp: I like that one ^

mgorny: Extra question: should we round existing values down or up? For example, `["apt"]` should go up to `quick` or down to `no`?

jaimergp: Are you asking whether `apt` is included in `quick` or `max`? 🤔 I guess that depends on how time consuming it is. It was included with `cache` before, right? In that sense, it'd be part of `quick`.

mgorny: No, I'm asking how to handle backwards compatibility with:

```yaml
azure:
  settngs_linux:
    free_disk_space: [apt]
```

mgorny:
Okay, updated to `"no" | quick | max`. Note that `"no"` sucks because of the Norway problem (it's being misinterpeted as boolean), so we probably want to change that to something else.

mgorny:
I've changed it to `none`. Please let me know if you prefer another term. I think we should avoid `no` because it's likely to be accidentally typed without quoting, and then break stuff.

conda_smithy/schema.py
h-vetinari: I would say: keep the logic simple. If that's not a big concern, then my preference would be: only the full set of apt,cache,docker gets mapped to max, everything else gets mapped to quick.

conda_smithy/schema.py
```diff
@@ -552,8 +552,8 @@ class WorkflowSettings(BaseModel):
 
     free_disk_space: Optional[
         Union[
-            Literal["no", "quick", "max"],
-            list[conditional_value(Union[Literal["no", "quick", "max"]], "no")],
+            Literal["none", "quick", "max"],
+            list[conditional_value(Union[Literal["none", "quick", "max"]], "none")],
```
h-vetinari: do we still need to (resp. want to) allow the list formulation here? Why not just demand a string literal?

h-vetinari (commented):
We could do `"skip"` instead of `"no"` or `"none`"

conda_smithy/schema.py
mgorny: Well, I made it even simpler: `"docker" in value` → `max`, non-empty `value` → `quick`. Is that okay, or should I change it?

conda_smithy/schema.py
mgorny: It's a list of `workflow_settings` dicts. The `value` itself is `Literal`.

conda_smithy/schema.py
h-vetinari: sounds good to me!

conda_smithy/schema.py
h-vetinari: Ah, right. Thx.

h-vetinari:
> We could do `"skip"` instead of `"no"` or `"none`"

I'd prefer "skip" to "none", if you and @jaimergp (or whoever else wants to opine) are OK with that. It reads more naturally IMO than `free_disk_space: "none"` (because, to me, we're qualifying the action of freeing, more so than the disk space in question).

```
free_disk_space: "skip" | "quick" | "max"
```

Otherwise this basically LGTM 👍

jaimergp:
I don't feel strongly about either. That one's fine.

mgorny:
Ok, changed to `skip` now.

mgorny:
Retested on https://github.com/conda-forge/polars-feedstock/pull/380. Should be ready.

jaimergp:
> Retested on https://github.com/conda-forge/polars-feedstock/pull/380

Can you enable it on all platforms? I want to check if linux-aarch64, osx and win work correctly.

mgorny:
Done. https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1535299&view=logs&j=f8c6bbb1-46dd-5b5d-cfc0-d4a86bc5bca7&t=1aab1f30-a8e4-5974-00ef-fa98aa7368a5 and others. Lemme know if you want to replace it with GHA afterwards.

mgorny:
Actually, here's GHA: https://github.com/mgorny/polars-feedstock/actions/runs/27221640076

h-vetinari (approved):
Thank you! Nice that we got all the major pieces in place now!

PR #2569, opened 2026-06-03
simplify create_pagefile.bat

h-vetinari:
Follow-up to #2562; there's no scenario where `"%PageFileDrive%" == ""` is ever true, so get rid of the condition for that; also delete a comment about not using delayed expansion, since that script _is_ using delayed expansion. Finally, break up a very long line, at least a little (the rest is a nested invocation, so leave that as-is).

CC @mgorny

### DIFF
--- a/conda_smithy/templates/create_pagefile.bat.tmpl
+++ b/conda_smithy/templates/create_pagefile.bat.tmpl
@@ -11,14 +11,13 @@ if /i "%CONDA_BLD_PATH:~0,2%" == "C:" (
     if exist D:\ set "PageFileDrive=D:"
 )
 
-:: Only run if PAGEFILE_SIZE is set; EntryPointPath needs to be set outside if-condition when not using EnableDelayedExpansion.
+:: Only run if PAGEFILE_SIZE is set
 if "%PAGEFILE_SIZE%" GTR "0" (
-    if not "%PageFileDrive%" == "" (
-        echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to %PAGEFILE_SIZE% GiB in %PageFileDrive%
-        REM Inspired by:
-        REM https://blog.danskingdom.com/allow-others-to-run-your-powershell-scripts-from-a-batch-file-they-will-love-you-for-it/
-        REM Drive-letter needs to be escaped in quotes
-        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& '%EntryPointPath%' -MinimumSize "%PAGEFILE_SIZE%GB" -MaximumSize "%PAGEFILE_SIZE%GB" -DiskRoot \"%PageFileDrive%\""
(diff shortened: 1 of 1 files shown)


conda_smithy/templates/create_pagefile.bat.tmpl
```diff
@@ -11,14 +11,13 @@ if /i "%CONDA_BLD_PATH:~0,2%" == "C:" (
     if exist D:\ set "PageFileDrive=D:"
 )
 
-:: Only run if PAGEFILE_SIZE is set; EntryPointPath needs to be set outside if-condition when not using EnableDelayedExpansion.
```
mgorny, line 14: To be honest, I've left that comment in case the file changed in the future, but I'm fine with removing it.

conda_smithy/templates/create_pagefile.bat.tmpl
```diff
@@ -11,14 +11,13 @@ if /i "%CONDA_BLD_PATH:~0,2%" == "C:" (
...
 if "%PAGEFILE_SIZE%" GTR "0" (
-    if not "%PageFileDrive%" == "" (
-        echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to %PAGEFILE_SIZE% GiB in %PageFileDrive%
-        REM Inspired by:
-        REM https://blog.danskingdom.com/allow-others-to-run-your-powershell-scripts-from-a-batch-file-they-will-love-you-for-it/
-        REM Drive-letter needs to be escaped in quotes
-        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& '%EntryPointPath%' -MinimumSize "%PAGEFILE_SIZE%GB" -MaximumSize "%PAGEFILE_SIZE%GB" -DiskRoot \"%PageFileDrive%\""
-    )
+    echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to %PAGEFILE_SIZE% GiB in %PageFileDrive%
+    REM Inspired by:
```
mgorny, line 17: BTW do we have an opinion on `REM` vs `::` here?

mgorny (approved):
Thanks!

conda_smithy/templates/create_pagefile.bat.tmpl
h-vetinari: `::` is nicer, but doesn't play well with indentation (of course it's possible to write the script; but it used to lead to warnings and/or weird output; perhaps things have improved...?).

My rule of thumb is: "if unindented, use `::`; if indented, use `REM`"

conda_smithy/templates/create_pagefile.bat.tmpl
h-vetinari: Since we're conditionally setting `PageFileDrive`, we can never remove delayed expansion here. 🤷

PR #2573, opened 2026-06-08
Add Depot.dev CI support

jaimergp:
Checklist
* [X] Added a ``news`` entry with any new deprecations added to the ``Deprecated`` section.

### DIFF
--- a/conda_smithy/github.py
+++ b/conda_smithy/github.py
@@ -601,6 +601,16 @@ def configure_github_app(
     app_slug_or_installation_id: str | int = None,
     remove: bool = False,
 ) -> None:
+    """
+    `app_slug_or_installation_id` can be a string or an integer. When a string
+    is passed, we will find the integer ID by matching the app slug.
+    For known installations (e.g. in conda-forge), you can find (and hardcode)
+    the ID with the following Github API call:
+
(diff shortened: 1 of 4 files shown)


conda_smithy/github.py
```diff
@@ -601,6 +601,16 @@ def configure_github_app(
...
 ) -> None:
+    """
+    `app_slug_or_installation_id` can be a string or an integer. When a string
+    is passed, we will find the integer ID by matching the app slug.
+    For known installations (e.g. in conda-forge), you can find (and hardcode)
+    the ID with the following Github API call:
+
+        /orgs/<YOUR-ORG>/installations
+
+    Annotate the .app_slug and .id fields.
```
beckermr: what does this mean?

beckermr (approved):

conda_smithy/github.py
jaimergp: The fields we need to write down are those two.

```suggestion
    The fields you need to write down are: `app_slug: str` and `id: int`. See
    the `enable_*_app()` and `disable_*_app()` functions above for examples.
```

jaimergp: Something like that?

PR #2575, opened 2026-06-08
fix: avoid false-positive maintainer lint on transient GitHub errors

chrisburr:
Checklist

The recipe-maintainer existence check treated any non-200 response as "maintainer does not exist". On the unauthenticated linting path this meant a GitHub rate limit (429), server error (5xx), or redirect caused a false-positive lint for valid maintainers.

Make the check tri-state: retry transient statuses with backoff, keep 404 and 3xx redirects (renamed/invalid usernames) flagged, and surface a new CF-008 hint instead of a lint when the check cannot be completed.

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -125,7 +125,7 @@
       "identifier": "CF-008",
       "category": "CF",
       "kind": "lint",
-      "added_in": "2026.6.0",
+      "added_in": "2026.6",
       "deprecated_in": "",
       "documentation": "The recipe maintainer existence check could not be completed.\n\nThis happens when GitHub cannot be reached or returns a transient error\n(e.g. a rate limit or a server error) while checking that a maintainer\nlisted in `extra.recipe-maintainers` is a valid Github user or\n`@conda-forge/*` team. Rather than risk a false-positive (or\nfalse-negative), the lint fails and asks for the check to be retried.",
       "message": "Could not verify that recipe maintainer ${team_or}\"${maintainer}\" exists due to a transient GitHub error. Please re-run the linter by commenting `@conda-forge-admin, please lint` on this pull request.",
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -121,6 +121,21 @@
       "path": ""
     },
     {
+      "name": "InconclusiveMaintainerCheck",
+      "identifier": "CF-008",
+      "category": "CF",
+      "kind": "lint",
+      "added_in": "2026.6.0",
+      "deprecated_in": "",
+      "documentation": "The recipe maintainer existence check could not be completed.\n\nThis happens when GitHub cannot be reached or returns a transient error\n(e.g. a rate limit or a server error) while checking that a maintainer\nlisted in `extra.recipe-maintainers` is a valid Github user or\n`@conda-forge/*` team. Rather than risk a false-positive (or\nfalse-negative), the lint fails and asks for the check to be retried.",
+      "message": "Could not verify that recipe maintainer ${team_or}\"${maintainer}\" exists due to a transient GitHub error. Please re-run the linter by commenting `@conda-forge-admin, please lint` on this pull request.",
+      "examples": [
+        "Could not verify that recipe maintainer \"some-user\" exists due to a transient GitHub error. Please re-run the linter by commenting `@conda-forge-admin, please lint` on this pull request.",
+        "Could not verify that recipe maintainer team \"@conda-forge/some-team\" exists due to a transient GitHub error. Please re-run the linter by commenting `@conda-forge-admin, please lint` on this pull request."
+      ],
+      "path": "recipe/(meta|recipe).yaml"
+    },
+    {
       "name": "NoDuplicateKeys",
       "identifier": "FC-001",
       "category": "FC",
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -1,7 +1,9 @@
 import copy
 import json
+import logging
 import os
 import sys
+import time
 from collections.abc import Mapping
 from functools import lru_cache
 from glob import glob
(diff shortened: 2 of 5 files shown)


conda_smithy/linter/messages/conda_forge.py
```diff
@@ -44,6 +44,43 @@ def examples(cls) -> list[Self]:
         ]
 
 
+@dataclass(kw_only=True)
+class MaintainerUnverified(LinterMessage):
```
jaimergp:
```suggestion
class InconclusiveMaintainerCheck(LinterMessage):
```

Thinking of the case when class names are used as identifiers, `MaintainerUnverified` sounds like we didn't verify their identity or something like that.

conda_smithy/lint_recipe.py
```diff
@@ -569,19 +654,23 @@ def run_conda_forge_specific(
...
+            lints.append(
+                msg.cf.MaintainerMissing(
+                    maintainer=maintainer, path=recipe_fname
+                ).as_string()
+            )
+        elif exists is None:
+            # the existence check could not be completed (e.g. GitHub rate
+            # limit); surface a hint instead of a false-positive lint
+            hints.append(
+                msg.cf.MaintainerUnverified(
```
jaimergp:
```suggestion
                msg.cf.InconclusiveMaintainerCheck(
```

jaimergp:
<3 Love it, thanks! Just one comment about the name of the class.

jaimergp:
Oh, and the tests are complaining about the class not appearing in order in the file (when sorting by identifier).

chrisburr:
From @beckermr on Zulip:

> Oh I see what happened. My 0.02 is that we should fail the linter and request people to retry but that is not a hill I will die on.

So this would mean making it an error and telling people the admin command to rerun the linter? I'm not opposed to doing that.

beckermr:
So again I don't care either way, but I do think having people relint is better in the end.

beckermr (approved):

news/maintainer-lint-transient.rst
```diff
@@ -0,0 +1,28 @@
...
+
+* <news item>
+
+**Changed:**
+
+* The recipe-maintainer existence lint no longer treats a transient GitHub
+  failure (rate limit, server error, or network error) as "maintainer does not
+  exist". Such checks are now retried and, if still inconclusive, the lint fails
+  with a message (``CF-008``) asking for the linter to be re-run, instead of
+  producing a false-positive "does not exist" lint.
```
jaimergp:
```suggestion
  producing a false-positive "does not exist" lint. (#2575)
```

news/maintainer-lint-transient.rst
```diff
@@ -0,0 +1,28 @@
...
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed false-positive ``Recipe maintainer "..." does not exist`` lints caused
+  by unauthenticated GitHub rate limiting during linting.
```
jaimergp:
```suggestion
  by unauthenticated GitHub rate limiting during linting. (#2575)
```

jaimergp (approved):

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -154,3 +154,40 @@ class DeprecatedEnvironmentVariable(LinterMessage):
...
+    This happens when GitHub cannot be reached or returns a transient error
+    (e.g. a rate limit or a server error) while checking that a maintainer
+    listed in `extra.recipe-maintainers` is a valid Github user or
+    `@conda-forge/*` team. Rather than risk a false-positive (or
+    false-negative), the lint fails and asks for the check to be retried.
+    """
+
+    kind = "lint"
+    identifier = "CF-008"
+    added_in = "2026.6.0"
```
h-vetinari: Hm; this is a bit tangential to the PR, but since it's the first time I notice it, I thought I'd comment on it. In contrast to our previous version scheme, it's hard to predict the version with CalVer, which makes populating `added_in:` non-obvious. I'm not sure the `.0` is all that useful, TBH.

Variants I could think of:
1. use `yyyy.m.0` (what this PR does)-- not a valid CalVer date, and can never match an actual smithy version
1. drop `.0` -- fuzzier and perhaps easier to map to concrete versions (but still not an actual version)
1. use the current date when adding new checks -- more obvious provenance (but also not an actual version)
1. do an update-job before release -- more work, but could reflect actual smithy version.

All of the above satisfy that `conda install conda-smithy>={ver}` will give a new enough smithy, assuming there's been a release already containing the new warning. Only 2. & 4. would work with `conda-smithy ={ver}` and only 4. would work with `conda-smithy =={ver}`.

I think 4. is too much work (and something we'd likely get wrong). I can live with all of 1./2./3., though I think the easiest would be 2.

h-vetinari (approved):
Thanks a lot for this. Given GH's increased flakiness, it's good to have this. As a concrete example, I've had 4-5 spurious comments about missing maintainers while working on a PR yesterday.

conda_smithy/linter/messages/conda_forge.py
jaimergp: Yea I was thinking about this too. It's something we can always retrofix anyway. I lean towards (2) as well, but the actual fix would be to put a placeholder `"next"` or something so the release ceremonies replace it with the actual tag.

... or we remove it altogether because maybe that's not useful enough for the complexity required.

h-vetinari: > but the actual fix would be to put a placeholder `"next"` or something so the release ceremonies replace it with the actual tag.

That's what I meant by 4.



> ... or we remove it altogether because maybe that's not useful enough for the complexity required.

I think overall, I'd prefer 1./2./3. to not filling in the `added_in:` stuff for new warnings. And as you say, we can still do retro-active fixes (in fact, I've been considering whether I want to fill in all the <3.56 values on some rainy day...)

beckermr: Let's do 2

conda_smithy/data/linter-messages.json
```diff
@@ -120,6 +120,21 @@
       "examples": [],
       "path": ""
     },
+    {
+      "name": "InconclusiveMaintainerCheck",
+      "identifier": "CF-008",
+      "category": "CF",
+      "kind": "lint",
+      "added_in": "2026.6.0",
```
beckermr:
```suggestion
      "added_in": "2026.6",
```

conda_smithy/linter/messages/conda_forge.py
beckermr:
```suggestion
    added_in = "2026.6"
```

PR #2566, opened 2026-05-30
Implement rattler-build debug mode for build-locally.py --debug

hmaarrfk:
CC: @wolfv i think this should add the capability to debug v1 recipes with docker and build-locally.py 

<details><summary>Claude's draft</summary>

Previously, running `python build-locally.py --debug` on a feedstock that uses `conda_build_tool: rattler-build` (v1 `recipe.yaml`) only printed "rattler-build currently doesn't support debug mode" and did nothing useful, losing the interactive local-debugging workflow that classic conda-build recipes have via `conda debug`.

rattler-build >=0.61 ships a `debug` subcommand that is the direct equivalent of `conda debug`: `rattler-build debug setup` resolves dependencies, fetches and patches sources, installs the build/host environments and writes work/conda_build.sh + work/build_env.sh, then stops before building; `rattler-build debug shell` opens an interactive shell with that environment activated.

This wires the rattler-build arm of the `BUILD_WITH_CONDA_DEBUG` branch in both build_steps.sh.tmpl (Linux/Docker) and run_osx_build.sh.tmpl (macOS) to call `rattler-build debug setup` then `rattler-build debug shell`, mirroring the conda-build `conda debug ... ; /bin/bash` behavior. A consistent CONDA_BLD_PATH is exported so setup and shell agree on the rattler-build-log.txt location, BUILD_OUTPUT_ID maps to `--output-name`, and the build-locally.py `--debug` help text is made tool-neutral.

</details>

Checklist
* [x] Added a ``news`` entry with any new deprecations added to the ``Deprecated`` section.
* [x] Regenerated schema JSON if schema altered (`python -m conda_smithy.schema`)
* [x] Regenerated linter documentation if any rules were added, removed or modified (`python -m conda_smithy.linter.messages`)

PR #2576, opened 2026-06-10
Drop mambabuild from `--debug` workflow; document divergence of build options

h-vetinari:
Follow-up to #2566. I strongly doubt that mambabuild/boa support is still functional, 
https://github.com/conda-forge/conda-smithy/blob/0c86c33cd15fbca34889c35fee5cd1aaba31aeac/conda_smithy/schema.py#L45-L46
much less that it'd work by calling `conda debug`.
https://github.com/conda-forge/conda-smithy/blob/e5c3c27237a815f92a10c459a42fe6310c1345ff/conda_smithy/templates/build_steps.sh.tmpl#L134
(as opposed to `{{ BUILD_CMD }} debug`, which would at least have a theoretical chance of working).

Deprecating and removing mambabuild support can happen independently, but given that this code clearly never worked, at least we can disable it in the template for now.

It's a pity that the `--extra-meta` interface is different between conda-build and rattler-build, otherwise there'd also be a lot of common ground in
https://github.com/conda-forge/conda-smithy/blob/e5c3c27237a815f92a10c459a42fe6310c1345ff/conda_smithy/templates/build_steps.sh.tmpl#L152-L166

### DIFF
--- a/conda_smithy/templates/build_steps.sh.tmpl
+++ b/conda_smithy/templates/build_steps.sh.tmpl
@@ -127,7 +127,7 @@ if [[ -f "${FEEDSTOCK_ROOT}/LICENSE.txt" ]]; then
 fi
 
 if [[ "${BUILD_WITH_CONDA_DEBUG:-0}" == 1 ]]; then
-    {%- if conda_build_tool != "rattler-build" %}
+    {%- if conda_build_tool.startswith("conda-build") %}
     if [[ "x${BUILD_OUTPUT_ID:-}" != "x" ]]; then
         EXTRA_CB_OPTIONS="${EXTRA_CB_OPTIONS:-} --output-id ${BUILD_OUTPUT_ID}"
     fi
@@ -137,7 +137,7 @@ if [[ "${BUILD_WITH_CONDA_DEBUG:-0}" == 1 ]]; then
 
     # Drop into an interactive shell
     /bin/bash
-    {%- else %}
+    {%- elif conda_build_tool == "rattler-build" %}
     export CONDA_BLD_PATH="${CONDA_BLD_PATH:-${FEEDSTOCK_ROOT}/build_artifacts}"
     rattler-build debug setup \
         --recipe "${RECIPE_ROOT}" \
--- a/conda_smithy/templates/build_steps.sh.tmpl
+++ b/conda_smithy/templates/build_steps.sh.tmpl
@@ -134,10 +134,12 @@ if [[ "${BUILD_WITH_CONDA_DEBUG:-0}" == 1 ]]; then
     #   - --clobber-file vs. none
     #   - none vs. --target-platform
     {%- if conda_build_tool.startswith("conda-build") %}
-    conda debug "${RECIPE_ROOT}" -m "${CI_SUPPORT}/${CONFIG}.yaml" \
+    conda debug \
+        "${RECIPE_ROOT}" \
+        -m "${CI_SUPPORT}/${CONFIG}.yaml" \
         ${EXTRA_CB_OPTIONS:-} \
-        --clobber-file "${CI_SUPPORT}/clobber_${CONFIG}.yaml"
         ${BUILD_OUTPUT_ID:+--output-id "${BUILD_OUTPUT_ID}"} \
+        --clobber-file "${CI_SUPPORT}/clobber_${CONFIG}.yaml"
 
     # Drop into an interactive shell
     /bin/bash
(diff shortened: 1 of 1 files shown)


conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -127,23 +127,23 @@ if [[ -f "${FEEDSTOCK_ROOT}/LICENSE.txt" ]]; then
 fi
 
 if [[ "${BUILD_WITH_CONDA_DEBUG:-0}" == 1 ]]; then
-    {%- if conda_build_tool != "rattler-build" %}
+    # options that are shared across build backends
     if [[ "x${BUILD_OUTPUT_ID:-}" != "x" ]]; then
         EXTRA_CB_OPTIONS="${EXTRA_CB_OPTIONS:-} --output-id ${BUILD_OUTPUT_ID}"
```
jaimergp: This one cannot be shared. `conda-build` uses `--output-id` but `rattler-build` uses `--output-name`.

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -127,17 +127,21 @@ if [[ -f "${FEEDSTOCK_ROOT}/LICENSE.txt" ]]; then
...
+    #   - 1 step (conda debug + manually open shell) vs. 2 step (rb debug {setup, shell})
+    #   - recipe is positional vs. --recipe "${RECIPE_ROOT}"
+    #   - --output-id vs. --output-name
+    #   - --clobber-file vs. none
+    #   - none vs. --target-platform
+    {%- if conda_build_tool.startswith("conda-build") %}
     conda debug "${RECIPE_ROOT}" -m "${CI_SUPPORT}/${CONFIG}.yaml" \
         ${EXTRA_CB_OPTIONS:-} \
         --clobber-file "${CI_SUPPORT}/clobber_${CONFIG}.yaml"
+        ${BUILD_OUTPUT_ID:+--output-id "${BUILD_OUTPUT_ID}"} \
```
jaimergp, line 141:
```suggestion
        --clobber-file "${CI_SUPPORT}/clobber_${CONFIG}.yaml" \
        ${BUILD_OUTPUT_ID:+--output-id "${BUILD_OUTPUT_ID}"}
```

h-vetinari: Gah, thanks for catching this

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -127,43 +127,59 @@ if [[ -f "${FEEDSTOCK_ROOT}/LICENSE.txt" ]]; then
...
     rattler-build debug shell
     {%- endif %}
 else
+    # differences between conda-build vs. rattler-build
+    #   - recipe is positional vs. --recipe "${RECIPE_ROOT}"
+    #   - --suppress-variables vs. none
+    #   - --clobber-file vs. none
+    #   - none vs. --target-platform
+    #   - --extra-meta a=b c=d vs. --extra-meta a=b --extra-meta c=d
     {%- if conda_build_tool != "rattler-build" %}
```
jaimergp, line 167: Don't you want the same `.startswith()` logic here? 🤔 If so, the `else` comment also applies to this block.

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -127,43 +127,59 @@ if [[ -f "${FEEDSTOCK_ROOT}/LICENSE.txt" ]]; then
...
         --recipe "${RECIPE_ROOT}" \
         -m "${CI_SUPPORT}/${CONFIG}.yaml" \
-        --target-platform "${HOST_PLATFORM}" \
+        ${EXTRA_CB_OPTIONS:-} \
         ${BUILD_OUTPUT_ID:+--output-name "${BUILD_OUTPUT_ID}"} \
-        ${EXTRA_CB_OPTIONS:-}
+        --target-platform "${HOST_PLATFORM}"
 
     rattler-build debug shell
     {%- endif %}
```
jaimergp, line 159: We should probably add an `else` branch reporting that whatever future tool being used that is not being evaluated is not supported in debug mode.
```suggestion
    {%- else %}
    echo "Build tool {{ conda_build_tool }} is not currently supported in debug mode"
    exit 1
    {%- endif %}
```

conda_smithy/templates/build_steps.sh.tmpl
h-vetinari: I was going by the values from
https://github.com/conda-forge/conda-smithy/blob/0c86c33cd15fbca34889c35fee5cd1aaba31aeac/conda_smithy/schema.py#L38-L49

where there are various things starting with conda-build, but only one for rattler. We could add an else-branch here though

jaimergp (approved):

h-vetinari:
Thanks for the review! :)

PR #2577, opened 2026-06-10
Fix configuring swap on Namespace Linux runners

mgorny:
Checklist

Namespace Linux runners default to using an overlay-backend filesystem that does not support swap files.  To use swap files, one needs to enable mounting scratch space using an appropriate action label.  Detect using Namespace along with enabled swap, and adjust labels and swap file path accordingly.

See https://namespace.so/docs/reference/github-actions/runner-configuration#setting-up-swap-space
Thanks to @jaimergp for figuring this out.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -1888,6 +1888,7 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
 
         data["gha_runs_on"] = []
         with_gpu = False
+        on_namespace = False
         for label in labels:
             if label.startswith("cirun-"):
                 # Patch Cirun runners to add some extra debug info
@@ -1896,6 +1897,10 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
                 # Having 'gpu' in one label name is enough to trigger
                 # the extra docker args needed on Linux
                 with_gpu = True
+            if label.startswith("namespace-profile-"):
+                on_namespace = True
+                if "-on-linux-" in label:
+                    label += ";container.privileged=true;container.mount-scratch=true"
             data["gha_runs_on"].append(label)
 
         workflow_settings = get_workflow_settings(
@@ -1906,9 +1911,6 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
             "windows-2022",
             "windows-2025",
         }.intersection(data["gha_runs_on"])
-        on_namespace = any(
-            x.startswith("namespace-profile-") for x in data["gha_runs_on"]
-        )
         fill_workflow_settings_defaults(
             workflow_settings,
             "github_actions",
--- a/conda_smithy/templates/create_pagefile.sh.tmpl
+++ b/conda_smithy/templates/create_pagefile.sh.tmpl
@@ -7,6 +7,9 @@ PAGEFILE_SIZE=${1}
 [[ ${PAGEFILE_SIZE} -le 0 ]] && exit
 
 SWAPFILE=/swapfile
+if [[ ${GHA_RUNS_ON} == *namespace-profile-* ]]; then
+    SWAPFILE=/namespace/scratch/swapfile
+fi
 # If there is already a swapfile, disable it and remove it
 if swapon --show | grep -q "^${SWAPFILE}"; then
     sudo swapoff "${SWAPFILE}" || true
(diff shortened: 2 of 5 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -1896,6 +1897,10 @@ def _github_actions_specific_setup(jinja_env, forge_config, forge_dir, platform)
                 # Having 'gpu' in one label name is enough to trigger
                 # the extra docker args needed on Linux
                 with_gpu = True
+            if label.startswith("namespace-profile-"):
+                on_namespace = True
+                if "-on-linux-" in label:
+                    label += ";container.privileged=true;container.mount-scratch=true"
```
jaimergp, line 1903: Should we check whether the user has already provided some "features" manually in CBC?

h-vetinari: In a perfect world perhaps, but since this isn't documented anywhere (on our side at least), I'm doubtful that people would just start adding stuff to `github_action_labels` and then expect it to work. I'm not against guarding against this case, but a bit doubtful whether it's worth the effort.

mgorny: If we did, what would do differently here?

h-vetinari: We'd have to check the contents of `github_action_labels` for semicola, split on the first one we find, then check if any of content after the first semicolon already covers what we're adding here.

jaimergp:
```suggestion
                if "-on-linux-":
                    features = dict.fromkeys(label.split(";"))
                    features.setdefault("container.privileged=true", None)
                    features.setdefault("container.mount-scratch=true", None)
                    label = ";".join(features)
```

h-vetinari: That doesn't cover the case where someone sets `container.privileged=false`. But we could do something like
```python
                if "-on-linux-" in label and ";" in label:
                    runner, rest = label.split(";", maxsplit=1)
                    try:
                        features = dict([tuple(x.split("=")) for x in rest.split(";")])
                    except ValueError:
                        raise ValueError(f"a label with properties needs to be of the form `label;a=b;c=d`; received:\n{label}")
                    features.setdefault("container.privileged", "true")
                    features.setdefault("container.mount-scratch", "true")
                    label = ";".join([runner] + [f"{k}={v}" for k, v in features.items()])
```

jaimergp: Don't we need to override `privileged` and `mount-scratch` regardless, because we need them as `true` for swapfile to work?

h-vetinari: Sure, but then it's just a question of using `features["container.privileged"] = "true"` instead of `features.setdefault()`. The problem with using `"container.privileged=true"` as the dict key is that it will not conflict with `"container.privileged=false"`

mgorny: But is the added complexity really worth the effort? I dare say using Namespace will be rare enough, and having people specify custom properties even rarer. Then, having people specify label that's in conflict with swapfile requirements *and* enable swapfile is a corner case of a corner case with no good solution.

h-vetinari: I agree, hence my

> I'm not against guarding against this case, but a bit doubtful whether it's worth the effort.

above

h-vetinari: OTOH, 9 lines of code (incl. basic error handling) may be within the realm of "affordable" effort

jaimergp: I regret having asked hah but now the code is there so 🤷 As you wish, @mgorny, you are not wrong about the corner casey aspect of this. I'll approve regardless.

jaimergp (approved):

conda_smithy/configure_feedstock.py
mgorny: I just did a quick test and if you end up with `...;;container.privileged=false;container.mount-scratch=false;container.privileged=true;container.mount-scratch=true`, things work as expected (i.e. `true` takes precedence). So I'd say let's just leave things as-is. Worst case, we're getting ugly-ish labels but they're still correct (for a definition of "correct").

jaimergp: So last one wins, cool. Thanks for checking.

PR #2579, opened 2026-06-11
Split GitHub Actions workflows into smaller steps

### LINKED ISSUE #2571: Break out build preparations in to separate step on GHA
After https://github.com/conda-forge/conda-smithy/pull/2562, we currently treat the split into separate steps differently between platforms; azure shows a separate step

[this should be a screenshot, but image upload is broken right now?]

while GHA lumps everything together under the build step

[this should be a screenshot, but image upload is broken right now?]

That's directly visible from the GHA template, where both scripts are called together in the same step
https://github.com/conda-forge/conda-smithy/blob/7ed6300656d706b62cd98734494401cf1e7577cc/conda_smithy/templates/github-actions.yml.tmpl#L218-L219

There already exists a preparatory step that does on windows
https://github.com/conda-forge/conda-smithy/blob/7ed6300656d706b62cd98734494401cf1e7577cc/conda_smithy/templates/github-actions.yml.tmpl#L182-L185
resp. on linux
https://github.com/conda-forge/conda-smithy/blob/7ed6300656d706b62cd98734494401cf1e7577cc/conda_smithy/templates/github-actions.yml.tmpl#L62-L64

I'd like to break this kind of processing out of the main build step, because the main build will often generate 1000s of lines of codes quickly, which makes it infeasible (given GHA's constraints) to see the beginning of the log until the job failed.

In that way, trivial questions like "did the swapfile get created correctly?" become unnecessarily hard to answer, given that the job will likely fail much later if that wasn't the case actually. In contrast, the logs of completed steps are usually[^1] immediately visible.

[^1]: I say usually, because non-live logs in GHA runs have been broken for a few weeks.

I'm not opposed to lumping everything together in one setup step, but OTOH, these steps don't cost anything, so we might as well do separate ones?

We, may also take stuff like the `binfmt_misc` configuration into such a setup step (see below). Ideally the build step is then just a couple of env vars and the build script invocation.
https://github.com/conda-forge/conda-smithy/blob/7ed6300656d706b62cd98734494401cf1e7577cc/conda_smithy/templates/github-actions.yml.tmpl#L121-L142
Sidenote: the `::group::` / `::endgroup::` above are mismatched in terms of if-branches; more to the point, AFAICT they don't do anything.

CC @mgorny

mgorny:
Checklist

Fixes #2571

Split binfmt_misc and page file setup into separate GHA steps.  Make the latter conditional to swap being enabled for the particular job, and therefore remove now-redundant checks from the scripts.

### DIFF
--- a/conda_smithy/templates/create_pagefile.bat.tmpl
+++ b/conda_smithy/templates/create_pagefile.bat.tmpl
@@ -11,13 +11,10 @@ if /i "%CONDA_BLD_PATH:~0,2%" == "C:" (
     if exist D:\ set "PageFileDrive=D:"
 )
 
-:: Only run if PAGEFILE_SIZE is set
-if "%PAGEFILE_SIZE%" GTR "0" (
-    echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to %PAGEFILE_SIZE% GiB in %PageFileDrive%
-    REM Inspired by:
-    REM https://blog.danskingdom.com/allow-others-to-run-your-powershell-scripts-from-a-batch-file-they-will-love-you-for-it/
-    REM Drive-letter needs to be escaped in quotes
-    PowerShell -NoProfile -ExecutionPolicy Bypass ^
-        -Command "& '%EntryPointPath%' -MinimumSize "%PAGEFILE_SIZE%GB" -MaximumSize "%PAGEFILE_SIZE%GB" -DiskRoot \"%PageFileDrive%\""
-)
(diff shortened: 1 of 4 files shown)


mgorny:
Demo in https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/actions/runs/27361767756/job/80850686837. I've deliberately left `linux_aarch64` jobs with pagefile disabled.

h-vetinari (approved):
Very nice! :)

Does the sequence and granularity of the steps on azure match this? If not, can we synchronize the azure steps with what GHA is doing as of this PR?

mgorny:
Almost. binfmt_misc setup was out of order, so I just moved it to where AzP are doing it.

h-vetinari (approved):
Thank you!

PR #2580, opened 2026-06-11
Add hint: noarch python recipes should test python_min and latest Python

killua156:
### Description

Adds a v1-only hint (`R1-004`): when a `noarch: python` recipe's Python test runs against a single Python version, suggest

```yaml
tests:
  - python:
      python_version:
        - ${{ python_min }}.*
        - "*"
```

`noarch: python` packages install on every Python `>= python_min`, but most staged-recipes only test `python_min`, leaving the newest interpreter unverified (removed stdlib modules, deprecations turned errors, etc.). Reviewers already suggest this pattern by hand, e.g. https://github.com/conda-forge/staged-recipes/pull/33110#discussion_r3357913563 — this automates it.

Not hinted when:
- the `run` requirements cap Python's upper bound (the test env solves back to `python_min`, so a `"*"` entry would be redundant),
- the version list already includes `"*"` (incl. abi3-style conditional lists),
- the recipe isn't `noarch: python`, has no Python test, or is a v0 recipe.

Skippable via `linter.skip: [hint_noarch_python_test_latest]`. Includes tests, news entry, and regenerated `linter-messages.json`.

### Checklist

* [x] Added a `news` entry

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -909,6 +909,18 @@
       "path": "recipe/recipe.yaml"
     },
     {
+      "name": "PythonVersionIndependentTestLatest",
+      "identifier": "R1-005",
+      "category": "R1",
+      "kind": "hint",
+      "added_in": "2026.7",
+      "deprecated_in": "",
+      "documentation": "Python version-independent packages (e.g. abi3) are built once against\n`python_min` but install on every Python version at or above it, so tests\nshould cover both ends of that range.",
+      "message": "This package is Python version-independent (e.g. abi3): it is built once but installs on every Python version at or above `python_min`, yet the Python test only runs against a single version. Consider testing against both the minimum and the latest supported Python:\n```yaml\ntests:\n  - python:\n      python_version:\n        - ${{ python_min }}.*\n        - \"*\"\n```",
+      "examples": [],
+      "path": "recipe/recipe.yaml"
+    },
+    {
       "name": "MacOSDeploymentTargetRename",
       "identifier": "RC-000",
       "category": "RC",
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -897,6 +897,18 @@
       "path": "recipe/recipe.yaml"
     },
     {
+      "name": "NoarchPythonTestLatest",
+      "identifier": "R1-004",
+      "category": "R1",
+      "kind": "hint",
+      "added_in": "2026.6",
+      "deprecated_in": "",
+      "documentation": "`noarch: python` packages install on every Python version at or above\n`python_min`, so tests should cover both ends of that range.",
+      "message": "`noarch: python` packages install on every Python version at or above `python_min`, but the Python test only runs against a single version. Consider testing against both the minimum and the latest supported Python:\n```yaml\ntests:\n  - python:\n      python_version:\n        - ${{ python_min }}.*\n        - \"*\"\n```",
+      "examples": [],
+      "path": "recipe/recipe.yaml"
+    },
+    {
       "name": "MacOSDeploymentTargetRename",
       "identifier": "RC-000",
       "category": "RC",
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -901,7 +901,7 @@
       "identifier": "R1-004",
       "category": "R1",
       "kind": "hint",
-      "added_in": "2026.6",
+      "added_in": "2026.7",
       "deprecated_in": "",
       "documentation": "`noarch: python` packages install on every Python version at or above\n`python_min`, so tests should cover both ends of that range.",
       "message": "`noarch: python` packages install on every Python version at or above `python_min`, but the Python test only runs against a single version. Consider testing against both the minimum and the latest supported Python:\n```yaml\ntests:\n  - python:\n      python_version:\n        - ${{ python_min }}.*\n        - \"*\"\n```",
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -361,6 +361,67 @@ def hint_noarch_python_use_python_min(
         hints.append(msg.r.PythonMinPin(recommendations=recommendations).as_string())
 
 
+def _noarch_python_tests_cover_latest(tests_section, run_reqs):
+    """True if every python test covers the latest Python via a "*" entry,
+    or if the run requirements cap python's upper bound (making a latest-python
+    test entry redundant)."""
+    for req in flatten_v1_if_else(run_reqs or []):
+        if isinstance(req, str) and req.strip().split()[0] == "python" and "<" in req:
+            return True
+
+    for test in tests_section or []:
+        if not isinstance(test, Mapping) or "python" not in test:
+            continue
+        python_version = (test.get("python") or {}).get("python_version")
+        if isinstance(python_version, str):
+            python_version = [python_version]
+        if not isinstance(python_version, list):
+            python_version = []
+        # The latest-Python marker is the exact entry `"*"`. We match it by
+        # equality, so the `.*` in a min-pin like `${{ python_min }}.*` (or its
+        # rendered `3.10.*`) is never mistaken for it. (The v1 parser also
+        # forbids a bare `- *`, so `"*"` is always a quoted string here.)
+        if "*" not in flatten_v1_if_else(python_version):
+            return False
+    return True
+
+
+def hint_noarch_python_test_latest(
+    tests_section,
+    run_reqs,
+    outputs_section,
+    noarch_value,
+    recipe_version,
+    hints,
+):
+    if recipe_version != 1:
+        return
+
+    scopes = []
+    if outputs_section:
+        for output in outputs_section:
+            requirements = output.get("requirements", {})
+            output_run_reqs = (
+                requirements.get("run")
+                if isinstance(requirements, Mapping)
+                else requirements
+            )
+            scopes.append(
+                (
+                    output.get("tests"),
+                    output_run_reqs,
+                    output.get("build", {}).get("noarch"),
+                )
+            )
+    else:
+        scopes.append((tests_section, run_reqs, noarch_value))
+
+    for tests, run, noarch in scopes:
+        if noarch == "python" and not _noarch_python_tests_cover_latest(tests, run):
+            hints.append(msg.r.NoarchPythonTestLatest().as_string())
+            return
+
+
 def hint_space_separated_specs(
     requirements_section,
     test_section,
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -373,7 +373,7 @@ def _python_tests_cover_latest(tests_section, run_reqs):
     for test in tests_section or []:
         if not isinstance(test, Mapping) or "python" not in test:
             continue
-        python_version = (test.get("python") or {}).get("python_version")
+        python_version = test.get("python", {}).get("python_version", {})
         if isinstance(python_version, str):
             python_version = [python_version]
         if not isinstance(python_version, list):
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -377,10 +377,10 @@ def _noarch_python_tests_cover_latest(tests_section, run_reqs):
             python_version = [python_version]
         if not isinstance(python_version, list):
             python_version = []
-        # The latest-Python marker is the exact entry `"*"`. We match it by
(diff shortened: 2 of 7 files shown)


h-vetinari:
I don't think we want to make this into a lint; most simple packages work just fine when tested on one python version (c.f. a previous discussion on this topic in https://github.com/conda-forge/conda-forge.github.io/issues/2617).

I'm generally one of the more "testing-happy" people in conda-forge, so I'm not that hard to convince (and could imagine that we _might_ want to make this a hint), but others in the core-team have generally been concerned about wasting CI resources, and thus been against overly extensive testing.

Perhaps a middle-ground would be to document this pattern more visibly, while not recommending _all_ v1 python recipes do this. OTOH, if a package only does a couple `import foo` tests, then testing against a second python version using rattler-build is less than a minute of additional runtime, so definitely not an extreme increase either.

CC @conda-forge/core

beckermr:
For sure we should document this pattern if we have not.

We don't use a lot of extra CI resources with this change due to rattler-build's speed and the fact that noarch builds are very fast. 

I don't have strong opinions here though.

killua156:
One thing I want to raise separately though. If a recipe adds an explicit python_min that is already at or below the global python_min, it doesn't actually do anything, but it does become a maintenance problem later. Every time the global pin moves up, all those recipes need manual updates, which is a pain across hundreds of feedstocks.

So whatever we end up documenting, maybe it should also say to only set python_min when the package genuinely needs something newer than the global floor, and leave it out otherwise. That keeps recipes from going stale and avoids the cleanup work down the line.

cc:@conda-forge/core

killua156:
@ocefpaf please review!

h-vetinari:
You already have attention from two core members, please avoid spamming pings unnecessarily.

> One thing I want to raise separately though. If a recipe adds an explicit python_min that is already at or below the global python_min, it doesn't actually do anything, but it does become a maintenance problem later.

Why are we discussing this here? You already have a separate PR for that (#2581), which I've already reviewed favourably?

killua156:
sorry I got confused and commented it there. Thank you for your reviews and suggestions! @h-vetinari

isuruf:
> I don't think we want to make this into a lint; most simple packages work just fine when tested on one python version (c.f. a previous discussion on this topic in https://github.com/conda-forge/conda-forge.github.io/issues/2617).

This has been something that a few staged-recipes maintainers including me, @ocefpaf @matthewfeickert have been asking maintainers of new recipes. This formalizes the lint.

h-vetinari:
> This has been something that a few staged-recipes maintainers including me, @ocefpaf @matthewfeickert have been asking maintainers of new recipes. This formalizes the lint.

If so, that's great! No opposition from my side as I mentioned, I had just brought up the sentiment I had gotten in prior discussions on this.

isuruf:
> OTOH, if a package only does a couple import foo tests, then testing against a second python version using rattler-build is less than a minute of additional runtime, so definitely not an extreme increase either.

See https://github.com/prefix-dev/rattler-build/blob/9f7933b73d72f1d62b1152c9818ce39dd6a4b2a5/docs/testing.md?plain=1#L53-L59
This is only for import tests and pip checks. Therefore it's quite fast.

isuruf (approved):

conda_smithy/linter/messages/recipe.py
```diff
@@ -1218,4 +1218,28 @@ class RattlerBldBat(LinterMessage, _RecipeYamlMessage):
...
+@dataclass(kw_only=True)
+class NoarchPythonTestLatest(LinterMessage, _RecipeYamlMessage):
+    """
+    `noarch: python` packages install on every Python version at or above
+    `python_min`, so tests should cover both ends of that range.
+    """
+
+    kind = "hint"
+    identifier = "R1-004"
+    message = (
```
h-vetinari, line 1231:
```suggestion
    added_in: "2026.6"
    message = (
```

conda_smithy/data/linter-messages.json
```diff
@@ -896,6 +896,18 @@
       "examples": [],
       "path": "recipe/recipe.yaml"
     },
+    {
+      "name": "NoarchPythonTestLatest",
+      "identifier": "R1-004",
+      "category": "R1",
+      "kind": "hint",
+      "added_in": "<3.56",
```
h-vetinari:
```suggestion
      "added_in": "2026.6",
```

tests/test_lint_recipe.py
```diff
@@ -4228,6 +4228,159 @@ def test_lint_recipe_v1_python_min_in_python_version(text):
...
+    - python ${{ python_min }}.*
+  run:
+    - python >=${{ python_min }}
+
+tests:
+  - python:
+      imports:
+        - mypackage
+      python_version: ${{ python_min }}.*
+    """,
```
h-vetinari: Please also use `textwrap.dedent` for the parametrized entries here (like in #2581)

h-vetinari (commented):
A few nits, otherwise LGTM

conda_smithy/linter/hints.py
```diff
@@ -361,6 +361,63 @@ def hint_noarch_python_use_python_min(
...
+    for test in tests_section or []:
+        if not isinstance(test, Mapping) or "python" not in test:
+            continue
+        python_version = (test.get("python") or {}).get("python_version")
+        if isinstance(python_version, str):
+            python_version = [python_version]
+        if not isinstance(python_version, list):
+            python_version = []
+        if "*" not in flatten_v1_if_else(python_version):
+            return False
```
h-vetinari, line 386: How does this work with escaping? It should obviously not trigger for the `*` in `python ${{ python_min}}.*`. Thanksfully, the v1 parser doesn't allow bare `- *`, so at least one edge case falls away there.

killua156:
Thanks! Addressed the nits (rebased onto main, so the diff is refreshed):

- **`added_in = "2026.6"`** set on the `R1-004` message and regenerated into `linter-messages.json`.
- **`textwrap.dedent`** applied to all parametrized recipe entries.
- **Escaping / `*` matching**: good question. The check is an exact-equality membership test (`"*" in flatten_v1_if_else(python_version)`), not a substring or `endswith`, so the `.*` in `${{ python_min }}.*` (or its rendered `3.10.*`) is never treated as the latest-Python marker. The scalar `python_version: ${{ python_min }}.*` case (test `recipe-1`) asserts the hint *does* fire, which guards exactly this. And as you noted, the v1 parser forbids a bare `- *`, so `"*"` is always a quoted string. I added an inline comment spelling this out.

conda_smithy/linter/hints.py
killua156: Tested this directly against the linter to be sure. The check is exact-equality membership (`"*" in flatten_v1_if_else(python_version)`), not a substring/endswith, so a min-pin entry is never mistaken for the latest-* marker.

killua156:
@h-vetinari . suggested changes made! please review when possible :)

conda_smithy/linter/messages/recipe.py
h-vetinari: this is duplicated now, and I mistakenly suggested the wrong syntax (sorry about that). It needs to be removed now

conda_smithy/linter/hints.py
```diff
@@ -361,6 +361,67 @@ def hint_noarch_python_use_python_min(
...
+        python_version = (test.get("python") or {}).get("python_version")
+        if isinstance(python_version, str):
+            python_version = [python_version]
+        if not isinstance(python_version, list):
+            python_version = []
+        # The latest-Python marker is the exact entry `"*"`. We match it by
+        # equality, so the `.*` in a min-pin like `${{ python_min }}.*` (or its
+        # rendered `3.10.*`) is never mistaken for it. (The v1 parser also
+        # forbids a bare `- *`, so `"*"` is always a quoted string here.)
+        if "*" not in flatten_v1_if_else(python_version):
```
h-vetinari: This only works because
https://github.com/conda-forge/conda-smithy/blob/5b2bd15e9dac70c4ad6bc86a258b81ffdba6aed8/conda_smithy/linter/utils.py#L245
always returns a `list[str]`. If it ever returned a `str`, then `"*" in ...` becomes a substring-comparison, not an equality match within the list.

So let's document that
```suggestion
        # Check that the latest-Python marker is the exact entry `"*"`. Since
        # flatten_v1_if_else always returns a list, `"*" in ...` will never
        # (substring-)match for version pins like `${{ python_min }}.*`;
        # v1 also forbids bare `- *`, so we know it must be a string.
        if "*" not in flatten_v1_if_else(python_version):
```

conda_smithy/linter/hints.py
```diff
@@ -361,6 +361,67 @@ def hint_noarch_python_use_python_min(
...
+    or if the run requirements cap python's upper bound (making a latest-python
+    test entry redundant)."""
+    for req in flatten_v1_if_else(run_reqs or []):
+        if isinstance(req, str) and req.strip().split()[0] == "python" and "<" in req:
+            return True
+
+    for test in tests_section or []:
+        if not isinstance(test, Mapping) or "python" not in test:
+            continue
+        python_version = (test.get("python") or {}).get("python_version")
```
h-vetinari: This can fail if test sections have no `python_version:` at all. We're already linting against this for noarch recipes, but the linter should not fall on its face if that key is missing. Please add a fallback here, e.g.
```suggestion
        python_version = test.get("python", {}).get("python_version", {})
```
and add a test that absence of `python_version:` doesn't crash the linter.

conda_smithy/linter/messages/recipe.py
```diff
@@ -1218,4 +1218,30 @@ class RattlerBldBat(LinterMessage, _RecipeYamlMessage):
...
+class NoarchPythonTestLatest(LinterMessage, _RecipeYamlMessage):
+    """
+    `noarch: python` packages install on every Python version at or above
+    `python_min`, so tests should cover both ends of that range.
+    """
+
+    kind = "hint"
+    identifier = "R1-004"
+    added_in = "2026.6"
+    added_in: "2026.6"
```
h-vetinari:
```suggestion
```

conda_smithy/linter/messages/recipe.py
```diff
@@ -1218,4 +1218,30 @@ class RattlerBldBat(LinterMessage, _RecipeYamlMessage):
...
+@dataclass(kw_only=True)
+class NoarchPythonTestLatest(LinterMessage, _RecipeYamlMessage):
+    """
+    `noarch: python` packages install on every Python version at or above
+    `python_min`, so tests should cover both ends of that range.
+    """
+
+    kind = "hint"
+    identifier = "R1-004"
+    added_in = "2026.6"
```
```suggestion
    added_in = "2026.7"
```

news/noarch-python-test-latest.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* New hint ``R1-004``: ``noarch: python`` v1 recipes whose Python test only runs against a single Python version are now hinted to test against both ``${{ python_min }}.*`` and ``"*"`` (the latest supported Python), unless the ``run`` requirements cap the Python upper bound.
```
jaimergp:
```suggestion
* New hint ``R1-004``: ``noarch: python`` v1 recipes whose Python test only runs against a single Python version are now hinted to test against both ``${{ python_min }}.*`` and ``"*"`` (the latest supported Python), unless the ``run`` requirements cap the Python upper bound. (#2580)
```

conda_smithy/linter/messages/recipe.py
killua156: Done — removed the duplicate line.

conda_smithy/linter/messages/recipe.py
killua156: Done — removed.

conda_smithy/data/linter-messages.json
killua156: Now `2026.7` and regenerated to match the class.

tests/test_lint_recipe.py
killua156: Done, switched to `textwrap.dedent`.

conda_smithy/linter/hints.py
killua156: Right — relies on `flatten_v1_if_else` always returning a list.

conda_smithy/linter/hints.py
killua156: Missing `python_version` falls back to `[]`, so it won't crash.

conda_smithy/linter/messages/recipe.py
killua156: Done.

news/noarch-python-test-latest.rst
killua156: Done.

PR #2581, opened 2026-06-11
Add hint: remove python_min redefinitions that match the global pinning default

killua156:
### Description

Adds hint `R-051`: when a recipe redefines `python_min` to the same value as conda-forge's global pinning default, suggest removing the redundant override. Covers both formats:

- v1: `context.python_min`
- v0: `{% set python_min = "..." %}`

Per CFEP-25 the global pinning already provides `python_min`; redefining it to the same value is noise reviewers have to verify, and it silently overrides platforms where the default differs (e.g. win-arm64 uses a different floor). Overriding to a *higher* floor is intentional and not hinted.

The default value is fetched from the pinning feedstock's `conda_build_config.yaml` using the same cached, fail-quiet pattern as `hints.toml` (5s timeout, hourly cache, hint silently skipped if unreachable). Skippable via `linter.skip: [hint_redundant_python_min]`.

Includes tests (fetch mocked), news entry, and regenerated `linter-messages.json`.

Related: #2580 (another `python_min` hygiene hint).

### Checklist

* [x] Added a `news` entry

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -841,11 +841,13 @@
       "identifier": "R-052",
       "category": "R",
       "kind": "hint",
-      "added_in": "2026.6",
+      "added_in": "2026.7",
       "deprecated_in": "",
-      "documentation": "conda-forge's global pinning already provides the `python_min` variable,\nso recipes should not redefine it to the same value. Only override\n`python_min` when the package needs a higher floor than the global\ndefault.",
-      "message": "The recipe sets `python_min` to ${value}, which is identical to the default provided by conda-forge's global pinning. Remove the redefinition; it is redundant and unintentionally overrides platforms where the global default differs. Only set `python_min` when the package needs a higher floor than the global default.",
-      "examples": [],
+      "documentation": "conda-forge's global pinning already provides the `python_min` variable,\nso recipes should not redefine it to the same (or a lower) value. Doing so\nis redundant and unintentionally overrides platforms where the global\ndefault differs. Only override `python_min` when the package needs a\nhigher floor than the global default.",
+      "message": "The recipe sets `python_min` to ${value}, which is equal or lower than the default provided by conda-forge's global pinning. Please remove the redefinition.",
+      "examples": [
+        "The recipe sets `python_min` to 3.9, which is equal or lower than the default provided by conda-forge's global pinning. Please remove the redefinition."
+      ],
       "path": "recipe/(meta|recipe).yaml"
     },
     {
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -789,6 +789,18 @@
       "path": "recipe/(meta|recipe).yaml"
     },
     {
+      "name": "RedundantPythonMin",
+      "identifier": "R-052",
+      "category": "R",
+      "kind": "hint",
+      "added_in": "2026.6",
+      "deprecated_in": "",
+      "documentation": "conda-forge's global pinning already provides the `python_min` variable,\nso recipes should not redefine it to the same value. Only override\n`python_min` when the package needs a higher floor than the global\ndefault.",
+      "message": "The recipe sets `python_min` to ${value}, which is identical to the default provided by conda-forge's global pinning. Remove the redefinition; it is redundant and unintentionally overrides platforms where the global default differs. Only set `python_min` when the package needs a higher floor than the global default.",
+      "examples": [],
+      "path": "recipe/(meta|recipe).yaml"
+    },
+    {
       "name": "FormattedSelectors",
       "identifier": "R0-001",
       "category": "R0",
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -16,6 +16,7 @@ from conda_smithy.linter.utils import (
     find_local_config_file,
     flatten_v1_if_else,
     get_all_test_requirements,
+    get_global_pinning_python_min,
     is_selector_line,
 )
 from conda_smithy.utils import get_yaml
@@ -359,6 +360,25 @@ def hint_noarch_python_use_python_min(
 
     if recommendations:
         hints.append(msg.r.PythonMinPin(recommendations=recommendations).as_string())
+
+
+def hint_redundant_python_min(meta, recipe_text, recipe_version, hints):
+    if recipe_version == 1:
+        context = meta.get("context")
+        declared = context.get("python_min") if isinstance(context, Mapping) else None
+    else:
+        match = re.search(
+            r"""{%\s*set\s+python_min\s*=\s*["']([^"']+)["']""",
+            recipe_text or "",
+        )
+        declared = match.group(1) if match else None
+
+    if declared is None:
+        return
+
+    global_python_min = get_global_pinning_python_min()
+    if global_python_min is not None and str(declared) == global_python_min:
+        hints.append(msg.r.RedundantPythonMin(value=global_python_min).as_string())
 
 
 def hint_space_separated_specs(
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -377,7 +377,7 @@ def hint_redundant_python_min(meta, recipe_text, recipe_version, hints):
         return
 
     global_python_min = get_global_pinning_python_min()
-    if global_python_min is not None and str(declared) == global_python_min:
+    if global_python_min is not None and Version(str(declared)) <= Version(global_python_min):
         hints.append(msg.r.RedundantPythonMin(value=global_python_min).as_string())
 
 
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -9,6 +9,8 @@ from collections.abc import Generator, Mapping
 from glob import glob
 from typing import Any
 
+from conda.models.version import VersionOrder
+
 from conda_smithy.linter import conda_recipe_v1_linter
 from conda_smithy.linter import messages as msg
 from conda_smithy.linter.utils import (
@@ -377,8 +379,10 @@ def hint_redundant_python_min(meta, recipe_text, recipe_version, hints):
         return
 
     global_python_min = get_global_pinning_python_min()
-    if global_python_min is not None and Version(str(declared)) <= Version(global_python_min):
-        hints.append(msg.r.RedundantPythonMin(value=global_python_min).as_string())
+    if global_python_min is not None and VersionOrder(str(declared)) <= VersionOrder(
+        global_python_min
+    ):
+        hints.append(msg.r.RedundantPythonMin(value=str(declared)).as_string())
 
 
 def hint_space_separated_specs(
--- a/conda_smithy/linter/messages/recipe.py
+++ b/conda_smithy/linter/messages/recipe.py
@@ -1041,22 +1041,25 @@ class DuplicateRecipes(LinterMessage, _AnyRecipeMessage):
 class RedundantPythonMin(LinterMessage, _AnyRecipeMessage):
     """
     conda-forge's global pinning already provides the `python_min` variable,
-    so recipes should not redefine it to the same value. Only override
-    `python_min` when the package needs a higher floor than the global
-    default.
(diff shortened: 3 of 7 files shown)


conda_smithy/linter/utils.py
```diff
@@ -242,6 +242,36 @@ def load_linter_toml_metdata_internal(time_salt):
...
 
+def get_global_pinning_python_min():
+    # ensure we refresh the cache every hour
+    ttl = 3600
+    time_salt = int(time.time() / ttl)
+    return _get_global_pinning_python_min_internal(time_salt)
+
+
+@lru_cache(maxsize=1)
+def _get_global_pinning_python_min_internal(time_salt):
```
h-vetinari: Rather than duplicating `load_linter_toml_metadata_internal{,_internal}`, I'd like to see something like
```python
# cache size should be >= number of urls in use; old epochs are never needed again
@lru_cache(maxsize=5)
def _try_fetch_url_content(url: str, epoch_hour: int) -> Optional[str]:
    """private helper for _try_fetch_url_content_cached"""
    try:
        payload = requests.get(url, timeout=5)
    except Timeout:
        return None
    if payload.status_code != 200:
        # too bad, but not important enough to throw an error;
        # linter will rerun on the next commit anyway
        return None
    return payload.content.decode("utf-8")

def _try_fetch_url_content_cached(url: str) -> Optional[str]:
    """self-limited to update only once per hour"""
    epoch_hour = int(time.time() / 3600)  # time.time() is in seconds
    return _try_fetch_url_content(url, epoch_hour)

def load_linter_toml_metadata() -> Optional[str]:
    url = "https://raw.githubusercontent.com/conda-forge/conda-forge-pinning-feedstock/main/recipe/linter_hints/hints.toml"
    if (hints_toml_str := _try_fetch_url_content_cached(url)) is None:
        return None
    return tomllib.loads(hints_toml_str)

def get_global_pinning_python_min() -> Optional[str]:
    url = "https://raw.githubusercontent.com/conda-forge/conda-forge-pinning-feedstock/main/recipe/conda_build_config.yaml"
    if (pinning_req := _try_fetch_url_content_cached(url)) is None:
        return None
    ... # the rest
```

tests/test_lint_recipe.py
```diff
@@ -4216,6 +4216,106 @@ def test_lint_recipe_v1_python_min_in_python_version(text):
...
+        (
+            "recipe.yaml",
+            """
+context:
+  python_min: "3.10"
+
+package:
+  name: mypackage
+  version: 1.0.0
+    """,
```
h-vetinari: Please use `textwrap.dedent` to keep indentation for the parametrization

conda_smithy/data/linter-messages.json
```diff
@@ -776,6 +776,18 @@
       "examples": [],
       "path": "recipe/(meta|recipe).yaml"
     },
+    {
+      "name": "RedundantPythonMin",
+      "identifier": "R-051",
+      "category": "R",
+      "kind": "hint",
+      "added_in": "<3.56",
```
h-vetinari: This should be (c.f. discussion in #2575)
```suggestion
      "added_in": "2026.6",
```

h-vetinari (commented):
This hint makes sense, thanks for the PR!

conda_smithy/linter/messages/recipe.py
```diff
@@ -1022,6 +1022,27 @@ class UsePyPIOrg(LinterMessage, _AnyRecipeMessage):
...
+class RedundantPythonMin(LinterMessage, _AnyRecipeMessage):
+    """
+    conda-forge's global pinning already provides the `python_min` variable,
+    so recipes should not redefine it to the same value. Only override
+    `python_min` when the package needs a higher floor than the global
+    default.
+    """
+
+    kind = "hint"
+    identifier = "R-051"
```
h-vetinari:
```suggestion
    identifier = "R-052"
    added_in = "2026.6"
```

killua156:
Thanks for the review! Addressed all four points (rebased onto main, so the diff is updated):

- **Shared URL fetcher**: refactored to `_try_fetch_url_content(url, epoch_hour)` + `_try_fetch_url_content_cached(url)` as you sketched, and rewrote both `load_linter_toml_metadata` and `get_global_pinning_python_min` to use it. Pulled the common pinning-feedstock base into a `PINNING_FEEDSTOCK_RAW` constant so the two URLs don't repeat it; happy to inline it if you'd rather keep the full URLs in each function.
- **`textwrap.dedent`** applied to the parametrized recipes.
- **identifier → `R-052`** (R-051 is now taken by `DuplicateRecipes` on main after the rebase).
- **`added_in = "2026.6"`** set on the class and regenerated into `linter-messages.json`, per your suggestion and the #2575 discussion. Heads-up: `DuplicateRecipes` (merged after that discussion) uses the full-date `2026.6.14` form — let me know if you'd prefer I match that for consistency.

conda_smithy/linter/hints.py
```diff
@@ -361,6 +362,25 @@ def hint_noarch_python_use_python_min(
...
+            r"""{%\s*set\s+python_min\s*=\s*["']([^"']+)["']""",
+            recipe_text or "",
+        )
+        declared = match.group(1) if match else None
+
+    if declared is None:
+        return
+
+    global_python_min = get_global_pinning_python_min()
+    if global_python_min is not None and str(declared) == global_python_min:
```
isuruf: We want to hint when it is less than the supported minimum too.
```suggestion
    if global_python_min is not None and Version(str(declared)) <= Version(global_python_min):
```

killua156: yes that makes sense.

isuruf:
@killua156 can you fix the tests?

conda_smithy/linter/messages/recipe.py
```diff
@@ -1037,6 +1037,28 @@ class DuplicateRecipes(LinterMessage, _AnyRecipeMessage):
...
+    """
+    conda-forge's global pinning already provides the `python_min` variable,
+    so recipes should not redefine it to the same value. Only override
+    `python_min` when the package needs a higher floor than the global
+    default.
+    """
+
+    kind = "hint"
+    identifier = "R-052"
+    added_in = "2026.6"
```
```suggestion
    added_in = "2026.7"
```

conda_smithy/linter/utils.py
```diff
@@ -217,31 +218,56 @@ def _lint_package_version(version: Optional[str]) -> Optional[str]:
...
-
-
-load_linter_toml_metdata = load_linter_toml_metadata  # BW Compat
+PINNING_FEEDSTOCK_RAW = (
+    "https://raw.githubusercontent.com/conda-forge/conda-forge-pinning-feedstock/main"
+)
 
 
-@lru_cache(maxsize=1)
-def load_linter_toml_metdata_internal(time_salt):
```
jaimergp, line 231: We can't remove this (even if marked as internal) because it was not marked as private with a leading `_`. This needs to be deprecated first. Check the `deprecations.py` module for helper decorators.

conda_smithy/linter/utils.py
```diff
@@ -217,31 +218,56 @@ def _lint_package_version(version: Optional[str]) -> Optional[str]:
...
+
+
+def load_linter_toml_metadata():
+    url = f"{PINNING_FEEDSTOCK_RAW}/recipe/linter_hints/hints.toml"
+    if (hints_toml_str := _try_fetch_url_content_cached(url)) is None:
+        return None
     return tomllib.loads(hints_toml_str)
 
 
+load_linter_toml_metdata = load_linter_toml_metadata  # BW Compat
```
jaimergp: Use `conda_smithy.deprecations.deprecated.constant()` for this.

conda_smithy/linter/messages/recipe.py
```diff
@@ -1037,6 +1037,28 @@ class DuplicateRecipes(LinterMessage, _AnyRecipeMessage):
...
+    message = (
+        "The recipe sets `python_min` to ${value}, which is identical to "
```
jaimergp: Your comparison above uses `<=`, but this conveys `==`. Shall it say:
```suggestion
        "The recipe sets `python_min` to ${value}, which is equal or lower than "
```

conda_smithy/linter/messages/recipe.py
```diff
@@ -1037,6 +1037,28 @@ class DuplicateRecipes(LinterMessage, _AnyRecipeMessage):
...
+        "the default provided by conda-forge's global pinning. Remove the "
+        "redefinition; it is redundant and unintentionally overrides "
+        "platforms where the global default differs. Only set `python_min` "
```
```suggestion
        "the default provided by conda-forge's global pinning. Please "
        "remove the redefinition. You only need to set `python_min` "
```

jaimergp: No need to be that detailed in the hint message, I'd say. This detail can be added in the docstring instead.

conda_smithy/linter/messages/recipe.py
```diff
@@ -1037,6 +1037,28 @@ class DuplicateRecipes(LinterMessage, _AnyRecipeMessage):
     )
 
 
+@dataclass(kw_only=True)
+class RedundantPythonMin(LinterMessage, _AnyRecipeMessage):
```
jaimergp, line 1041: Can you provide the `examples()` method too, please? We are still backfilling the existing messages, so there are not many precedents, but hopefully we can slowly get there.

killua156:
Fixed the tests and addressed the review:

- Tests were failing on a `NameError` (`Version` wasn't imported) — switched to conda's `VersionOrder`. The hint now reports the recipe's declared value instead of the global default.
- Kept the `<=` check so a lower-than-default floor also gets hinted, and added a test for it.
- `load_linter_toml_metdata_internal` and the misspelled `load_linter_toml_metdata` alias are now deprecated via `conda_smithy.deprecations` instead of removed; internal callers moved to `load_linter_toml_metadata`.
- Reworded the message to "equal or lower than", moved the details to the docstring, added `examples()`, and set `added_in = "2026.7"`.
- Rebased on main and regenerated `linter-messages.json`.

conda_smithy/linter/hints.py
killua156: Kept `<=` and added a test for the lower-floor case.

conda_smithy/linter/utils.py
killua156: Restored it as a deprecated shim via `conda_smithy.deprecations` instead of removing it.

conda_smithy/linter/utils.py
killua156: Done — using `deprecated.constant()` for the alias now.

conda_smithy/linter/messages/recipe.py
killua156: Reworded to "equal or lower than" and moved the rationale to the docstring.

conda_smithy/linter/messages/recipe.py
killua156: Added.

isuruf (approved):

conda_smithy/linter/utils.py
```diff
@@ -217,31 +219,72 @@ def _lint_package_version(version: Optional[str]) -> Optional[str]:
...
+def load_linter_toml_metadata():
+    url = f"{PINNING_FEEDSTOCK_RAW}/recipe/linter_hints/hints.toml"
+    if (hints_toml_str := _try_fetch_url_content_cached(url)) is None:
+        return None
     return tomllib.loads(hints_toml_str)
 
 
+@deprecated(
+    "2026.7",
+    "2027.1",
```
h-vetinari: Our new deprecation policy is 60 days, not 6 month. AFAICT this should be 2026.9, but CC @jaimergp @beckermr for confirmation

h-vetinari (changes requested):

conda_smithy/linter/utils.py
jaimergp: [Yes](https://github.com/conda-forge/conda-smithy#deprecation-policy). You can even [add a timedelta of 60d](https://github.com/conda-forge/conda-smithy/blob/325dc4cbad6ba3b7afba62e70e222fe64c247c56/conda_smithy/deprecations.py#L309-L310) instead of computing the string.

h-vetinari (approved):
Thanks!

PR #2583, opened 2026-06-12
lint against duplicate recipes

h-vetinari:
Follow-up to #2473, where we added a lint about duplicate configuration files (ef3e9123ce98902e4736b6acc54aa59839f7dbac). I was surprised to find that we don't have the same lint for duplicate recipe files.

As another follow-up from that PR, revert the `fail-on-error: false` for coveralls (ea1897ffe91d9f22e514fc64a01254a39d374132)


beckermr (changes requested):
Needs a news item.

h-vetinari:
Thanks for the quick review! 🙏

Issue #2582, opened 2026-06-12
Further improve `workflow_settings:`


h-vetinari:
In #2349 we came up with a new design for the various knobs that may be required for tuning feedstock infrastructure, which had previously been provider-specific. As of today, we are able to do things like
```yaml
workflow_settings:
  build_workspace_dir:
    - value: C:\bld
      os: win
  free_disk_space:
    - value: quick    # or skip or max
  pagefile_size:
    - value: 15
  store_build_artifacts:
    - value: true
  tools_install_dir:
    - value: C:\Miniforge
      os: win
    - value: ~/miniforge3
      os: unix
```

Each key under `workflow_settings:` is a list that can be filtered à la
```yaml
workflow_settings:
  foo:
    - value: ...
      os: [linux, osx, unix, win]             # or any subset
      provider: [azure, github_actions, ...]  # or any subset
      platform: [linux-64, etc. as usual]     # or any subset
    - value: ...  # other list item for other combination
```

This issue is for migrating remaining functionality to this new setup, ensuring things are as cross-platform and cross-provider as possible, and possibly extend this with other things deemed useful.

* [ ] `max_parallel`: max number of concurrent jobs; currently implemented separately on AZP & GHA
* [ ] `timeout_minutes`: max duration of jobs; currently implemented separately on AZP & GHA
* [ ] `triggers` (GHA-only?): may not be necessary
* [x] `resize_partitions` (win-only): #2584
* [ ] `install_atl` (win-only): current implementation is AZP-only
   https://github.com/conda-forge/conda-smithy/blob/e1482e4eb087d9b5601c12e44b9c516a44e082ac/conda_smithy/schema.py#L110-L113

Additional tasks:
* [ ] testing the implementations from [#2349](https://github.com/conda-forge/conda-smithy/issues/2349) on all providers
   * [ ] Ensure `free_disk_space` works:
     * [x] AZP
     * [x] GHA
     * [ ] namespace
     * [ ] blacksmith
     * [ ] depot
   * [ ] Ensure `pagefile_size` works:
     * [x] AZP
     * [x] GHA
     * [x] namespace (#2577)
     * [ ] blacksmith
     * [ ] depot
   * [ ] Ensure `store_build_artifacts` works:
     * [x] AZP
     * [x] GHA
     * [ ] namespace
     * [ ] blacksmith
     * [ ] depot
* [ ] #2586
[...]

PR #2584, opened 2026-06-12
Implement `workflow_settings.resize_partitions`

mgorny:
Checklist

Convert `github_actions.resize_win_partitions` into a more generic `workflow_settings.resize_partitions`. However, for now it's only doing anything on Windows GHA, and it's unclear to me if any other CI provider will ever feature reusable free space.

### DIFF
--- a/news/2584-resize-partitions.rst
+++ b/news/2584-resize-partitions.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* ``workflow_settings.resize_partitions`` to control resizing partitions across CI providers. (#2584)
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/2584-resize-partitions.rst
+++ b/news/2584-resize-partitions.rst
@@ -8,7 +8,7 @@
 
 **Deprecated:**
 
-* <news item>
+* ``github_actions.resize_win_partitions`` was deprecated in favor of ``workflow_settings.resize_partitions``. (#2584)
 
 **Removed:**
 
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -2631,6 +2631,23 @@ def _read_forge_config(forge_dir, forge_yml=None):
                 }
             )
 
+    if "resize_win_partitions" in file_config.get("github_actions", {}):
+        if "resize_partitions" in file_config.get("workflow_settings", {}):
+            logger.warning(
+                "`github_actions.resize_win_partitions` is ignored when "
+                "`workflow_settings.resize_partitions` is set",
+            )
+            del config["github_actions"]["resize_win_partitions"]
+        else:
+            # Convert old keys to new settings.
+            config["workflow_settings"]["resize_partitions"].append(
+                {
+                    "provider": "github_actions",
+                    "os": "win",
+                    "value": config["github_actions"]["resize_win_partitions"],
+                }
+            )
+
     # check for conda-smithy 2.x matrix which we can't auto-migrate
     # to conda_build_config
     if file_config.get("matrix") and not os.path.exists(
--- a/conda_smithy/data/conda-forge.json
+++ b/conda_smithy/data/conda-forge.json
@@ -1149,7 +1149,7 @@
             }
           ],
           "default": [],
-          "description": "Free up disk space before building.\nTakes one of the following values:\n\n- `no` not to clean anything (the default)\n- `quick` to clean a subset of components that should yield\n  a quick space gain\n- `max` for the most significant space gain\n\nOn all platforms, `quick` removes a number of unnecessary system\ntools. On Linux, it additionally uninstall a few large system\npackages.\n\nThe `max` option additionally prunes unnecessary Docker images\non Linux.\n\nThe extent of cleanup may change in the future. However, changes\nwill only propagate through rerendering.\n\nPlease note that on Windows only C:\\ drive is cleaned up, so\nusing cleanup is only beneficial if this drive is used for\n`tools_install_dir`, `build_workspace_dir` or the page file.",
+          "description": "Free up disk space before building.\nTakes one of the following values:\n\n- `skip`: does not do anything (the default)\n- `quick`: cleans a subset of components that balances space gained\n  against the time necessary to do so\n- `max`: takes longer, but achieves more significant space gain\n\nOn all platforms, `quick` removes a number of unnecessary system\ntools. On Linux, it additionally uninstall a few large system\npackages.\n\nThe `max` option additionally prunes unnecessary Docker images\non Linux.\n\nThe extent of cleanup may change in the future. However, changes\nwill only propagate through rerendering.\n\nPlease note that on Windows only `C:\\` drive is cleaned up, so\nusing cleanup is only beneficial if this drive is used for\n`tools_install_dir`, `build_workspace_dir` or the page file.",
           "title": "Free Disk Space"
         },
         "resize_partitions": {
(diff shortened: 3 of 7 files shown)


mgorny:
Open question: do we want to add tests for every `workflow_settings` addition, or are we fine with the tests we have so far and assume new additions are boilerplate enough?

h-vetinari:
> However, for now it's only doing anything on Windows GHA, and it's unclear to me if any other CI provider will ever feature reusable free space.

Can we raise an error if it's specified too broadly (i.e. anything more general than `os: win` AND `provider: github_actions`)? This kind of thing has been one of my concerns since the introduction of `workflow_settings:`; I don't want people to wrongly believe something is more general than it actually is. That's both a bad developer experience (to find out after bashing your head against the wall why things don't work as expected that the option doesn't actually do what you thought it would), but more importantly, it precludes us from ever expanding such options to new or existing providers in a way without having to potentially deal with the aftermath on affected feedstocks.

That's why I've listed something along those lines at the end of #2582.

> Open question: do we want to add tests for every `workflow_settings` addition, or are we fine with the tests we have so far and assume new additions are boilerplate enough?

The smithy tests here for these things are mostly a theoretical exercise; what counts is validating this against actual CI (again something I put in #2582). So for this PR, I'm fine to keep the testing simple w.r.t. checking expectations.

I'd still like to improve on the testing side of "this condition is exclusive to that combination" or "these two `- value:`s overlap/conflict".

h-vetinari (changes requested):
Didn't you forget to deprecate the old value?

mgorny:
> Can we raise an error if it's specified too broadly (i.e. anything more general than `os: win` AND `provider: github_actions`)? This kind of thing has been one of my concerns since the introduction of `workflow_settings:`; I don't want people to wrongly believe something is more general than it actually is. That's both a bad developer experience (to find out after bashing your head against the wall why things don't work as expected that the option doesn't actually do what you thought it would), but more importantly, it precludes us from ever expanding such options to new or existing providers in a way without having to potentially deal with the aftermath on affected feedstocks.

Mixed feelings about this. I don't disagree with you, but it kinda feels annoying to be forcing users to replace a short:

```yaml
github_actions:
  resize_win_partitions: true
```

with complex:

```yaml
workflow_settings:
  resize_partitions:
    - provider: github_actions
      os: win
      value: true 
```

especially that for most of Windows providers, it still won't do anything (just waste a lot of time trying to resize already-maxxed out partitions).

Maybe we should shelve it for now, to avoid needless churn.

Or maybe it just makes sense to do it as you say, it's rare enough.

h-vetinari:
> replace a short: [foo] with complex: [bar]

To me that's par for the course with a more general approach, and not at all a problem in my view. Rather, it becomes immediately clear where this setting applies. And trying to expand it beyond what's possible (or implemented) should IMO warn/error.

> especially that for most of Windows providers, it still won't do anything

That's kinda my point. If it's not implemented (or only useable on GHA), we shouldn't lead people to believe that it's applicable on other windows providers.

> Maybe we should shelve it for now, to avoid needless churn.

I mean, the migration from whatever hodgepodge we have now towards `workflow_settings:` strikes me as pretty good timing (only have to touch it once), BUT, I'll happily concede that it's not the most high impact / high prio thing.

> Or maybe it just makes sense to do it as you say, it's rare enough.

For `resize_partitions`, definitely. The principle isn't rare though IMO; we'll have more and more keys that need to reflect some restrictions on their applicable domain.

h-vetinari (approved):
LGTM, but it's quite late for me, so I'll leave others to check this out, or have another look when I'm fresh later. 🙃

mgorny:
> > especially that for most of Windows providers, it still won't do anything
> 
> That's kinda my point. If it's not implemented (or only useable on GHA), we shouldn't lead people to believe that it's applicable on other windows providers.

Sorry, what I meant is that even most GHA providers won't do a thing here. IIRC it only does anything with our CIrun boxes.

news/2584-resize-partitions.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* ``workflow_settings.resize_partitions`` to control resizing partitions across CI providers. (#2584)
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
```
jaimergp: Let's add the deprecated value here.

jaimergp:
I agree with @h-vetinari; we could have a single top-level key per setting·OS·CI combo like `resize_partitions_win_gha` and `resize_partitions_osx_azp` and that would be "very simple", but we would soon end up with a soup of keys. If we had to think about whether this is worth it or not, I could join the debate, but most of the code is already written so let's just polish the UX. Is it very tricky to catch the non-Win, non-GHA cases for the warnings/hints? If very tricky, current state is sufficient.

jaimergp (approved):

h-vetinari:
> Sorry, what I meant is that even most GHA providers won't do a thing here

This actually highlights a bigger problem. Currently we have various different CI providers that are all lumped together as
```yaml
provider:
  <target>: github_actions  # is it MSFT, namespace, blacksmith, depot, cirun?!
```
so we cannot reflect this (IMO necessary) distinction in our `workflow_settings` yet. Do we want to add a (ignore bikeshed colour) `subprovider:`? Or should we break up `provider: github_actions` into smaller pieces?

> most of the code is already written so let's just polish the UX. Is it very tricky to catch the non-Win, non-GHA cases for the warnings/hints? If very tricky, current state is sufficient.

IMO this doesn't have to happen in this PR. But I put this point in #2582, because IMO what's necessary is a setup that allows restricting certain keys under `workflow_settings:` to a specific domain (e.g. only applicable on windows, or only applicable on a specific provider), and then a framework that applies this structurally (i.e. not through case-by-case logic).

mgorny:
> Or should we break up provider: github_actions into smaller pieces?

That's an interesting idea. If blacksmith/namespace didn't have a bunch of different profiles, it would be a nice way of avoiding manually specifying all the labels.

mgorny:
I'll add the platform check later today.

jaimergp:
> Or should we break up provider: github_actions into smaller pieces?

That's going to create some churn. What if we introduce a syntax to glob/match github_actions_labels? This way we don't have to keep definitions so up-to-date. Like:

```
provider:
  <target>: github_actions@/glob-or-regex-goes-here/
```

mgorny:
I'm sorry but I don't really see what that'd accomplish.

h-vetinari:
I like the `github_actions@<actual_provider>` idea. It's nicely concise and self-explanatory. I don't think we'd even need to glob or regex necessarily, just bare values `github_actions@namespace` would be enough for the vast majority of cases I could imagine that need the provider-distinction (not least because you can always add another `workflow_settings: foo: ...` list entry if there's more than one provider in play).

Naturally, `github_actions` without the `@` would mean all possible providers, and we could hint (rather than lint) if certain requested features are not available on the provider that's requested through `github_action_labels`.

h-vetinari:
> I'm sorry but I don't really see what that'd accomplish.

As a concrete example for `resize_partitions`,  which only really works on `cirun` (AFAIU)
```yaml
workflow_settings:
  resize_partitions:
    - value: true
      provider: github_actions
      # --> lint, not restricted to win

workflow_settings:
  resize_partitions:
    - value: true
      provider: github_actions
      os: win
      # --> hint if github_action_labels specifies a non-cirun provider for windows

workflow_settings:
  resize_partitions:
    - value: true
      provider: github_actions@cirun
      os: win
      # --> no hint or lint, because the config is restricted to a combination that's known to work
```

mgorny:
I'm sorry for being unclear. I was referring specifically to Jaime's top-level `provider:` example.

That said, such level of granularity doesn't seem helpful. In the end, I don't expect people to be using two different GHA Windows providers simultaneously. What we've done with PyTorch feedstock was not only rare, but awfully hard to set up in the first place.

h-vetinari (approved):
This PR LGTM; we can continue the discussion about how to deal with providers in a separate issue

PR #2590, opened 2026-06-14
stop depending on implicit mutation of lints/hints

h-vetinari:
To me, code like

https://github.com/conda-forge/conda-smithy/blob/c771d9b52815e29bf5f52bebf502c085b3d62147/conda_smithy/lint_recipe.py#L251-L264

is a major code smell, because there's no indication (besides the meta-context of "we're collecting lints") that `lints` is being mutated from call to call. Same for `hints`. 

In this PR, I'm suggesting to switch away from implicit mutation, towards explicit passing of these quantities. This makes the code much easier to reason about (IMO).

### DIFF
--- a/conda_smithy/lint_recipe.py
+++ b/conda_smithy/lint_recipe.py
@@ -178,19 +178,19 @@ def lintify_meta_yaml(
             lints.append(f"Unsupported recipe.yaml schema version {schema_version}")
             return lints, hints
 
-    sources_section = get_section(meta, "source", lints, recipe_version)
-    build_section = get_section(meta, "build", lints, recipe_version)
-    requirements_section = get_section(meta, "requirements", lints, recipe_version)
+    sources_section, lints = get_section(meta, "source", lints, recipe_version)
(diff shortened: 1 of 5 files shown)


h-vetinari:
pre-commit.ci autofix

jaimergp:
Sadly this changes the return type of public symbols, which is covered by the [deprecation policy](https://github.com/conda-forge/conda-smithy#deprecation-policy) and needs a 60 day period. Not sure if it's all worth the trouble of adding the shims for bwcompat.

h-vetinari:
We haven't been very consistent about adding underscores to private functions. We could add underscores first (keeping the old names as direct copies), and once the 60 days for that are over, we remove the old names and then can change our freshly re-privatized symbols

jaimergp:
I like that @h-vetinari. I'd give that a try and/or even more radically we could even think of a global private move except for whatever we want to leave public. Thoughts @beckermr?

beckermr:
I do not care in the least as long as we respect the deprecation periods.

PR #2519, opened 2026-04-13
[PyCon DE Sprints] README: Update development setup instructions, general cleanup

NMertsch:
This PR updates the README file:

- Update development setup instructions
- Add `pytest` and `pre-commit` instructions
- Migrate from rst-style headers to md-style headers- 
- Minor grammar and style fixes
- Update token setup instructions and contact information

Checklist
* [x] Added a ``news`` entry
* [x] Regenerated schema JSON if schema altered (`python -m conda_smithy.schema`)

PR #2593, opened 2026-06-14
fix some pyrefly errors (part 1)

pavelzw:
Checklist

https://github.com/conda-forge/conda-smithy/pull/2591#discussion_r3409369860

this brings down the pyrefly errors from 223 to 204

my agent said the following

```
• Implemented the first “obvious fixes” batch.

  Changes made:

  - Removed dead Python 2 urllib fallback in bootstrap-obvious-ci-and-miniconda.py.
  - Fixed small uninitialized/local narrowing issues in token rotation and linter noarch checks.
  - Tightened simple signatures in ci_register, plugin, github, and validate_schema.
  - Made feedstock_io.get_repo() handle missing pygit2 and missing repos explicitly.
  - Corrected the datetime(...) construction in deprecation date arithmetic.
  - Fixed test helper and env-var preservation typing.

  Verification:

  - pixi run pyrefly ... now reports 203 errors, down from 222.
  - Focused pytest passed: 20 passed, 1 xfailed.
  - pixi run ruff check ... passed on touched files.
  - git diff --check passed.
```

done with codex but i looked over all fixes and made sure they make sense

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,34 @@
+# Developing conda-smithy
+
+## Set up Development Environment
+
+To install conda-smithy from source:
+
+* Install `conda`
+* Fork and clone this repository: `git clone https://github.com/YOUR-USERNAME/conda-smithy.git`. Change to it: `cd conda-smithy`.
+* Create a new conda environment with all requirements based on [environment.yml](environment.yml): `conda env create`.
+* Activate the environment: `conda activate conda-smithy`.
+* Install conda-smithy: `pip install --no-deps --editable .`
+
+To run all tests:
+
+```sh
+$ pytest
+```
+
+To run all code checks:
+
+```sh
+# staged changes
+$ pre-commit run
+# all files
+$ pre-commit run --all-files
+```
+
+To run pyrefly code checks:
+
+```sh
+$ pyrefly check --python-interpreter-path $(which python) --output-format min-text --count-errors=1 --search-path .
+```
+
+If you encounter pyrefly issues that you don't agree with, feel free to add a `# pyrefly: ignore[<error-type>]` comment to that line.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,3 +32,4 @@ $ pyrefly check --python-interpreter-path $(which python) --output-format min-te
 ```
 
 If you encounter pyrefly issues that you don't agree with, feel free to add a `# pyrefly: ignore[<error-type>]` comment to that line.
+An ignored `pyrefly` error is insufficient reason to block the merge of a pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ To install conda-smithy from source:
 
 * Install `conda`
 * Fork and clone this repository: `git clone https://github.com/YOUR-USERNAME/conda-smithy.git`. Change to it: `cd conda-smithy`.
-* Create a new conda environment with all requirements based on [environment.yml](environment.yml): `conda env create`.
+* Create a new conda environment with all requirements based on [environment.yml](environment.yml): `conda env create -f environment.yml`.
 * Activate the environment: `conda activate conda-smithy`.
 * Install conda-smithy: `pip install --no-deps --editable .`
 
--- a/bootstrap-obvious-ci-and-miniconda.py
+++ b/bootstrap-obvious-ci-and-miniconda.py
@@ -11,11 +11,7 @@ import argparse
 import os
 import platform
 import subprocess
-
-try:
-    from urllib.request import urlretrieve
-except ImportError:
-    from urllib import urlretrieve
+from urllib.request import urlretrieve
 
 MINICONDA_URL_TEMPLATE = (
     "https://repo.continuum.io/miniconda/Miniconda{major_py_version}-"
--- a/conda_smithy/anaconda_token_rotation.py
+++ b/conda_smithy/anaconda_token_rotation.py
@@ -62,6 +62,7 @@ def rotate_anaconda_token(
 
     anaconda_token = _get_anaconda_token()
 
+    gh = None
     if github_actions:
         gh = Github(gh_token())
 
@@ -167,6 +168,7 @@ def rotate_anaconda_token(
                             raise RuntimeError(err_msg)
 
                 if github_actions:
+                    assert gh is not None
                     try:
                         rotate_token_in_github_actions(
                             user, project, anaconda_token, token_name, gh
--- a/conda_smithy/anaconda_token_rotation.py
+++ b/conda_smithy/anaconda_token_rotation.py
@@ -62,10 +62,6 @@ def rotate_anaconda_token(
 
     anaconda_token = _get_anaconda_token()
 
-    gh = None
-    if github_actions:
-        gh = Github(gh_token())
-
     # capture stdout, stderr and suppress all exceptions so we don't
     # spill tokens
     failed = False
@@ -145,8 +141,7 @@ def rotate_anaconda_token(
                             raise e
                         else:
                             err_msg = (
-                                f"Failed to rotate token for {user}/{project}"
-                                " on azure!"
+                                f"Failed to rotate token for {user}/{project} on azure!"
                             )
                             failed = True
                             raise RuntimeError(err_msg)
@@ -168,7 +163,7 @@ def rotate_anaconda_token(
                             raise RuntimeError(err_msg)
 
                 if github_actions:
-                    assert gh is not None
+                    gh = Github(gh_token())
                     try:
                         rotate_token_in_github_actions(
                             user, project, anaconda_token, token_name, gh
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -29,8 +29,7 @@ def get_repo(path, search_parent_directories=True):
 
 
 def get_repo_root(path):
-    repo = get_repo(path)
-    if repo is None:
+    if (repo := get_repo(path)) is None:
         return None
     return repo.workdir.rstrip(os.path.sep)
 
--- a/conda_smithy/feedstock_io.py
+++ b/conda_smithy/feedstock_io.py
@@ -9,7 +9,10 @@ def get_repo(path, search_parent_directories=True):
     repo = None
     try:
         import pygit2
+    except ImportError:
+        return None
 
+    try:
         if search_parent_directories:
             path = pygit2.discover_repository(path)
         if path is not None:
@@ -19,8 +22,6 @@ def get_repo(path, search_parent_directories=True):
                 no_search = pygit2.GIT_REPOSITORY_OPEN_NO_SEARCH
 
             repo = pygit2.Repository(path, no_search)
-    except ImportError:
-        pass
     except pygit2.GitError:
         pass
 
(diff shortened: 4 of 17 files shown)


pyproject.toml
```diff
@@ -31,6 +31,10 @@ include-package-data = true
 write_to = "conda_smithy/_version.py"
 write_to_template = "__version__ = '{version}'"
 
+[tool.pyrefly]
+project-includes = ["**/*.py"]
+search-path = [".", "tests"]
+
```
pavelzw, line 37: this is needed, otherwise `pixi run pyrefly check --python-interpreter-path $(pixi run which python)` doesn't show all errors

conda_smithy/feedstock_io.py
```diff
@@ -19,19 +22,17 @@ def get_repo(path, search_parent_directories=True):
...
 
     return repo
 
 
 def get_repo_root(path):
-    try:
-        return get_repo(path).workdir.rstrip(os.path.sep)
-    except AttributeError:
+    repo = get_repo(path)
+    if repo is None:
```
h-vetinari: What do we think about walruses? (I find it nice to reduce the repetition of `foo = ...; if foo ...`)
```suggestion
    if (repo := get_repo(path)) is None:
```

h-vetinari (approved):
I'll admit I don't know pyrefly well, but I'm broadly in favour of using it (and potentially other type checkers) in CI; some of these fixes look a bit doubtful in terms of benefit, but nothing I'm actually opposed to. Thanks for the PR!

h-vetinari:
I'll leave this open for a bit so other can chime in if desired

h-vetinari:
@conda-forge/core, any thoughts towards adding more static checks here? If we go down this road, this PR will be one of many until we fix the >200 or so issues currently flagged by pyrefly (see #2591)

If there are no further comments, I'm planning to merge this in ~48h or so.

isuruf:
> @conda-forge/core, any thoughts towards adding more static checks here? 

Unfortunately the first three files in the diff I looked at are false positives. Does adding more checks justify having to workaround false positives?

h-vetinari:
I know what you mean; I noted above

> some of these fixes look a bit doubtful in terms of benefit, but nothing I'm actually opposed to.

However, this is IMO a bit like formatting questions with black, i.e. not really worth arguing about, if someone thought about a structural approach and considers it worth linting.

pyrefly should help with topics like #2589, so that'd be an upside for which I could live with the opinionated-but-unnecessary changes. On the other hand, we could perhaps do this differently (e.g. mypy; tough I don't claim to know the differences between that and pyrefly).

There's also a possibility to create an exceptions file (see #2591, which has "passing" pyrefly checks despite not fixing anything), though this would currently be >2600 lines long. Since the those exceptions are very granular - e.g.
```json
    {
      "line": 172,
      "column": 72,
      "stop_line": 172,
      "stop_column": 74,
      "path": "conda_smithy/anaconda_token_rotation.py",
      "code": -2,
      "name": "unbound-name",
      "description": "`gh` may be uninitialized",
      "concise_description": "`gh` may be uninitialized",
      "severity": "error"
    },
```
this sounds like something we'd have to keep up-to-date somewhat laboriously (though perhaps there's automation for that), so I'd rather just get rid of these errors; assuming we want to adopt pyrefly.

Perhaps you could open an issue about adding pyrefly (and/or about more type-checking in general) @pavelzw? It seems we might need more discussion about this that isn't really suited for a random PR.

pavelzw:
I would feel more comfortable adding an autoupdater for py-rattler pinnings if we had type checking as mentioned in https://github.com/conda-forge/conda-smithy/issues/2589#issuecomment-4701337024

In the current state of this repo, probably all occurrences of py-rattler are covered by unit tests(?) so we would catch issues in the PR anyway but if py-rattler usage grows, there might be some hidden branches that are not covered by the testing suite and latest then I think type checking is needed. 

(I'm in general a fan of being able to see the types of something when working within my IDE, if this is properly supported in the project through type checkers, I'm all for it)

pavelzw:
#2602

pavelzw:
there has been no activity on the issue and py-rattler not being up to date is becoming an issue again (https://github.com/conda-forge/conda-smithy/issues/2608). how should we proceed here?

as mentioned in https://github.com/conda-forge/conda-smithy/issues/2589#issuecomment-4700991904, this would help a lot with automatic updates for typed projects like py-rattler

isuruf:
I'm against using (for lack of a better word) stupid tools. We end up battling the tool instead of writing code when we introduce these tools.

baszalmstra:
> I'm against using (for lack of a better word) stupid tools. We end up battling the tool instead of writing code when we introduce these tools.

I'm sorry, but this comment is not constructive at all. Can you clarify what you mean by "stupid tools"? Does pyrefly fall in that category? Are you against adding a type checker to this repository? Or this one in particular?

The reason this was proposed is that we want to pin versions of dependencies relatively strictly. But that adds the problem that the versions of these dependencies become stale quickly. The proposal was to make it easier to _automatically_ bump versions of dependencies so that part of solving this issue can be automated.

Furthermore, we are having a hard time getting people to contribute to this repository. A type checker would help catch bugs earlier and faster and make the API of the code easier to navigate. This makes it easier for people to start contributing.

isuruf:
> Can you clarify what you mean by "stupid tools"?

Tools that reject certain coding styles just because the author of the tool does not like that coding style.

> Does pyrefly fall in that category?

Yes, it does when I look at the diff of this PR.

> Are you against adding a type checker to this repository?

Not at all. A type checker that doesn't limit legitimate code would be great to have.

baszalmstra:
> A type checker that doesn't limit legitimate code would be great to have. 

Do you have one in mind? Or at least one that is more favorable? I'm just curious if we can make your criticism more actionable.

isuruf:
> Do you have one in mind? Or at least one that is more favorable?

No, I don't. I suggest you try a few type checkers and see which one doesn't reject legitimate code.

baszalmstra:
Can you point out the parts that are legitimate and which the type checker should not reject? That would help!

isuruf:
As I said above, the first three files in the diff are bad changes

pavelzw:
<details>
<summary>outdated</summary>

i still do think that the `runreqs_spacing` change makes sense.

image

if this `if` block is not hit, you'll end up walking into a undefined variable that produces weird behavior in line 86.

other type checkers such as ty or mypy also flagged this when enabling `--error possibly-unresolved-reference` (or `--check-untyped-defs` in mypy's case) but i imagine they have similar odd behavior around other places tbh. they also all produce a similar amount of errors on this codebase so i don't really see a benefit of switching to a different type checker apart from giving me some more work to do

</details>

see [`d027420` (this PR)](https://github.com/conda-forge/conda-smithy/pull/2593/commits/d0274205d3c4a204df05ad45a30c706676cf3475)

baszalmstra:
That means we’re kind of at a stalemate here. @isuruf, do you feel strongly enough about this to block the PR?

lucascolley:
how about this for the third file?

```diff
diff --git a/conda_smithy/anaconda_token_rotation.py b/conda_smithy/anaconda_token_rotation.py
index bcd6091d..b1d0e035 100644
--- a/conda_smithy/anaconda_token_rotation.py
+++ b/conda_smithy/anaconda_token_rotation.py
@@ -62,10 +62,6 @@ def rotate_anaconda_token(

     anaconda_token = _get_anaconda_token()

-    gh = None
-    if github_actions:
-        gh = Github(gh_token())
-
     # capture stdout, stderr and suppress all exceptions so we don't
     # spill tokens
     failed = False
@@ -168,7 +164,7 @@ def rotate_anaconda_token(
                             raise RuntimeError(err_msg)

                 if github_actions:
-                    assert gh is not None
+                    gh = Github(gh_token())
                     try:
                         rotate_token_in_github_actions(
                             user, project, anaconda_token, token_name, gh
```

pavelzw:
applied @lucascolley's suggestion

lucascolley (approved):

CONTRIBUTING.md
pavelzw, line 1: this is a minimal contributing.md partially stolen from https://github.com/conda-forge/conda-smithy/pull/2519. eventually, i'd like to introduce a more reproducible pixi workflow to this repo, then this might also change.

CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+$ pre-commit run --all-files
+```
+
+To run pyrefly code checks:
+
+```sh
+$ pyrefly check --python-interpreter-path $(which python) --output-format min-text --count-errors=1 --search-path .
+```
+
+If you encounter pyrefly issues that you don't agree with, feel free to add a `# pyrefly: ignore[<error-type>]` comment to that line.
```
pavelzw, line 34: as discussed in today's conda/conda-forge meeting with @isuruf.

will add this also to CI once all issues are fixed

conda_smithy/linter/hints.py
pavelzw, line 1: i updated this file that you had issues with @isuruf, ptal again

conda_smithy/anaconda_token_rotation.py
pavelzw, line 1: i updated this change that you had issues with, ptal again @isuruf

CONTRIBUTING.md
lucascolley, line 34:
```suggestion
If you encounter pyrefly issues that you don't agree with, feel free to add a `# pyrefly: ignore[<error-type>]` comment to that line.
An ignored `pyrefly` error is insufficient reason to block the merge of a pull request.
```

pavelzw (commented):
@isuruf ptal again

CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
+# Developing conda-smithy
+
+## Set up Development Environment
+
+To install conda-smithy from source:
+
+* Install `conda`
+* Fork and clone this repository: `git clone https://github.com/YOUR-USERNAME/conda-smithy.git`. Change to it: `cd conda-smithy`.
+* Create a new conda environment with all requirements based on [environment.yml](environment.yml): `conda env create`.
```
isuruf:
```suggestion
* Create a new conda environment with all requirements based on [environment.yml](environment.yml): `conda env create -f environment.yml`.
```

wolfv (approved):
Thank you for working on this!

bootstrap-obvious-ci-and-miniconda.py
jaimergp, line 1: I wouldn't spend too much time on the scripts placed outside `conda_smithy` and `tests`. I don't think we using or testing them, which means we are not maintaining them either.

pyproject.toml
```diff
@@ -31,6 +31,10 @@ include-package-data = true
 write_to = "conda_smithy/_version.py"
 write_to_template = "__version__ = '{version}'"
 
+[tool.pyrefly]
+project-includes = ["**/*.py"]
```
jaimergp:
```suggestion
project-includes = ["conda_smithy/**/*.py", "tests/**/*.py"]
```

As per https://github.com/conda-forge/conda-smithy/pull/2593/files#r3588715593 (I hope that's the syntax)

bootstrap-obvious-ci-and-miniconda.py
pavelzw: this was all codex, only some tokens were used

lucascolley: > I don't think we using or testing them, which means we are not maintaining them either.

can we just delete them, then?

pyproject.toml
pavelzw: i personally would rather just ensure we don't have any random .py files to begin with but no super strong opinion 🤷🏻

bootstrap-obvious-ci-and-miniconda.py
jaimergp: Probably, but not in this PR.

lucascolley: gh-2629

jaimergp: See https://github.com/conda-forge/conda-smithy/pull/2628

isuruf (approved):

PR #2588, opened 2026-06-14
fix: update deps, emulate feedstock CI env, fix testing paths

h-vetinari:
Towards finding out what's going on with: #2587

This matches the current [state](https://github.com/conda-forge/conda-smithy-feedstock/blob/cfed8476a22f44e86d5c5806e297458a97303b51/recipe/meta.yaml#L35-L36) of conda-smithy-feedstock, where we're seeing failures (likely due to some environmental differences): https://github.com/conda-forge/conda-smithy-feedstock/pull/391

closes #2587

PR #2594, opened 2026-06-14
don't pin rattler-build

h-vetinari:
Follow-up to #2588 
> I don't think we need the rattler-build pin? We're only interacting with that through `py-rattler`? At least, as long as py-rattler controls its constraints correctly, I don't think we need this.
> 
> I'd prefer not to have this constraint, because it didn't break anything yet (AFAICT), and it'd be one more fast-moving thing to keep in sync

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -39,7 +39,6 @@ dependencies:
   - pixi>=0.59.0
   - jsonschema
   - exceptiongroup
-  # rattler's APIs are subject to change, pin to minor versions
+  # py-rattler's APIs are subject to change, pin to minor versions
   - py-rattler>=0.22,<0.23
   - rattler-build-conda-compat>=1.4.15,<2.0.0a0
-  - rattler-build>=0.66.0,<0.67.0a0


beckermr (changes requested):
A change in rattler-build literally just broke things, so we should pin.

h-vetinari:
> A change in rattler-build literally just broke things, so we should pin.

It didn't? The failure was across a whole range of r-b versions, which would have continued to work fine, had it not been for a change in smithy (setting an env var) that uncovered a latent bug - but nothing that justifies pinning such a fast-moving library

beckermr:
Nope. `rattler-build` produced invalid JSON output due to how a new feature (the github integration) was implemented.

beckermr:
An easy way to see where the bug originates is to ask how it might have been prevented in the development of the new feature that caused the bug. The answer is that `rattler-build` needs to add a test upstream that ensures that when the CLI is called with `--render-only`, it outputs valid JSON under all conditions.

beckermr:
Had that test been added during development, it would have caught the error long before we had to go find it.

h-vetinari:
I disagree. It was triggered by the env var. There's no other reason to expect from many months of experience that r-b would regularly break smithy.

In terms of the order of events here, I didn't notice on a phone (while on a call) that you had put this in my PR, so you can't hold your "change request" over this PR. If you wanna get technical, we can either revert the whole of #2588 and then reapply the changes without this pin (and you can then propose this separately), or we merge this PR.

beckermr:
Feel free to revert whatever or merge whatever you want. 

I do not want to be tracking down subtle bugs due to rattler-build, and so want it pinned. If you'd like to take on that work, then I am happy to have it not pinned.

h-vetinari:
The context is that we're already far behind the (much slower-moving) `py-rattler` pin, with no-one to pick up the slack of keeping this up-to-date (I'm asking/pushing for this in #2589). 

If we add a pin on r-b, that'll just mean we will be frozen out of any improvements for long stretches at a time. I'd much prefer the (so far) very manageable risk that rattler vs. py-rattler messes up its internal API contract.

beckermr:
Indeed. If you want that pin updated, then you can feel free to update it. 

I am happy to wait until rattler-build is stable enough for that pin to be removed before I put time into rolling it out or supporting it further.

Issue #2574, opened 2026-06-08
`pixi run rerender` task is broken, "conda-smithy version is out-of-date"


lucascolley:
```console
❯ pixi run rerender
✨ Pixi task (rerender in smithy): conda-smithy rerender: (Rerender the feedstock.)
Traceback (most recent call last):
  File "/Users/lucascolley/ghq/github.com/conda-forge/array-api-compat-feedstock/.pixi/envs/smithy/bin/conda-smithy", line 10, in <module>
    sys.exit(main())
  File "/Users/lucascolley/ghq/github.com/conda-forge/array-api-compat-feedstock/.pixi/envs/smithy/lib/python3.10/site-packages/conda_smithy/cli.py", line 748, in main
    args.subcommand_func(args)
  File "/Users/lucascolley/ghq/github.com/conda-forge/array-api-compat-feedstock/.pixi/envs/smithy/lib/python3.10/site-packages/conda_smithy/cli.py", line 596, in __call__
    self._call(args, tmpdir)
  File "/Users/lucascolley/ghq/github.com/conda-forge/array-api-compat-feedstock/.pixi/envs/smithy/lib/python3.10/site-packages/conda_smithy/cli.py", line 601, in _call
    configure_feedstock.main(
  File "/Users/lucascolley/ghq/github.com/conda-forge/array-api-compat-feedstock/.pixi/envs/smithy/lib/python3.10/site-packages/conda_smithy/configure_feedstock.py", line 2937, in main
    check_version_uptodate("conda-smithy", __version__, True)
  File "/Users/lucascolley/ghq/github.com/conda-forge/array-api-compat-feedstock/.pixi/envs/smithy/lib/python3.10/site-packages/conda_smithy/configure_feedstock.py", line 2647, in check_version_uptodate
    raise RuntimeError(f"{msg} Exiting.")
RuntimeError: conda-smithy version (3.56.3) is out-of-date (2026.5.29) in conda-smithy's environment. Exiting.
```

lucascolley:
`pixi update conda-smithy` fixes this

h-vetinari:
This is expected...?

lucascolley:
If expected, it would be nice to provide a more obviously actionable error message.

[...]

PR #2598, opened 2026-06-15
Use `pixi exec` for non-build tasks

jaimergp:
Checklist

Idea suggested by @chrisburr in [#general > how is conda-smithy kept up-to-date in Pixi lock file?](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/how.20is.20conda-smithy.20kept.20up-to-date.20in.20Pixi.20lock.20file.3F/with/603281721)

### DIFF
--- a/conda_smithy/templates/pixi.toml.tmpl
+++ b/conda_smithy/templates/pixi.toml.tmpl
@@ -53,12 +53,12 @@ cmd = "inspect_artifacts --recipe-dir {{ recipe_dir }} -m .ci_support/{{ variant
 description = "List contents of {{ feedstock_name }} packages built for variant {{ variant }}"
 {%- endfor %}
 
-{%- set _feature_platforms = platforms | select("in", shellcheck_platforms) -%}
-{%- if (shellcheck is defined) and shellcheck.enabled and _feature_platforms -%}
-{%- set with_shellcheck = " --with shellcheck" -%}
-{%- else -%}
-{%- set with_shellcheck = "" -%}
-{%- endif -%}
+{%- set _feature_platforms = platforms | select("in", shellcheck_platforms) %}
+{%- if (shellcheck is defined) and shellcheck.enabled and _feature_platforms %}
+{%- set with_shellcheck = " --with shellcheck" %}
+{%- else %}
+{%- set with_shellcheck = "" %}
+{%- endif %}
 
 [feature.smithy]
 [feature.smithy.tasks.build-locally]
--- a/conda_smithy/templates/pixi.toml.tmpl
+++ b/conda_smithy/templates/pixi.toml.tmpl
@@ -12,6 +12,27 @@ channels = ["conda-forge"]
 platforms = {{ platforms | tojson }}
 requires-pixi = ">={{ pixi_version }}"
 
+{%- set _feature_platforms = platforms | select("in", shellcheck_platforms) %}
+{%- if (shellcheck is defined) and shellcheck.enabled and _feature_platforms %}
+{%- set with_shellcheck = " --with shellcheck" %}
+{%- else %}
+{%- set with_shellcheck = "" %}
+{%- endif %}
+
+[tasks.build-locally]
+cmd = "pixi exec --force-reinstall --list='^python$' python ./build-locally.py"
+description = "Build packages locally using the same setup scripts used in conda-forge's CI"
+[tasks.smithy]
+cmd = "pixi exec --force-reinstall --list='^conda-smithy$' conda-smithy"
+description = "Run conda-smithy. Pass necessary arguments."
+[tasks.rerender]
+cmd = "pixi exec --force-reinstall --list='^conda-smithy$' conda-smithy rerender"
+description = "Rerender the feedstock."
+
+[tasks.lint]
+cmd = "pixi exec{{ with_shellcheck }} --list='^conda-smithy$' conda-smithy lint --conda-forge {{ recipe_dir }}"
+description = "Lint the feedstock recipe"
+
 [feature.build.dependencies]
 {%- for spec_name, spec_constraints in build_tool_deps_dict | dictsort %}
 {{ spec_name }} = "{{ spec_constraints }}"
@@ -53,32 +74,7 @@ cmd = "inspect_artifacts --recipe-dir {{ recipe_dir }} -m .ci_support/{{ variant
 description = "List contents of {{ feedstock_name }} packages built for variant {{ variant }}"
 {%- endfor %}
 
-{%- set _feature_platforms = platforms | select("in", shellcheck_platforms) %}
-{%- if (shellcheck is defined) and shellcheck.enabled and _feature_platforms %}
-{%- set with_shellcheck = " --with shellcheck" %}
-{%- else %}
-{%- set with_shellcheck = "" %}
-{%- endif %}
-
-[feature.smithy]
-{#- --force-reinstall tries to update everything but only relinks the changed packages #}
-[feature.smithy.tasks.build-locally]
-cmd = "pixi exec --force-reinstall --list=python python ./build-locally.py"
-description = "Build packages locally using the same setup scripts used in conda-forge's CI"
-[feature.smithy.tasks.smithy]
-cmd = "pixi exec --force-reinstall --list=conda-smithy conda-smithy"
-description = "Run conda-smithy. Pass necessary arguments."
-[feature.smithy.tasks.rerender]
-cmd = "pixi exec --force-reinstall --list=conda-smithy conda-smithy rerender"
-description = "Rerender the feedstock."
-
-[feature.smithy.tasks.lint]
-cmd = "pixi exec{{ with_shellcheck }} conda-smithy lint --conda-forge {{ recipe_dir }}"
-description = "Lint the feedstock recipe"
-
 [environments]
-build = {features = ["build"], no-default-feature = true}
-smithy = {features = ["smithy"], no-default-feature = true}
-
-# This is a copy of default, to be enabled by build_steps.sh during Docker builds
-# __PLATFORM_SPECIFIC_ENV__ = {features = ["build"], no-default-feature = true}
+build = ["build"]
+# This is a copy of 'build', to be enabled by build_steps.sh during Docker builds
+# __PLATFORM_SPECIFIC_ENV__ = ["build"]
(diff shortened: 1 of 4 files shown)


chrisburr:
The more I've thought about this the more I think it's the right way to go.

conda_smithy/templates/pixi.toml.tmpl
```diff
@@ -12,71 +12,69 @@ channels = ["conda-forge"]
...
-[tasks."debug-{{ variant }}"]
+[feature.build.tasks."debug-{{ variant }}"]
 cmd = "{{ conda_build_tool }} debug {{ recipe_dir }} -m .ci_support/{{ variant }}.yaml"
 description = "Debug {{ feedstock_name }} with variant {{ variant }} directly (without setup scripts)"
 {%- else %}
-[tasks."build-{{ variant }}"]
+[feature.build.tasks."build-{{ variant }}"]
 cmd = "{{ conda_build_tool }} build --recipe {{ recipe_dir }} -m .ci_support/{{ variant }}.yaml"
 description = "Build {{ feedstock_name }} with variant {{ variant }} directly (without setup scripts)"
 {%- endif %}
```
h-vetinari, line 71: As of #2566, we now have debug also for rattler-build.

lucascolley (commented):
This is an improvement so I think we should merge it!

I think it can still fail when there are cache hits right around the time that new conda-smithy releases hit conda-forge. My preferred solution would be to place an exact pin on `conda-smithy` in the manifest, and have `rerender` update that pin during the error path at https://github.com/conda-forge/conda-smithy/blob/3e8e7681a212052a0bad58c8a8302ab19cf8024a/conda_smithy/configure_feedstock.py#L2866-L2871 If I'm not wrong, the user would then just need to be told to `pixi run rerender` one more time.

But that idea can stay in an gh-2574 until a need arises, this is probably good enough :)

jaimergp:
> I think it can still fail when there are cache hits right around the time that new conda-smithy releases hit conda-forge. My preferred solution would be to place an exact pin on conda-smithy in the manifest [...]

Then the pin is tied to the `conda-smithy` source, so every Pixi update would need a smithy release 😬 It feels to me that getting the latest version of `conda-smithy` is for `pixi exec` to verify. Isn't that how it works?

lucascolley:
> Then the pin is tied to the `conda-smithy` source

No, I'm not suggesting hardcoding the version. The code I pointed to is already grabbing the version dynamically with a `rattler` query: https://github.com/conda-forge/conda-smithy/blob/3e8e7681a212052a0bad58c8a8302ab19cf8024a/conda_smithy/configure_feedstock.py#L2846-L2855

conda_smithy/templates/pixi.toml.tmpl
jaimergp: Added in [3b546a8](https://github.com/conda-forge/conda-smithy/pull/2598/commits/3b546a89538f7740ff6c99ed6dbefc7232044f7c)

jaimergp:
Ah, ok, but still, if it's out of date, we need to rerender with an outdated version, so we can't get out of that catch 22, can we? Users are also free to pass `--no-check-uptodate` though.

lucascolley:
> Ah, ok, but still, if it's out of date, we need to rerender with an outdated version, so we can't get out of that catch 22, can we? Users are also free to pass `--no-check-uptodate` though.

The first `pixi run rerender` would not rerender with the outdated version — instead, it would (just) update the `conda-smithy` pin and instruct the user to run the task again. The second `pixi run rerender` would then automatically update the lock file to satisfy the new pin, and the rerender would occur with the up-to-date smithy version.

We could even customise the message printed by the first invocation, e.g.

```
INFO: conda-smithy version was out of date!
INFO: {insert advice for Pixi users to `pixi run rerender`}
INFO: {insert advice for non-Pixi users}
INFO:conda_smithy.configure_feedstock:You can commit the changes with:

    git commit -m "MNT: updated conda-smithy workspace pin to 2026.6.14, with conda-smithy 2026.6.13

Other tools:
- conda-build 26.5.0
- rattler-build 0.66.1
- rattler-build-conda-compat 1.4.15"
```

h-vetinari:
> Users are also free to pass `--no-check-uptodate` though.

I'm not in favour of this... The entire point of that check is that feedstocks get rerendered with a current version (which has been important given the rate of changes recently).

Perhaps I don't know pixi tasks well enough, but I'd naïvely imagine that this could be structured as a dag of tasks where the rerender depends on an update-smithy task?

lucascolley:
> Perhaps I don't know pixi tasks well enough, but I'd naïvely imagine that this could be structured as a dag of tasks where the rerender depends on an update-smithy task?

I don't have much experience with Pixi tasks that mutate their own workspace, but in principle that could work, yes, just calling `pixi update conda-smithy`. Maybe we would need to make sure that the environment is re-activated before the rerender?

jaimergp:
I think we would need to:

- First, verify whether the `exec` cache really gets stale.
- If that's the case, we can add a custom message for `pixi`-enabled feedstocks telling them to run `pixi clean cache --exec` or something like that.

And if that becomes sufficiently frequent and annoying, we can then investigate whether we look into this pinning strategy. It'd be more interesting to invest that effort into making `pixi exec` auto-update when needed, though (if that's even the problem).

Edit: a quick look looks like the cache is more static than I thought. `--force-reinstall` could help but it's also wasteful because it reinstalls even if there are no new versions available, right? Would an `--update` flag be sufficient?

jaimergp:
Hm, I'm more confused now by this issue: https://github.com/prefix-dev/pixi/issues/2914. Is it auto-updating always or not?

jaimergp:
As per a conversation in Prefix's Discord, turns out that `--force-reinstall` does not relink EVERYTHING. It does perform a fresh solve and then relinks whatever changed (including transitive). IMO that's good enough for this use case.

> Perhaps I don't know pixi tasks well enough, but I'd naïvely imagine that this could be structured as a dag of tasks where the rerender depends on an update-smithy task?

I think that could work, but we would still have a smithy installation per cloned feedstock. Not sure if we want to keep that.

h-vetinari:
> I think that could work, but we would still have a smithy installation per cloned feedstock. Not sure if we want to keep that.

Isn't that the status quo with pixi currently?

jaimergp:
Not sure I'm following, but `pixi exec` writes temporary envs in a global cache, not per workspace. Like a disposable (cached) `pixi global install XXX` + running `XXX`.

conda_smithy/templates/pixi.toml.tmpl
```diff
@@ -12,71 +12,73 @@ channels = ["conda-forge"]
...
-{%- if (shellcheck is defined) and shellcheck.enabled and _feature_platforms -%}
-{%- set _smithy_features = _smithy_features + ["shellcheck"] -%}
-[feature.shellcheck]
-dependencies.shellcheck = "*"
-platforms = {{ _feature_platforms | sort | tojson }}
-
-{% endif -%}
 [environments]
-smithy = {{ _smithy_features | sort | tojson }}
+build = {features = ["build"], no-default-feature = true}
```
lucascolley: Is it true that the default feature is empty now?

conda_smithy/templates/pixi.toml.tmpl
```diff
@@ -12,71 +12,73 @@ channels = ["conda-forge"]
...
+{%- set with_shellcheck = " --with shellcheck" %}
+{%- else %}
+{%- set with_shellcheck = "" %}
+{%- endif %}
+
+[feature.smithy]
+{#- --force-reinstall tries to update everything but only relinks the changed packages #}
 [feature.smithy.tasks.build-locally]
-cmd = "python ./build-locally.py"
+cmd = "pixi exec --force-reinstall --list=python python ./build-locally.py"
```
lucascolley: awesome, `pixi exec --list` was my first contribution to Pixi :)

lucascolley (approved):

conda_smithy/templates/pixi.toml.tmpl
jaimergp: Yes, I noticed that `pixi` rendered the `default` env unconditionally

lucascolley: WDYT?

```diff
diff --git a/conda_smithy/templates/pixi.toml.tmpl b/conda_smithy/templates/pixi.toml.tmpl
index d6f69331..b00a8e4e 100644
--- a/conda_smithy/templates/pixi.toml.tmpl
+++ b/conda_smithy/templates/pixi.toml.tmpl
@@ -60,25 +60,23 @@ description = "List contents of {{ feedstock_name }} packages built for variant
 {%- set with_shellcheck = "" %}
 {%- endif %}

-[feature.smithy]
 {#- --force-reinstall tries to update everything but only relinks the changed packages #}
-[feature.smithy.tasks.build-locally]
+[tasks.build-locally]
 cmd = "pixi exec --force-reinstall --list=python python ./build-locally.py"
 description = "Build packages locally using the same setup scripts used in conda-forge's CI"
-[feature.smithy.tasks.smithy]
+[tasks.smithy]
 cmd = "pixi exec --force-reinstall --list=conda-smithy conda-smithy"
 description = "Run conda-smithy. Pass necessary arguments."
-[feature.smithy.tasks.rerender]
+[tasks.rerender]
 cmd = "pixi exec --force-reinstall --list=conda-smithy conda-smithy rerender"
 description = "Rerender the feedstock."

-[feature.smithy.tasks.lint]
+[tasks.lint]
 cmd = "pixi exec{{ with_shellcheck }} conda-smithy lint --conda-forge {{ recipe_dir }}"
 description = "Lint the feedstock recipe"

 [environments]
-build = {features = ["build"], no-default-feature = true}
-smithy = {features = ["smithy"], no-default-feature = true}
+build = ["build"]

-# This is a copy of default, to be enabled by build_steps.sh during Docker builds
-# __PLATFORM_SPECIFIC_ENV__ = {features = ["build"], no-default-feature = true}
+# This is a copy of `build`, to be enabled by build_steps.sh during Docker builds
+# __PLATFORM_SPECIFIC_ENV__ = ["build"]
```

EDIT: hmm maybe would still want a separate env to avoid having to set a `default-environment` for the smithy tasks

jaimergp: Let me check

lucascolley: to isolate the smithy tasks to one env I suppose

```toml
build = { features = ["build"], no-default-feature = true }
```

is still the easiest way

jaimergp: Yea, that works too, thanks!

baszalmstra (approved):

baszalmstra:
Anything that needs to happen before we can merge this?

lucascolley:
thanks all!

PR #2596, opened 2026-06-15
copy all files as non-executable by default

h-vetinari:
As it turns out, there are circumstances where conda-smithy will wrongly populate a rerendered recipe with executable bits also for non-executable files, e.g.
```
[main 444ef60] MNT: Re-rendered with conda-smithy 2026.6.14 and conda-forge-pinning 2026.06.14.21.57.4
 12 files changed, 210 insertions(+), 108 deletions(-)
 delete mode 100755 .azure-pipelines/azure-pipelines-linux.yml
 mode change 100644 => 100755 .ci_support/README
 mode change 100644 => 100755 .gitattributes
 create mode 100644 .github/workflows/conda-build.yml
 mode change 100644 => 100755 .gitignore
 mode change 100644 => 100755 .scripts/logging_utils.sh
 mode change 100644 => 100755 LICENSE.txt
```
which is what we ran into with https://github.com/conda-forge/conda-forge-bot/pull/6237.

In fact, we're definitely generating executable bits on `.azure-pipelines/*.yamls`, e.g. on the current main branch of arrow-cpp-feedstock (note the 755 at the beginning):
```
>git ls-files --stage .azure-pipelines
100755 9d232948952fed848626c3534ae48192e88540b8 0       .azure-pipelines/azure-pipelines-osx.yml
100755 2244f1131dc4a489b8366f51c42ce69188b0e478 0       .azure-pipelines/azure-pipelines-win.yml
```
due to the fact that those files are treated as templates
https://github.com/conda-forge/conda-smithy/blob/3e8e7681a212052a0bad58c8a8302ab19cf8024a/conda_smithy/configure_feedstock.py#L1990-L2004
which is borked, because
https://github.com/conda-forge/conda-smithy/blob/3e8e7681a212052a0bad58c8a8302ab19cf8024a/conda_smithy/configure_feedstock.py#L1713
just sets the executable bit unconditionally for everything.

So to clean this up, copy everything as non-executable by default (https://github.com/conda-forge/conda-smithy/commit/9457bb2bdc3b89c056e60c000251912259f36486), then distinguish executable files from templates, and add the exec bit only for the former. Add a test that validates that all executable files have the exec bit, and that all other files _don't_.

PR #2599, opened 2026-06-15
fix: do not set exec perms on all azure files

beckermr:
Checklist

This PR fixes a bug in the azure CI scripts where all files were set as executable when in fact only the non-yml files need executable permissions.


tests/test_rerender_permissions.py
```diff
@@ -0,0 +1,84 @@
...
+]
+
+
+ALL_EXECUTABLE_FILES = EXECUTABLE_TEMPLATES + [
+    ".circleci/checkout_merge_commit.sh",
+    ".circleci/fast_finish_ci_pr_build.sh",
+    "build-locally.py",
+    ".azure-pipelines/azure-pipelines-linux.yml",
+    ".azure-pipelines/azure-pipelines-osx.yml",
+    ".azure-pipelines/azure-pipelines-win.yml",
```
beckermr:
```suggestion
```

beckermr:
@pre-commit.ci autofix

tests/test_perms.py
```diff
@@ -0,0 +1,85 @@
...
+]
+
+
+ALL_EXECUTABLE_FILES = EXECUTABLE_TEMPLATES + [
+    ".circleci/checkout_merge_commit.sh",
+    ".circleci/fast_finish_ci_pr_build.sh",
+    "build-locally.py",
+    ".azure-pipelines/azure-pipelines-linux.yml",
+    ".azure-pipelines/azure-pipelines-osx.yml",
+    ".azure-pipelines/azure-pipelines-win.yml",
```
beckermr:
```suggestion
```

conda_smithy/configure_feedstock.py
```diff
@@ -1742,8 +1744,16 @@ def _render_template_exe_files(forge_config, jinja_env, template_files, forge_di
...
+
+        if add_exe:
+            # Fix permission of template shell files
+            set_exe_file(target_fname, True)
+
+
+def _render_template_exe_files(forge_config, jinja_env, template_files, forge_dir):
+    _render_template_files(
+        forge_config, jinja_env, template_files, forge_dir, add_exe=True
+    )
```
h-vetinari, line 1756: I'm not a big fan of this. I find that breaking up `_render_template_exe_files` into two separate function calls is much cleaner, because otherwise it's way too easy to forget again that all the other CI providers are currently calling `_render_template_exe_files` without making a conscious, documented distinction between templates and executable files.

Also,
https://github.com/conda-forge/conda-smithy/blob/ef2133572859fa570e78eaf2029c486416f730c8/conda_smithy/configure_feedstock.py#L3006-L3018
is simply wrong now; not only because it doesn't contain all relevant scripts, but because it's actually checking
https://github.com/conda-forge/conda-smithy/blob/ef2133572859fa570e78eaf2029c486416f730c8/conda_smithy/configure_feedstock.py#L1721-L1726
if templates are rendered differently, which should apply to all of them.

tests/test_rerender_permissions.py
h-vetinari, line 1: That test was a few hours work and needs to come with my commit authorship (e.g. https://github.com/conda-forge/conda-smithy/commit/58aeb07c452fa34ae83f311d4fdddd7c2f312454). If you want to break it out into a separate test (or otherwise modify it slightly), feel free to add a `Co-Authored-By:` tag.

beckermr: Yep absolutely. This was for me to debug.

conda_smithy/configure_feedstock.py
beckermr: Yeah this pr doesn't need to be merged and neither does the other one. The real issue is the release was not uploaded correctly. Hopefully you can shed light on that.

h-vetinari: > this pr doesn't need to be merged and neither does the other one.

We still have to fix _at least_:
* the exec bits on azure yaml files
* the incomplete list in `get_common_scripts`
* add a test for exec bits
* and we should split `_render_template_exe_files` just because it's all confused about itself

> The real issue is the release was not uploaded correctly. Hopefully you can shed light on that.

See [here](https://github.com/conda-forge/conda-smithy/pull/2596#issuecomment-4712383005). In addition to the above, I want us to assume a more defensive posture on executable bits, because that's a trivial attack vector, and shutting that off more tightly is good practice anyway.

PR #2600, opened 2026-06-15
fix: Use the feedstock name instead of meta.name to register feedstocks where the package name is dependent on a variant

xhochy:
Feedstock creation for `go-compiler` currently fails with a conda-smithy issue as the name contains a variant variable. 

```
Making feedstock for go-compiler
  File "/home/runner/Miniforge3/bin/conda-smithy", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/runner/Miniforge3/lib/python3.13/site-packages/conda_smithy/cli.py", line 774, in main
    args.subcommand_func(args)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/runner/Miniforge3/lib/python3.13/site-packages/conda_smithy/cli.py", line 149, in __call__
    package=argparse.Namespace(name=meta.name())
                                    ~~~~~~~~~^^
  File "/home/runner/Miniforge3/lib/python3.13/site-packages/rattler_build_conda_compat/render.py", line 139, in name
    check_bad_chrs(name, "package/name")
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/Miniforge3/lib/python3.13/site-packages/conda_build/metadata.py", line 834, in check_bad_chrs
    raise CondaBuildUserError(
        f"Bad character(s) ({''.join(sorted(invalid))}) in {field}: {value}."
    )
conda_build.exceptions.CondaBuildUserError: Bad character(s) ( $) in package/name: go-${{ go_variant_str }}-compiler.
```

The recipe contains a `feedstock-name` which should be used instead.

See also: https://github.com/conda-forge/admin-requests/actions/runs/27467405766/job/81192070603

AI-Disclaimer: The fix is manually written, but DeepSeek-V4 helped with writing a unit test.

### DIFF
--- a/news/feedstock-name-init.rst
+++ b/news/feedstock-name-init.rst
@@ -1,4 +1,3 @@
 **Fixed:**
 
 * Use the feedstock name instead of ``meta.name`` to register feedstocks where the package name is dependent on a variant.
-
--- a/news/feedstock-name-init.rst
+++ b/news/feedstock-name-init.rst
@@ -1,3 +1,3 @@
 **Fixed:**
 
-* Use the feedstock name instead of ``meta.name`` to register feedstocks where the package name is dependent on a variant.
+* Use the feedstock name instead of ``meta.name`` to register feedstocks where the package name is dependent on a variant. (#2600)
--- a/news/feedstock-name-init.rst
+++ b/news/feedstock-name-init.rst
@@ -1,0 +1,4 @@
+**Fixed:**
+
+* Use the feedstock name instead of ``meta.name`` to register feedstocks where the package name is dependent on a variant.
+
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -351,8 +351,6 @@ def test_init_with_feedstock_name(go_compiler_recipe):
     Regression test: When the package name depends on a variant but the recipe
     has a feedstock-name in extra, the Init command should use the feedstock
     name (not meta.name()) to create the feedstock directory.
-
-    See https://github.com/conda-forge/staged-recipes/recipes/go-compiler/recipe.yaml
     """
     parser = argparse.ArgumentParser()
     subparser = parser.add_subparsers()
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -357,9 +357,7 @@ def test_init_with_feedstock_name(go_compiler_recipe):
     init_obj = cli.Init(subparser)
 
     recipe = go_compiler_recipe.recipe
-    feedstock_dir_template = os.path.join(
-        recipe, "{package.name}-feedstock"
-    )
+    feedstock_dir_template = os.path.join(recipe, "{package.name}-feedstock")
     args = InitArgs(
         recipe_directory=os.path.join(recipe, "recipe"),
         feedstock_directory=feedstock_dir_template,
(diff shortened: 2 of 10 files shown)


conda_smithy/utils.py
```diff
@@ -61,10 +61,11 @@ def get_feedstock_name_from_meta(
...
     """Get the feedstock name from a parsed meta.yaml or recipe.yaml."""
-    if "feedstock-name" in meta.meta["extra"]:
-        return meta.meta["extra"]["feedstock-name"]
-    elif "parent_recipe" in meta.meta["extra"]:
-        return meta.meta["extra"]["parent_recipe"]["name"]
+    extra = meta.meta.get("extra", {})
+    if "feedstock-name" in extra:
+        return extra["feedstock-name"]
+    elif "parent_recipe" in extra:
+        return extra["parent_recipe"]["name"]
```
h-vetinari: Review note: this change is essentially a no-op; it only assigns a variable to `meta.meta["extra"]`.

tests/test_cli.py
```diff
@@ -346,6 +346,31 @@ def test_render_variant_mismatches(testing_workdir):
         assert data["a"] == data["b"]
 
 
+def test_init_with_feedstock_name(go_compiler_recipe):
+    """
+    Regression test: When the package name depends on a variant but the recipe
+    has a feedstock-name in extra, the Init command should use the feedstock
+    name (not meta.name()) to create the feedstock directory.
+    """
```
h-vetinari, line 354: IMO this should not be a CLI test, but go under `test_configure_feedstock.py`; as a lot of tests in that module show, we can call the rerender directly, without going through the CLI

xhochy: This is not a rerender, but a bug in the acutal CLI code. `test_configure_feedstock.py` tests functionality in the `conda_smithy.configure_feedstock` module. This file here tests `conda_smithy.cli`.

conda_smithy/utils.py
jaimergp: FWIW I didn't know we supported this mapping:

```yaml
extra:
  parent_recipe:
    name: XXX
```

??? Or is that a rendered `recipe.yaml` artifact?

news/feedstock-name-init.rst
```diff
@@ -0,0 +1,3 @@
+**Fixed:**
+
+* Use the feedstock name instead of ``meta.name`` to register feedstocks where the package name is dependent on a variant.
```
jaimergp:
```suggestion
* Use the feedstock name instead of ``meta.name`` to register feedstocks where the package name is dependent on a variant. (#2600)
```

conda_smithy/utils.py
xhochy: I will revert this no-op change, but first: The `parent_recipe` is coming from a rendered recipe. This is not happening in the path this PR is about, but in other code paths in conda-smithy.

xhochy: This was actually necessary, but the better fix is to adjust the test recipe instead of doing the change here. In practice, we should never encounter a _valid_ recipe without an `extras` section. Better fail in that scenario.

xhochy:
@jaimergp @h-vetinari Can you give this another look, please?

jaimergp (approved):
LGTM. I'll note that I'm not sure how this interacts with the clobbering checks in the staged-recipes linter ("does the resulting feedstock exist already?"), but that should not block here.

xhochy:
This change should not have any effect on the linter as it only fixes the repository creation step (coming after the linter)

PR #2601, opened 2026-06-15
Move Windows to GHA by default (and other small platform fixes)

jaimergp:
Checklist

With https://github.com/conda-forge/conda-smithy/issues/2349 closed, we have feature parity between Azure and GHA, so we can switch to GHA on Windows too (including `arm64`!).

### DIFF
--- a/conda_smithy/schema.py
+++ b/conda_smithy/schema.py
@@ -734,6 +734,7 @@ class ConfigModel(BaseModel):
             osx_64: osx_64
             osx_arm64: osx_arm64
             win_64: win_64
+            win_arm64: win_arm64
         ```
         """),
     )
@@ -854,7 +855,9 @@ class ConfigModel(BaseModel):
 
         * `linux_64`
         * `osx_64`
+        * `osx_arm64`
         * `win_64`
+        * `win_arm64`
         * `linux_aarch64`
         * `linux_ppc64le`
         * `linux_s390x`
@@ -863,6 +866,7 @@ class ConfigModel(BaseModel):
         The following CI services are available:
 
         * `azure`
+        * `github_actions`
         * `circle`
         * `travis`
         * `appveyor`
--- a/conda_smithy/schema.py
+++ b/conda_smithy/schema.py
@@ -10,7 +10,7 @@ from inspect import cleandoc
 from typing import Annotated, Any, Literal, Optional, Union
 
 import yaml
-from conda.base.constants import KNOWN_SUBDIRS
+from conda.base.constants import PLATFORMS
 from pydantic import BaseModel, ConfigDict, Field, WithJsonSchema, create_model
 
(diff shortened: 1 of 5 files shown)


jaimergp:
Tested at https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/1796. Note that `win-arm64` doesn't seem to be part of the `python` zipped keys; had to use `ca-certificates` to let the WOA runners play.

h-vetinari:
> Note that `win-arm64` doesn't seem to be part of the `python` zipped keys

It's [explicitly handled](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/1e6807e13462ff3e723744f6801c7db6613adffb/recipe/conda_build_config.yaml#L945), but there's only 3.14+ builds.

> With #1971 closed

Did you mean #2349? :)

conda_smithy/schema.py
```diff
@@ -863,6 +866,7 @@ class ConfigModel(BaseModel):
         The following CI services are available:
 
         * `azure`
+        * `github_actions`
         * `circle`
         * `travis`
         * `appveyor`
```
h-vetinari, line 872: Unrelated to this PR: we really should start cleaning these out of the codebase; there's even more ancient providers not mentioned here still hanging around (woodpecker, drone, ...), and we're carrying all that code for nothing IMO.

news/2601-gha-windows.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Github Actions is now the default platform for Windows builds. (#2601)
+* Github Actions is now the default platform for native Linux ARM builds. (#2601)
```
h-vetinari, line 8: I think the more interesting question is if we want to add `linux-aarch64` by default, even if it's not added explicitly under `providers:`. Sounds like the right time for it?

(For osx-arm64, we need a bigger pool on osx: https://github.com/conda-forge/conda-forge.github.io/issues/2829)

h-vetinari (approved):
Basically LGTM, though I think enabling linux-aarch64 by default would be worth a thought

isuruf:
> I think enabling linux-aarch64 by default would be worth a thought

-1 on this. Instead, we should send the migrator for linux-aarch64 to enable it for the rest of the nodes and when we close that migrator (after X% has been issued), we enable it by default.

h-vetinari:
> Instead, we should send the migrator for linux-aarch64 to enable it for the rest of the nodes and when we close that migrator (after X% has been issued), we enable it by default.

Fine by me. To me, that migrator was the arch migrator itself; switching from `build_platform: linux_aarch: linux_64` to `provider: linux_aarch: default` has been essentially completely hassle-free for me so far. 🤷

jaimergp:
Any other thoughts about this @conda-forge/core?

Issue #2586, opened 2026-06-13
Enforce config accuracy for `workflow_settings` that don't apply to all combinations


h-vetinari:
Part of #2582

There are two main axes here:
* linting/raising is config under `workflow_setting` requests something that we know the underlying platform/provider cannot actually achieve
* distinguishing the various _actual_ providers that are currently hiding behind `provider: github_actions`.

Some relevant quotes from #2584, where this was recently discussed (this may be my bias, feel free to edit if you have the rights, or comment what's missing).

---
@h-vetinari
> @mgorny: However, for now it's only doing anything on Windows GHA, and it's unclear to me if any other CI provider will ever feature reusable free space.

Can we raise an error if it's specified too broadly (i.e. anything more general than `os: win` AND `provider: github_actions`)? This kind of thing has been one of my concerns since the introduction of `workflow_settings:`; I don't want people to wrongly believe something is more general than it actually is. That's both a bad developer experience (to find out after bashing your head against the wall why things don't work as expected that the option doesn't actually do what you thought it would), but more importantly, it precludes us from ever expanding such options to new or existing providers in a way without having to potentially deal with the aftermath on affected feedstocks.

---

@h-vetinari 
> @mgorny Sorry, what I meant is that even most GHA providers won't do a thing here

This actually highlights a bigger problem. Currently we have various different CI providers that are all lumped together as
```yaml
provider:
  <target>: github_actions  # is it MSFT, namespace, blacksmith, depot, cirun?!
```
so we cannot reflect this (IMO necessary) distinction in our `workflow_settings` yet. Do we want to add a (ignore bikeshed colour) `subprovider:`? Or should we break up `provider: github_actions` into smaller pieces?

---

@jaimergp 
> @h-vetinari: Or should we break up provider: github_actions into smaller pieces?

That's going to create some churn. What if we introduce a syntax to glob/match github_actions_labels? This way we don't have to keep definitions so up-to-date. Like:

```yaml
provider:
  <target>: github_actions@/glob-or-regex-goes-here/
```

---

@h-vetinari 
I like the `github_actions@<actual_provider>` idea. It's nicely concise and self-explanatory. I don't think we'd even need to glob or regex necessarily, just bare values `github_actions@namespace` would be enough for the vast majority of cases I could imagine that need the provider-distinction (not least because you can always add another `workflow_settings: foo: ...` list entry if there's more than one provider in play).

[...]

PR #2603, opened 2026-06-24
Replace `ValueError`s for `workflow_settings` with lints

mgorny:
Checklist

Work-in-progress, part of #2586:

- [x] path variables not restricted to unix/win
- [x] `resize_partitions` on non-GHA/Win
- [x] duplicate workflow setting values
- [x] conflicting `os` / `platform` restrictions?

One significant difference is that we're linting if any path is not os-restricted, irrespective of whether the platform is actually supported by the recipe or not. I don't think there's an easy way to determine whether a platform is used without adding significant performance impact (like rerendering), and either way it's more future-proof to always specify `os` there.

I've also skipped the part where we actually try to guess whether it's valid Unix/Windows path. The `ValueError` was done this way since it was non-trivial to figure out whether variable was actually os-restricted or not. Since we're checking that explicitly now, I guess we can assume the user knows what they're doing.


mgorny:
CC @h-vetinari for early feedback.

mgorny:
I've added a rough equivalent of the check we have for overlapping values. However, as it stands now it doesn't cover cases like:

```yaml
- provider: github_actions
  value: ...
- provider: github_actions
  os: linux
  value: ...
```

i.e. when one condition is a subset of the other one. The logic's already complex and I'm at my wits end. The whole approach there probably needs to be rethought.

mgorny:
Okay, I have an idea how to solve this. The code's going to be ugly, but we can look at optimizing it from a working prototype.

mgorny:
That said, after implementing all that neck-breaking logic to detect potential overlaps, I'm starting to think that perhaps we should utilize the opposite approach: check that all entries use the same conditions. So rather than stating that:

```yaml
workflow_settings:
  ...:
  - provider: github_actions
    value: ...
  - provider: github_actions
    os: linux
    value: ...
```

may potentially overlap, we state that "all entries must have the same granularity". Then the overlap check would be limited to checking for exact condition matches. The end result would be roughly the same, but _I think_ both the logic and the messages would be easier to comprehend.

One corner case to consider is whether we want to allow combining disjoint `os:` and `platform:` conditions, i.e. something like:

```yaml
workflow_settings:
  ...:
  - os: linux
    value: ...
  - platform: win_64
    value: ...
```

or whether we require that both specify either/both `os` and `platform`.

conda_smithy/data/linter-messages.json
```diff
@@ -135,6 +135,54 @@
...
+      "kind": "lint",
+      "added_in": "TODO",
+      "deprecated_in": "",
+      "documentation": "Lint when a path variable in `workflow_settings` is not correctly platform-specific.",
+      "message": "`workflow_settings.${setting}[${index}]` specifies path `${value}` without restricting it to Unix / Windows via the `os` or `platform` keys (applies to ${os}).\n",
+      "examples": [],
+      "path": ""
+    },
+    {
+      "name": "WorkflowSettingsNonSpecific",
```
h-vetinari: I'm wondering if we can find a better name for this, possibly something like
```suggestion
      "name": "WorkflowSettingsSpecificEntryTooLoose",
```

conda_smithy/data/linter-messages.json
```diff
@@ -135,6 +135,54 @@
...
+      "kind": "lint",
+      "added_in": "TODO",
+      "deprecated_in": "",
+      "documentation": "Lint when there are overlapping entries in `workflow_settings`.",
+      "message": "`workflow_settings.${setting} has potentially overlapping entries:\n${entries}.\n",
+      "examples": [],
+      "path": ""
+    },
+    {
+      "name": "WorkflowSettingsNonPlatformSpecificPath",
```
h-vetinari, line 163: love these!

tests/test_lint_recipe.py
```diff
@@ -5009,5 +5009,106 @@ def test_run_conda_forge_specific_routes_missing_to_lints(monkeypatch):
...
+    recipe_dir = tmp_path / "recipe"
+    recipe_dir.mkdir()
+    (recipe_dir / "meta.yaml").write_text(textwrap.dedent("""
+        package:
+          name: foo
+        """))
+
+    cfyml.write_text(textwrap.dedent(r"""
+        workflow_settings:
+          # bad: a single valie
```
h-vetinari:
```suggestion
          # bad: a single value
```

h-vetinari (commented):
Wow, I loooove this! 🤩 

Thank you @mgorny!

Just some very minor nits

conda_smithy/data/linter-messages.json
mgorny: "Specific entry is not specific enough" :-).

mgorny:
By the way, any clue what do we put in the version field now? Given that CalVer made the versions completely unpredictable.

h-vetinari:
> By the way, any clue what do we put in the version field now? Given that CalVer made the versions completely unpredictable.

We [discussed](https://github.com/conda-forge/conda-smithy/pull/2575#discussion_r3378598494) this in a recent PR; please use `2026.7` here

mgorny:
Okay, done.

h-vetinari (approved):
Thank you!

PR #2604, opened 2026-06-25
Update PyPI URL hints to ref files.pythonhosted.org

mgorny:
Checklist

xref https://github.com/conda-forge/conda-forge.github.io/issues/2878

Update the hints to suggest `files.pythonhosted.org` rather than `pypi.org`, as the former is the canonical URL per upstream documentation:
https://docs.pypi.org/api/#predictable-urls

It saves a single unnecessary redirect and reduces load off pypi.org, though it seems that currently both hostnames are served off the same server. It also follows suit with what other distributions are doing, e.g. Arch Linux and Gentoo:
https://wiki.archlinux.org/title/Python_package_guidelines#Source https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/pypi.eclass#n173


mgorny:
Open questions:
- <del>Should I rename the hint class and function? It's oddly specific.</del> Done that.
- Should I extend the hint to existing `pypi.org` uses, or leave it for now?

conda_smithy/linter/messages/recipe.py
```diff
@@ -1006,18 +1006,18 @@ class UsePip(LinterMessage, _AnyRecipeMessage):
 
 
 @dataclass(kw_only=True)
-class UsePyPIOrg(LinterMessage, _AnyRecipeMessage):
+class LegacyPyPIURL(LinterMessage, _AnyRecipeMessage):
```
jaimergp, line 1011: I'm afraid we can't just rename this. Needs to go through a deprecation cycle first. Simply introduce the new one and deprecate this one.

jaimergp: This can be done easily with one of the helpers in `deprecations.py`. See examples of `deprecated.constant()` in https://github.com/search?q=repo%3Aconda%2Fconda+deprecated.constant&ref=opensearch&type=code .

conda_smithy/linter/hints.py
```diff
@@ -31,20 +31,18 @@ def hint_pip_usage(build_section, hints):
                 hints.append(msg.r.UsePip().as_string())
 
 
-def hint_sources_should_not_mention_pypi_io_but_pypi_org(
-    sources_section: list[dict[str, Any]], hints: list[str]
-):
+def hint_legacy_pypi_url(sources_section: list[dict[str, Any]], hints: list[str]):
```
jaimergp, line 39: Same here, we can't just rename public symbols. Use the `deprecated` decorator.

conda_smithy/linter/messages/recipe.py
mgorny: Hmm, how about keeping the old name then? I don't think it's really critical how it's named.

mgorny:
Added deprecated aliases, and rebased. I've followed the conda deprecation policy, i.e. aimed for 2027.3.

conda_smithy/linter/hints.py
```diff
@@ -32,20 +34,25 @@ def hint_pip_usage(build_section, hints):
...
     """
     for source_section in sources_section:
         source = source_section.get("url", "") or ""
         sources = [source] if isinstance(source, str) else source
         if any(s.startswith("https://pypi.io/") for s in sources):
-            hints.append(msg.r.UsePyPIOrg().as_string())
+            hints.append(msg.r.LegacyPyPIURL().as_string())
+
+
+@deprecated("2026.7", "2027.3", addendum="Use hint_legacy_pypi_url() instead")
```
jaimergp:
```suggestion
@deprecated("2026.7", "2026.10", addendum="Use hint_legacy_pypi_url() instead")
```

It's 60 days in conda-smithy. Used October for good measure.

conda_smithy/linter/messages/recipe.py
```diff
@@ -1006,22 +1008,31 @@ class UsePip(LinterMessage, _AnyRecipeMessage):
...
     message = (
-        "PyPI default URL is now pypi.org, and not pypi.io."
+        "PyPI default URL is now files.pythonhosted.org, and not pypi.io."
         " You may want to update the default source url."
     )
 
 
+deprecated.constant(
+    "2026.7",
+    "2027.3",
```
```suggestion
    "2026.10",
```

Same here

jaimergp (approved):

PR #2607, opened 2026-06-30
Add a lint for mismatched feedstock name

### LINKED ISSUE #2605: Lint against mismatched feedstock names
In https://github.com/conda-forge/admin-requests/pull/2176 we have found that smithy will infer the wrong feedstock name when the main output name doesn't match the repository name. These cases need `extra.feedstock-name` forcefully, otherwise API calls that configure the repo fail. This is easy to miss so we should have a lint rule to prevent this drift.

mgorny:
Checklist

Fixes #2605

Add a check that compares the feedstock name (set in `extra.feedstock-name` or inferred from the recipe name) against the git remote, and reports a lint if they do not match.


conda_smithy/linter/messages/conda_forge.py
```diff
@@ -191,3 +191,17 @@ def examples(cls) -> list[Self]:
...
+
+@dataclass(kw_only=True)
+class MismatchedFeedstockName(LinterMessage):
+    """
+    Lint when the feedstock name is mismatched to recipe name.
+    """
+
+    kind = "lint"
+    identifier = "CF-009"
+    added_in = "TODO"
```
jaimergp:
```suggestion
    added_in = "26.7"
```

conda_smithy/linter/lints.py
```diff
@@ -1018,3 +1020,55 @@ def process_recursively(key: str, value: Any) -> None:
...
+) -> None:
+    """Lint that feedstock-name is specified when it doesn't match the recipe name"""
+
+    # v1 recipes use "recipe" or "package", v0 just "package"
+    recipe_name = (
+        meta["recipe"]
+        if recipe_version == 1 and "recipe" in meta
+        else meta.get("package", {}).get("name")
+    )
+    feedstock_name = meta.get("extra", {}).get("feedstock-name") or recipe_name
```
jaimergp, line 1162: At this point `meta` is already rendered as a dict, right? No need to parse the raw text.

mgorny: I'm confused here. FWICS `meta` is just the original YAML, so it can technically have missing keys. There are things like `extra_section` but there's no `recipe_section`, and either way it didn't really seem useful to pass two sections when I can just get them again from `meta`.

jaimergp: I was asking mostly because in some cases (bot infra, IIRC), we are dealing with raw text (unrendered, unparsed Jinja YAML) and then we can't just deal with the dictionary. But looks like are fine here, so marking this as resolved.

conda_smithy/linter/lints.py
```diff
@@ -1139,3 +1141,61 @@ def lint_invalid_workflow_settings(
...
+    else:
+        return
+    if remote.url is None:
+        return
+
+    parsed_url = urlsplit(remote.url)
+    if not parsed_url.netloc:
+        # Maybe it's ssh-style netloc:path.
+        parsed_url = urlsplit(f"git+ssh://{remote.url.replace(':', '/')}")
+    feedstock_name_re = re.compile(r"/conda-forge/([^/]+)-feedstock(?:\.git)?/?")
```
jaimergp: I think there's a setting somewhere in the code where we allow users to specify a different organization. Can we use it here instead of hardcoding `conda-forge`?

conda_smithy/linter/messages/conda_forge.py
```diff
@@ -259,3 +259,17 @@ class WorkflowSettingsSpecificEntryTooLoose(LinterMessage):
...
+
+
+@dataclass(kw_only=True)
+class MismatchedFeedstockName(LinterMessage):
+    """
+    Lint when the feedstock name is mismatched to recipe name.
+    """
+
+    kind = "lint"
+    identifier = "CF-013"
```
jaimergp, line 271: We are intentionally making this lint a conda-forge-only rule, correct? I agree but I'd like to know your rationale.

news/mismatched-feedstock-name-2607.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* A check for ``feedstock-name`` not matching the feedstock repository. (#2607)
```
jaimergp:
```suggestion
* A check for a recipe's ``extra.feedstock-name`` not matching the feedstock repository name. (#2607)
```

conda_smithy/linter/messages/conda_forge.py
mgorny: Yeah, my understanding is that we're resolving a problem specific to conda-forge workflows.

conda_smithy/linter/lints.py
mgorny: Oh, indeed. Curious enough, there's also a `repo_name` field that is used in place of feedstock name in a few places, but it doesn't seem to be used consistently, so I guess it's best to ignore its existence for the scope of this check.

jaimergp (approved):

mgorny:
Thanks!

tests/test_lint_recipe.py
```diff
@@ -5548,5 +5549,158 @@ def test_invalid_workflow_settings(tmp_path):
...
+@mock.patch("conda_smithy.linter.lints.get_repo")
+def test_lint_feedstock_name(
+    repo_mock,
+    tmp_path: Path,
+    remotes: dict[str, str],
+    feedstock_name: str | None,
+    expected: str | None,
+) -> None:
+    recipe = """\
+package:
```
h-vetinari, line 5598: Would be nice to use `textwrap.dedent` in such cases

PR #2610, opened 2026-07-03
Align `py-rattler` pin with `conda-rattler-solver` `0.1.1`

### LINKED ISSUE #2608: Incompatible `py-rattler` pins between `conda-smithy` and `conda`
Currently `conda-smithy` pins `py-rattler` to `0.22`:

https://github.com/conda-forge/conda-smithy-feedstock/blob/d68baf836f7ae1175d07c104821ea46d304374b9/recipe/meta.yaml#L67

However `conda`, [through its dependency on `conda-rattler-solver`]( https://github.com/conda-forge/conda-feedstock/blob/c3c3469c8675ce4fc7e665c4bdcc80c029df6a2c/recipe/meta.yaml#L76 ), is [pinning `py-rattler` to `0.25`]( https://github.com/conda-forge/conda-rattler-solver-feedstock/blob/aec93f73f5ed8a1ac8a60f8a67be28f77d087e5d/recipe/recipe.yaml#L26 ).

As `conda-smithy` requires `conda`, these incompatible `py-rattler` pins mean a newer `conda` and `conda-smithy` cannot be installed together.

Could we update `conda-smithy` to a compatible `py-rattler` version with `conda`?

How can we ensure this issue does not reoccur? Could we drop the pin in `conda-smithy` (just depending on `py-rattler` unpinned)? Or are there other options worth considering here?

jakirkham:
Recently [`conda` added a dependency on `conda-rattler-solver`]( https://github.com/conda-forge/conda-feedstock/blob/c3c3469c8675ce4fc7e665c4bdcc80c029df6a2c/recipe/meta.yaml#L76 ). As [`conda-rattler-solver` depends on (and pins) `py-rattler`]( https://github.com/conda-forge/conda-rattler-solver-feedstock/blob/aec93f73f5ed8a1ac8a60f8a67be28f77d087e5d/recipe/recipe.yaml#L26 ), it is possible for `conda` to pull in a different pining for `py-rattler` via `conda-rattler-solver`.

To address, this bump the `py-rattler` pin to match recent `conda-rattler-solver` release `0.1.1` needed by recent `conda` release `26.5.3`. That way newer versions of `conda` can be installed with `conda-smithy` (instead of older versions of `conda` being picked up.

<hr>

Checklist

Fixes https://github.com/conda-forge/conda-smithy/issues/2608
Part of https://github.com/conda-forge/conda-smithy/issues/2589


jakirkham:
Looks like [CI installs `conda-forge-tick`]( https://github.com/conda-forge/conda-smithy/blob/a483fac836304b2c6234eb297adfd3faaa73b4e7/.github/workflows/tests.yml#L33 ) before installing `conda-smithy`. However [`conda-forge-tick` depends on `conda-smithy`]( https://github.com/conda-forge/conda-forge-tick-feedstock/blob/04a710c28d2a4cdbeb323034178ac62dae91d2d1/recipe/meta.yaml#L43 ). This is problematic given the development version of `conda-smithy` from this repo has not been installed yet; so, this step is effectively installing the last `conda-smithy` release, which we don't want in `conda-smithy`'s own CI. Think we need to separate these two steps.

h-vetinari:
Sounds like [conda-forge-feedstock-ops](https://github.com/conda-forge/conda-forge-feedstock-ops-feedstock/blob/481f6b01333c85cf9904c140e8f884e10160397c/recipe/meta.yaml#L37) needs an update first

h-vetinari:
Sigh, we have a bootstrapping problem here; https://github.com/conda-forge/conda-forge-feedstock-ops/pull/85 cannot update the py-rattler pin, because it depends on smithy. Sounds like we need to break the cycle, e.g. by first releasing smithy with a new pin (disabling the tests that require `conda-forge-tick` until feedstock-ops can be updated to match).

Maybe there's better alternatives I'm not seeing.

CC @beckermr @jaimergp

beckermr:
The conda-forge-tick integrations should be optional. If not, we should fix them so they are, and mark the tests as xfail. We can deal
with reenabling the tests later.

environment.yml
```diff
@@ -40,5 +40,5 @@ dependencies:
   - jsonschema
   - exceptiongroup
   # py-rattler's APIs are subject to change, pin to minor versions
-  - py-rattler>=0.22,<0.23
+  - py-rattler>=0.25.0,<0.26.0a0
```
jaimergp: If the API changes didn't break us, why restrict to latest minor? We can do

```suggestion
  - py-rattler>=0.22.0,<0.26.0a0
```

.github/workflows/tests.yml
```diff
@@ -30,14 +36,14 @@ jobs:
...
       - name: install conda-smithy
         run: |
-          conda uninstall --force --yes conda-smithy
           python -m pip install -v --no-build-isolation -e .
-          git config --global user.email "smithy@smithy.smithy"
-          git config --global user.name "smithy"
+
+      - name: install conda-forge-tick
+        run: |
+          conda install --yes 'conda-forge-tick>=2026.0'
```
jaimergp, line 48: Shouldn't this be part of `environment.yml`?

h-vetinari: No, because it's not installable on windows

jaimergp: Hm, worth checking #2525 again for these things, but ok for now. A comment would be welcome though.

```suggestion
          # Linux-only, can't be specified on environment.yml
          conda install --yes 'conda-forge-tick>=2026.0'
```

news/2610-bump.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Bump ``py-rattler`` upper bound pin to 0.26. (#2610)
```
h-vetinari, line 7: That's... IMO a somewhat misleading description for `py-rattler>=0.22.0,<0.26.0a0`, because it reads as if 0.26 is included (or, if reading with less than 100% concentration, might even sound like it's pinned to 0.26).

How about "allow wider range of `py-rattler` versions"?

environment.yml
```diff
@@ -41,5 +41,5 @@ dependencies:
   - jsonschema
   - exceptiongroup
   # py-rattler's APIs are subject to change, pin to minor versions
-  - py-rattler>=0.22,<0.23
+  - py-rattler>=0.22.0,<0.26.0a0
```
h-vetinari, line 44: Keeping the lower bound as-is is risky, because the circular dependence with feedstock-ops means that we're not actually ever testing the most recent py-rattler build here when we lift the upper bound. I think we might have to add a separate CI job that tests with newest py-rattler (which is its own headache to specify or implement).

h-vetinari: Or alternatively, run the test suite also without `conda-forge-tick`, which is what brings in the feedstock ops.

PR #2617, opened 2026-07-09
Allow v1 dollar schema

bollwyvl:
Checklist

References:
- [x] https://github.com/prefix-dev/rattler-build/pull/2645
- [x] https://github.com/prefix-dev/recipe-format/pull/82
- [x] https://github.com/conda/ceps/pull/186

Changes:
- [x] prepend `$schema` to allowed `recipe.yaml` fields
- [x] fix a typing issue on linter message
- [x] update `conda-forge.yaml` schema to allow optional `$schema`


news/v1-dollar-schema.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Allow ``recipe.yaml`` to declare its schema inline with ``$schema`` instead of a comment
```
jaimergp:
```suggestion
* Linter: Allow ``recipe.yaml`` to declare its schema inline with ``$schema`` instead of a comment. (#2617)
```

jaimergp:
@bollwyvl, do you want to mark this ready for review? LGTM!

bollwyvl:
Should we also allow `$schema` in `conda-forge.yml`?

jaimergp:
Good idea, go for it!

bollwyvl:
I can't figure out how to make the linter happy with this change. `pre-commit` is a menace.

bollwyvl:
(rolled back to the passing [fdc13e2](https://github.com/conda-forge/conda-smithy/pull/2617/commits/fdc13e21783bb0f48566773df35ca1aa53dc67c9))

bollwyvl:
At least it looks like the test fails aren't my fault...

bollwyvl:
Hooray! back to :heavy_check_mark:

bollwyvl:
Hooray! back to ✔️

tests/test_lint_recipe.py
```diff
@@ -4340,6 +4340,7 @@ def test_lint_recipe_v1_invalid_schema_version():
     with tempfile.TemporaryDirectory() as tmpdir:
         with open(os.path.join(tmpdir, "recipe.yaml"), "w") as f:
             f.write(textwrap.dedent("""
+                    $schema: ./schema.json
```
h-vetinari, line 4344: It looks like we don't check that `schema.json` exists, or where should this come from? IMO that would be a kinda obvious thing to lint on, if we want to allow `$schema`?

bollwyvl: While _usually_ a URL, it's technically a URI. Many of the language servers will even pop a warning if it's not one that has been seen before (or from an unknown host). That being said, a larger number of the lint/hint rules _could_ be written as schema (it has `not`, `oneOf`, `if/then`, etc), and Some Day, smithy could ship its own v1 schema that extended the upstream one at `../.ci_support/schema.json` or something.

Until then, as it is discarded everywhere, I feel like the initial goal is being able to use `$schema` instead of magic comments; if folk start building recipe conventions around it, that's awesome.

h-vetinari: My point is: where should the language server find `schema.json`? I'll admit I'm not familiar with this, so you may have to explain it as if talking to a small child 😅

bollwyvl: Sure: the most common is going to be a URL, going from:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/prefix-dev/recipe-format/main/schema.json
```

to:

```yaml
$schema: https://raw.githubusercontent.com/prefix-dev/recipe-format/main/schema.json
```

In both cases, a first-time user would be prompted (maybe visually with squiggles under the `$schema: ` line) as to whether the URL should be resolved. If accepted, that schema would get used per keystroke for diagnostics, tab completion, etc. 

If denied, an editor may fall back to a built-in schema, such as vendored by (or pre-trusted :crying_cat_face: ) [schema store](https://www.schemastore.org/), which allows for redirecting to _other_ places.

On _any_ document, schema validation can be disabled, leaving only "well-formedness" (kind of a stretch in YAML, but there you go) checks.

To the specific point: relative URIs are also supported by most tools as values of `$schema`, and indeed, it _would_ be relevant to check if the file exists in this case, if we knew how it was going to get used. For now, we don't know what kind of patterns will be useful, with some possible ideas described above, only that e.g. `rattler-build` _won't_ propagate this to some output data, which would complicate otherwise-useful local URLs.

h-vetinari: Thanks for the explanation.

> relative URIs are also supported by most tools as values of `$schema`, and indeed, it _would_ be relevant to check if the file exists in this case, if we knew how it was going to get used.

What I don't understand is how the manner of downstream usage affects the fact that `$schema: foo.json` is pointless if `foo.json` cannot be found. To me, both `$schema: ./path/to/non_existent.json` and `$schema: https://broken.link/schema.json` should cause a linter error IMO (or at the very least, a warning). Otherwise, we're just leaving known-bad configuration in the file, which to me is really counter to how the linter should work.

What do we gain by being lenient on this (aside from avoiding implementation effort, I'll give you that)? Is it that language servers may have other ways of finding `foo.json` (like the schema store) that are hard/impossible for us to predict?

bollwyvl: > hard/impossible for us to predict?

Yep, that's kind of my feeling at this point. If it's a non-existent file, some folks will click "accept," and then see squiggles on the value, but it might still `rattler-build build`.

h-vetinari: I'm OK with this PR, despite some unease about the lack of checking for what values are behind the `$schema` key. I'll leave this to @jaimergp / @wolfv / @baszalmstra or whoever else is more familiar with this topic.

PR #2622, opened 2026-07-15
Define build_platform and pass it to conda-build and rattler-build

jaimergp:
Checklist

Comes from https://github.com/conda-forge/conda-forge.github.io/issues/1940#issuecomment-4984071281.

Changes:

- On Windows, define `%BUILD_PLATFORM%` directly from GHA/AZP job matrix
- We also implement a fallback where BUILD_PLATFORM is inferred from PROCESSOR_ARCHITECTURE
- For `conda-build`, we export `CONDA_SUBDIR=%BUILD_PLATFORM%`
- For `rattler-build`, we pass `--build-platform %BUILD_PLATFORM%`

Needs these PRs for full effect:

- https://github.com/conda/conda-build/pull/6047
- https://github.com/prefix-dev/rattler-build/pull/2664
- https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/423
- https://github.com/conda/conda/pull/16417

### DIFF
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -189,7 +189,7 @@ call :end_group
 
 :: Build the recipe
 echo Building recipe
-{%- if conda_build_tool.startswith(("conda-build", "mambabuild") %}
+{%- if conda_build_tool.startswith(("conda-build", "mambabuild")) %}
 set "_OLD_CONDA_SUBDIR=%CONDA_SUBDIR%"
 set "CONDA_SUBDIR=%BUILD_PLATFORM%"
 {%- endif %}
@@ -207,7 +207,7 @@ conda-build.exe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml --suppress-varia
 rattler-build.exe build --recipe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml %EXTRA_CB_OPTIONS% --build-platform %BUILD_PLATFORM% --target-platform %HOST_PLATFORM%
 {%- endif %}
 if !errorlevel! neq 0 exit /b !errorlevel!
-{%- if conda_build_tool.startswith(("conda-build", "mambabuild") %}
+{%- if conda_build_tool.startswith(("conda-build", "mambabuild")) %}
 set "_OLD_CONDA_SUBDIR="
 set "CONDA_SUBDIR=%_OLD_CONDA_SUBDIR%"
 {%- endif %}
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -68,7 +68,8 @@ if "%PIXI_CACHE_DIR%"=="%MINIFORGE_HOME%" (
 )
 move /y pixi.toml pixi.toml.bak
 set "arch=64"
-if "%PROCESSOR_ARCHITECTURE%"=="ARM64" set "arch=arm64"
+{#- TODO: Enable this line again once we have bootstrapped all the dependencies for a full base environment on win-arm64; emulate win-64 for now  #}
+{#- if "%PROCESSOR_ARCHITECTURE%"=="ARM64" set "arch=arm64" #}
 powershell -NoProfile -ExecutionPolicy unrestricted -Command "(Get-Content pixi.toml.bak -Encoding UTF8) -replace 'platforms = .*', 'platforms = [''win-%arch%'']' | Out-File pixi.toml -Encoding UTF8"
 :: Git on Windows needs to run post link scripts to properly set up SSL certificates
 pixi config set --global run-post-link-scripts insecure
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -193,7 +193,7 @@ conda-build.exe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml --suppress-varia
 {%- elif conda_build_tool == "conda-build" %}
 conda-build.exe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml --suppress-variables %EXTRA_CB_OPTIONS%
 {%- elif conda_build_tool == "rattler-build" %}
-rattler-build.exe build --recipe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml %EXTRA_CB_OPTIONS% --target-platform %HOST_PLATFORM%
+rattler-build.exe build --recipe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml %EXTRA_CB_OPTIONS% --build-platform %BUILD_PLATFORM% --target-platform %HOST_PLATFORM%
 {%- endif %}
 if !errorlevel! neq 0 exit /b !errorlevel!
 
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -183,6 +183,10 @@ call :end_group
 
 :: Build the recipe
 echo Building recipe
+{%- if conda_build_tool.startswith(("conda-build", "mambabuild") %}
+set "_OLD_CONDA_SUBDIR=%CONDA_SUBDIR%"
+set "CONDA_SUBDIR=%BUILD_PLATFORM%"
+{%- endif %}
 {%- if conda_build_tool == "mambabuild" %}
 conda-mambabuild.exe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml --suppress-variables %EXTRA_CB_OPTIONS%
 {%- elif conda_build_tool == "conda-build+conda-libmamba-solver" %}
@@ -197,6 +201,10 @@ conda-build.exe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml --suppress-varia
 rattler-build.exe build --recipe "{{ recipe_dir }}" -m .ci_support\%CONFIG%.yaml %EXTRA_CB_OPTIONS% --build-platform %BUILD_PLATFORM% --target-platform %HOST_PLATFORM%
 {%- endif %}
 if !errorlevel! neq 0 exit /b !errorlevel!
+{%- if conda_build_tool.startswith(("conda-build", "mambabuild") %}
+set "_OLD_CONDA_SUBDIR="
+set "CONDA_SUBDIR=%_OLD_CONDA_SUBDIR%"
+{%- endif %}
 
 call :start_group "Inspecting artifacts"
 :: inspect_artifacts was only added in conda-forge-ci-setup 4.9.4
--- a/conda_smithy/templates/run_win_build.bat.tmpl
+++ b/conda_smithy/templates/run_win_build.bat.tmpl
@@ -131,6 +131,12 @@ if !errorlevel! neq 0 exit /b !errorlevel!
 {%- endif %}
 
 :: Set basic configuration
+if [%BUILD_PLATFORM%]==[] (
+    echo Defining build platform
+    if "%PROCESSOR_ARCHITECTURE%"=="AMD64" set "BUILD_PLATFORM=win-64"
+    if "%PROCESSOR_ARCHITECTURE%"=="ARM64" set "BUILD_PLATFORM=win-arm64"
+    if "%PROCESSOR_ARCHITECTURE%"=="x86"   set "BUILD_PLATFORM=win-32"
+)
 echo Setting up configuration
 setup_conda_rc .\ ".\{{ recipe_dir }}" .\.ci_support\%CONFIG%.yaml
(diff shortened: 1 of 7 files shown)


jaimergp:
cc @isuruf @baszalmstra

beckermr (approved):

isuruf:
You can just install the win-64 version of rattler-build on a win-arm64 machine. I don't see why this is necessary.

baszalmstra:
This is necesary because win-64 rattler-build will default to a build environment that is win-64 instead of arm64.

isuruf:
> This is necesary because win-64 rattler-build will default to a build environment that is win-64 instead of arm64.

Which is exactly what we want.

baszalmstra:
This is my understanding:

We have a bit of a boostrapping problem. We want to use native win-arm64 runners. We need `conda-forge-ci-setup` to run on win-arm64 runners. `conda-forge-ci-setup` is an arch specific package with no support for win-arm64. This would require,  amongst others, packages like `conda` which depends on arch specific python packages like `py-rattler` that need to be build for `win-arm64`. Since we do not have native win-arm64 runners yet, we need to cross compile them (which is what we have been doing for other packages). But that requires `cross-python`, which is something we do not want to support for `win-arm64` (because its hacked together as far as I understand).

The solution we discussed in the last conda meeting was to enable using `win-arm64` runners but install `conda-forge-ci-setup` for `win-64`. Emulating the ci setup. This will bring in `rattler-build` (or `conda-build`) for `win-64`, but we use it to build a native package (not cross-compiling from `win-64` -> `win-arm64`). So the `build-platform` and `host-platform` will both be `win-arm64` but the build will be orchestrated by `win-64` `rattler-build`.

But, `rattler-build` compiled for `win-64` will think the native platform (`build-platform`) is `win-64`. So we need to force it to use `win-arm64` as the build-platform because we are actually running it on a `win-arm64` runner.

@jaimergp Is this correct?

jaimergp:
> You can just install the win-64 version of rattler-build on a win-arm64 machine. I don't see why this is necessary.

That's what we are doing, but as Bas said:

> But, rattler-build compiled for win-64 will think the native platform (build-platform) is win-64. So we need to force it to use win-arm64 as the build-platform because we are actually running it on a win-arm64 runner.

---

> @jaimergp Is this correct?

100% spot on.

---

> > This is necesary because win-64 rattler-build will default to a build environment that is win-64 instead of arm64.
> Which is exactly what we want.

No, because in that case we have `build_platform != target_platform` and then we hit a wall with `cross_python` not being available. If we define build_platform as win-arm64, then we can build Python packages that use `recipe.yaml` (e.g. see https://github.com/conda-forge/menuinst-feedstock/pull/61). `conda-build` seems to be picking up the build platform just fine (even if running from `win-64` envs) and doesn't need this (example at https://github.com/conda-forge/pycosat-feedstock/pull/37).

---

I also considered creating a separate `win-arm64` environment for `rattler-build` alone and executing the binary without activation, but that diff is larger than needed for this workaround. And worst case, the `--build-platform` flag is redundant and doesn't add anything.

baszalmstra:
Ill try this with `py-rattler`.

isuruf:
> No, because in that case we have build_platform != target_platform and then we hit a wall with cross_python not being available. If we define build_platform as win-arm64, then we can build Python packages that use recipe.yaml (e.g. see https://github.com/conda-forge/menuinst-feedstock/pull/61). conda-build seems to be picking up the build platform just fine (even if running from win-64 envs) and doesn't need this (example at https://github.com/conda-forge/pycosat-feedstock/pull/37).

This should be fixed instead of being worked around by a hack. Maybe change `build_platform != target_platform` to `build_platform != target_platform and not win`

jaimergp:
But I don't see it as a hack 🤔 The `build_platform` _is_ `win-arm64` because we are using an ARM runner on Github Actions (`windows-11-arm`). If anything the hack is on having the build tool installed on a `win-64` environment to work around the bootstrapping problem for now. Once we have all the necessary dependencies available then we can install `base` for `win-arm64` and then the `--build-platform` directive will be redundant (but not incorrect).

jaimergp:
Ok after talking with Isuru I've understood why there's a problem with this approach: `cmd.exe` thinks is running on `win-64` (see [this line](https://github.com/conda-forge/pycosat-feedstock/actions/runs/29444322022/job/87450867587?pr=37#step:6:1095) despite everything else being ARM). We'd need to find a way for a win-64 process to start a fresh win-arm64 process, and I don't know if that's possible.

Putting this in draft until then.

jaimergp:
Ok, there seems to be a way to pass the desired target architecture for the subprocess on Windows: [`start /machine arm64`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/start). So if we contribute a fix to conda-build and rattler-build, this should tell CMD what it's actually running on. Docs do say that's in preview, so we need to check which version the CI runners are on. cc @baszalmstra.

baszalmstra:
But does that mean that if a win-64 process starts a sub-process in emulation that that subprocess also runs under win-64?

isuruf:
> But does that mean that if a win-64 process starts a sub-process in emulation that that subprocess also runs under win-64?

Yes, this is true universally. For eg: Rosetta 2 and qemu userspace.

baszalmstra:
So after some investigation I think this is false. 

 Launching cmd.exe does not make it inherently win-64. On a Windows-on-ARM machine, resolving C:\Windows\System32\cmd.exe normally launches the native ARM64 (or Arm64X) system cmd.exe;
 a child can have a different architecture from its parent.

However, `PROCESSOR_ARCHITECTURE` is not a reliable way to establish that: it is just an inherited/mutable environment variable. In particular, rattler-build currently forwards its own value:

 ```rust
   vars.insert(var.to_string(), std::env::var(var).ok());
 ```

So an emulated win-64 rattler-build could pass along `AMD64` even if its cmd.exe child is native ARM64.

baszalmstra:
> Yes, this is true universally. For eg: Rosetta 2 and qemu userspace. 

How though? If you have an executable that is only available as arm64, rosetta would have to convert arm64 to x64 and back?

isuruf:
> How though? If you have an executable that is only available as arm64, rosetta would have to convert arm64 to x64 and back?

Most binaries on Apple silicon are fat and have both arm64 and x86_64 slices.

baszalmstra:
> So after some investigation I think this is false.

I retract that statement after trying it, I think you're right

baszalmstra:
> Putting this in draft until then.

But is this really a problem for now? We would use this to build the low level packages and can switch to completely native after that.

The compiler activations etc are still properly targeting win-arm64 (as can be seen a few lines up in the logs) which I think is what really matters.

jaimergp:
> But is this really a problem for now? We would use this to build the low level packages and can switch to completely native after that.

I think we've been lucky so far, but a `cmd.exe` identifying as AMD64 may bring problems down the line. 

I've run some [experiments](https://github.com/jaimergp/windows-11-arm-tests) to see how can we can spawn ARM64 CMDs via `start` but that command doesn't really allow for its non-interactive usage (it launches the process and forgets, even with `/wait`, so there's no stdout/stderr connection). Then I learned that `%PROCESS_ARCHITECTURE%` gets inherited even if the underlying process changes architecture, and it's more reliable to query powershell directly with `[System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture`. Proof at https://github.com/jaimergp/windows-11-arm-tests/actions/runs/29518099786/job/87688097320. I think it'd be fine to override `PROCESS_ARCHITECTURE` if the powershell source of truth disagrees.

So an alternative (not super pretty, tbh) is to run a pure ARM64 intermediate launcher (e.g. a Python ARM64) that then launches the CMD executable for the batch script with the right architecture. For that we would need to create a `win-arm64` environment for that need only.

jaimergp:
Ok, as per the results in https://github.com/jaimergp/windows-11-arm-tests/actions/runs/29523869103/job/87707238242, there's a way to do it with `start`:

```python
# /b keeps it headless, /wait propagates the exit code
native_wrapper = f"start /b /wait /machine arm64 cmd.exe BATCH_FILE_HERE"
# shell=True because 'start' is a cmd.exe built-in
result = subprocess.run(
    native_wrapper,
    shell=True,
    capture_output=True,
    text=True
)
```

h-vetinari:
Not specifically about the win-arm64 migration (and I'll admit I'm not up to speed with the fat binary vs. emulation vs. xyz stuff), I think we should just generally pass the `--build-platform` to rattler-build; I recently ran into an [issue](https://github.com/prefix-dev/rattler-build/issues/2632) where a v1 translation couldn't use `build_platform` in the recipe `context:`, and according to @wolfv, the only way rattler can know about this is by passing the respective CLI argument.

jaimergp:
So as we stand, I think this PR still has merit, but both conda-build and rattler-build need to patch how CMD is invoked:

- https://github.com/conda/conda-build/pull/6047
- https://github.com/prefix-dev/rattler-build/pull/2664

isuruf:
It seems like we get `build_platform == win-arm64` on a x64 emulated machine on windows-arm64. This is https://github.com/python/cpython/issues/98962

We should first fix that in conda, so that the build_platform reported is correct. Otherwise you get platform == win-arm64 on a win-64 installation of conda. cc @zooba

isuruf:
rattler/rattler-build doesn't have this problem because it correctly uses the rust functions and not python

zooba:
> It seems like we get `build_platform == win-arm64` on a x64 emulated machine on windows-arm64. This is [python/cpython#98962](https://github.com/python/cpython/issues/98962)

The answer here is what I was saying in the linked issue - either you're querying the "architecture of subprocesses that can be launched on this machine" (`platform`) or the "architecture of the running process" (`sys`).

Windows applies emulation layers at every process boundary, so every time you launch a new process it'll check the executable file and see which subsystem it should run under. This has been the design since 16/32-bit were the options, and has been the same for 64/32-bit and now for ARM64/64/32 (on ARM processors). It's always layered for maximum compatibility.

My recommendation would be to always specify the architecture something should build for (CLI arg or env var of your own design), allow overriding it (probably the same arg but exposed to the user/CI script), and default to `sysconfig.get_platform` (a.k.a. the build target that will be compatible with the running process).

All the Windows compilers are cross-compiling all the time, which is largely why you don't see special support for it - either it's so transparent that it's ignored/forgotten, or it's trivially supported with a single flag, and occasionally it requires a flag and a directory. Swapping the x86 _built_ compiler (the default) for the x64 one makes no difference to the output unless you tell it to build for x64 (which `vcvarsall.bat` will do for you; hence, transparent). Rust likewise makes you specify the target triple, which encodes the information needed here.

So assume that you need to always specify the target platform, use `sysconfig.get_platform` as a default, but override it in this case so that the x64 environment can be used to build ARM64 binaries. conda-forge has got _by far_ the best setup for being able to do this easily, you should be able to get a really long way building on regular x64 machines this way.

jaimergp:
Upstream PRs in conda and conda-build are now merged and scheduled for the July release in the coming days. Marking this as ready again, and will come in full effect once the tools have new releases. This only impacts win-arm64 PRs so I guess we can merge before tools are ready though, and included it in the next smithy release.

conda_smithy/templates/run_win_build.bat.tmpl
```diff
@@ -130,6 +131,12 @@ if !errorlevel! neq 0 exit /b !errorlevel!
 {%- endif %}
 
 :: Set basic configuration
+if [%BUILD_PLATFORM%]==[] (
+    echo Defining build platform
+    if "%PROCESSOR_ARCHITECTURE%"=="AMD64" set "BUILD_PLATFORM=win-64"
+    if "%PROCESSOR_ARCHITECTURE%"=="ARM64" set "BUILD_PLATFORM=win-arm64"
+    if "%PROCESSOR_ARCHITECTURE%"=="x86"   set "BUILD_PLATFORM=win-32"
+)
```
isuruf: Why do we need this fallback?

jaimergp: Off-CI workflows, like the `build-locally` script.

isuruf: They are set by conda-forge-ci-setup even for `build-locally`.

jaimergp: I was going to argue that ci-setup [does](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/df6d1792ab1fc496d922c6dd9c7fffe6e338f867/recipe/run_conda_forge_build_setup_win.bat#L115) it via `conda` which is sensitive to the parent process, but actually so does this one via `build-locally`. It would only make sense if the script was run directly without any Python, so yea, fair enough, I'll remove it here.

jaimergp: Done at [9e867ad](https://github.com/conda-forge/conda-smithy/pull/2622/commits/9e867addd3574bf61d3ba73a91aaca7a44f586fa)

isuruf (approved):

h-vetinari:
> Upstream PRs in conda and conda-build are now merged and scheduled for the July release in the coming days.

Do we need to raise the lower bounds for conda/conda-build here?

PR #2623, opened 2026-07-17
Add test for executable bits of rendered feedstock content.

h-vetinari:
Broken out from #2596; it contains the fix for executable AZP yaml files, and a test that tests the expected executable bit of all rendered files. The test depends on the clean-up (not least because it needs to know which files are supposed to be executable); a04bcc288dbc509712e45cd8882f6c05e44c61d9 can be left out if preferred, it's a minor clean-up on top to guard more templates against conflicting rerenders.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -210,8 +210,22 @@ GITHUB_ACTIONS_RUNS_ON = {
 }
 
 
-ALL_EXECUTABLE_FILES = [
-    ".circleci/checkout_merge_commit.sh",
+NON_EXECUTABLE_TEMPLATES = [
+    ".azure-pipelines/azure-pipelines-linux.yml",
+    ".azure-pipelines/azure-pipelines-osx.yml",
+    ".azure-pipelines/azure-pipelines-win.yml",
+    ".drone.yml",
+    ".travis.yml",
+    "README.md",
+    "appveyor.yml",
+    "azure-pipelines.yml",
+    "circle.yml",
+    "github-actions.yml",
+    "pixi.toml",
+]
+
+
+EXECUTABLE_TEMPLATES = [
     ".scripts/SetPageFileSize.ps1",
     ".scripts/build_steps.sh",
     ".scripts/create_conda_build_artifacts.bat",
@@ -219,10 +233,16 @@ ALL_EXECUTABLE_FILES = [
     ".scripts/create_pagefile.bat",
     ".scripts/create_pagefile.sh",
     ".scripts/free_disk_space.sh",
-    ".scripts/logging_utils.sh",
     ".scripts/run_docker_build.sh",
     ".scripts/run_osx_build.sh",
     ".scripts/run_win_build.bat",
+]
+
+
+ALL_EXECUTABLE_FILES = EXECUTABLE_TEMPLATES + [
+    ".circleci/checkout_merge_commit.sh",
+    ".circleci/fast_finish_ci_pr_build.sh",
+    ".scripts/logging_utils.sh",
     "build-locally.py",
 ]
 
@@ -1736,7 +1756,7 @@ def _render_template_files(forge_config, jinja_env, template_files, forge_dir):
         # ensure trailing newline
         if new_file_contents[-1] != "\n":
             new_file_contents += "\n"
-        if target_fname in get_common_scripts(forge_dir) and os.path.exists(
+        if target_fname in iter_all_templates(forge_dir) and os.path.exists(
             target_fname
         ):
             with open(target_fname, encoding="utf-8") as fh:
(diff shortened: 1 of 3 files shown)


conda_smithy/configure_feedstock.py
```diff
@@ -210,6 +210,43 @@
...
+    ".scripts/run_osx_build.sh",
+    ".scripts/run_win_build.bat",
+]
+
+
+ALL_EXECUTABLE_FILES = EXECUTABLE_TEMPLATES + [
+    ".circleci/checkout_merge_commit.sh",
+    ".circleci/fast_finish_ci_pr_build.sh",
+    ".scripts/logging_utils.sh",
+    "build-locally.py",
```
jaimergp, line 250: I think this one needed to be executable too because it's currently starting with a polyglot script:

```
#!/bin/sh
"""exec" "python3" "$0" "$@" #"""  # fmt: off # fmt: on
```

conda_smithy/configure_feedstock.py
```diff
@@ -3003,19 +3055,9 @@ def clear_variants(forge_dir):
             remove_file(config)
 
 
-def get_common_scripts(forge_dir):
```
jaimergp, line 3006: Can't just remove it, needs to be deprecated first.

conda_smithy/configure_feedstock.py
jaimergp, line 250: We should have a test to make sure that all files that can end up in the feedstock (via direct copy or templating) are classified in one of these.

h-vetinari: It is listed under `ALL_EXECUTABLE_FILES`? So it gets checked whether it was made executable. Perhaps you misread something? That, or I'm not understanding your comment.

h-vetinari:
pre-commit.ci autofix

conda_smithy/configure_feedstock.py
h-vetinari: This was a PITA, but good to have done, because a fair few of these were not correct or missing. 😅

conda_smithy/configure_feedstock.py
h-vetinari: Done.

conda_smithy/configure_feedstock.py
jaimergp: Yea, you are right. I thought this was a template but it's not.

tests/test_configure_feedstock.py
```diff
@@ -760,6 +766,77 @@ def test_secrets(py_recipe, jinja_env):
...
+
+    configure_feedstock.main(
+        forge_file_directory=recipe_dir,
+        forge_yml=forge_yml,
+        no_check_uptodate=True,
+        commit=True,
+    )
+
+    # sanity check for pytest failure logs: check content of recipe folder
+    subprocess.call(["ls", "-lla"], cwd=recipe_dir, stdout=sys.stderr)
```
jaimergp: This makes the test Unix-only, but it's not skipped on Windows. And the test doesn't probably apply to Windows anyway, so let's mark it as skipped on this platform?

tests/test_configure_feedstock.py
```diff
@@ -760,6 +766,77 @@ def test_secrets(py_recipe, jinja_env):
...
+    for file in iter_files(recipe_dir):
+        fname = str(file).replace("\\", "/")
+        # we expect all executable files to have the exec bit,
+        # and all non-executable files to not have it
+        assert is_executable(file) == (fname in ALL_EXECUTABLE_FILES)
+
+        # check that we know the provenance of all files that got generated
+        allowlist = [
+            # variant configs
+            ".ci_support/",
```
jaimergp, line 828: Aren't these generated too? Yes, they are only YAML, but they shouldn't have an executable bit, right?

tests/test_configure_feedstock.py
```diff
@@ -760,6 +766,77 @@ def test_secrets(py_recipe, jinja_env):
         )
 
 
+@pytest.mark.parametrize("provider", ["azure", "github_actions", "drone", "travis"])
```
jaimergp, line 769:
```suggestion
@pytest.mark.parametrize("provider", ["azure", "github_actions", "drone", "travis"])
@pytest.mark.skipif(os.name == "nt", reason="Executable bits relevant on Unix only")
```

tests/test_configure_feedstock.py
h-vetinari: They don't have an exec bit, that's being tested just a few lines up. The allowlist here is for the test you asked, about testing that we statically know all the kinds of files that are being generated.

For the variants, we don't have a common pattern how the file names will be populated (in contrast to all the other support files, which always have a predictable path, at least per provider), so we can't expect them to be part of a universal `ALL_SUPPORT_FILES`.

tests/test_configure_feedstock.py
h-vetinari: It passes on my windows box, but then I do have unix-y utilities on my path. I think it'd make sense to make this debug output platform-aware, rather than skipping this test for some combinations

tests/test_configure_feedstock.py
h-vetinari: As mentioned above, the tests can pass on windows (modulo presence of `ls`, which I can fix on Monday)

h-vetinari: Fixed by making the test runnable on windows also without the unix-y binaries from git.

jaimergp (approved):

PR #2625, opened 2026-07-21
correct free_disk_space hint to list supported values

pavelzw:
Checklist

observed in https://github.com/conda-forge/polars-feedstock/pull/386#issuecomment-5032153911


beckermr (changes requested):
This PR is wrong. The doc string has a deprecation notice and then documents the old options. Instead of removing the old options, the doc string needs to specify that the new options are different than the old ones.

beckermr (approved):

PR #2631, opened 2026-07-24
Artifact upload improvements

### LINKED ISSUE #2365: `create_conda_build_artifacts.sh` ends up using two layers of artifact compression
### Comment:

Not sure if this is the right place to discuss this; chose it because `create_conda_build_artifacts.sh` is created by smithy.

I've noticed that the CI build artifacts are distributed as a zip archive inside another zip archive, with both archives being compressed separately. Against all odds, the second layer of compression actually shaves some data from big archives (probably because it ends up compressing the metadata). However, for really big archives it's quite inconvenient since you effectively need to decompress twice.

Depending on what we're aiming for, I have two suggestions to choose from:

1. [Disable artifact compression](https://github.com/actions/upload-artifact?tab=readme-ov-file#altering-compressions-level-speed-v-size) at `upload-artifact` level. Then the internal archive will remain compressed, and since the outer archive will be uncompressed, it will be suitable for efficiently accessing the inner archive without unpacking it to the filesystem. However, it will be slightly larger (~130M for PyTorch).
2. Disable inner archive compression. We will still have to unpack it, but I think this will improve the overall compression, since we'd be benefiting from cross-file matches.

mgorny:
Checklist

Fixes #2365

Changes so far:
- [x] disable creating the outer .zip archive for GHA artifacts
- [x] change the archive format to .tar.zst
- [x] fix exclude/include patterns to be less greedy and account for rattler-build better
- [x] split `work` directory from built artifacts?
- [x] extend the above to Windows
- [x] handle (partial) archive creation failures gracefully

### DIFF
--- a/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
@@ -84,6 +84,7 @@ jobs:
           export BLD_ARTIFACT_PREFIX=conda_pkgs_failed
           # Archive the CONDA_BLD_PATH environments only when the job fails
           export ENV_ARTIFACT_PREFIX=conda_envs
+          export WRK_ARTIFACT_PREFIX=conda_work
         fi
         ./.scripts/create_conda_build_artifacts.sh
     displayName: Prepare conda build artifacts
@@ -95,6 +96,13 @@ jobs:
     inputs:
       targetPath: $(BLD_ARTIFACT_PATH)
       artifactName: $(BLD_ARTIFACT_NAME)
+
+  - task: PublishPipelineArtifact@1
+    displayName: Store conda work directoryartifacts
+    condition: not(eq(variables.WRK_ARTIFACT_PATH, ''))
+    inputs:
+      targetPath: $(WRK_ARTIFACT_PATH)
+      artifactName: $(WRK_ARTIFACT_NAME)
 
   - task: PublishPipelineArtifact@1
     displayName: Store conda build environment artifacts
--- a/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-linux.yml.tmpl
@@ -73,13 +73,15 @@ jobs:
   - script: |
         export ARTIFACT_STAGING_DIR="$(Build.ArtifactStagingDirectory)"
         # Archive everything in CONDA_BLD_PATH except environments
-        export BLD_ARTIFACT_PREFIX=conda_artifacts
+        export BLD_ARTIFACT_PREFIX=conda_pkgs
         export CI=azure
         export CI_RUN_ID=$(build.BuildNumber).$(system.JobAttempt)
         export CONDA_BLD_PATH=$(build_workspace_dir)
         export FEEDSTOCK_NAME=$(basename ${BUILD_REPOSITORY_NAME})
         export MINIFORGE_HOME=$(tools_install_dir)
         if [[ "$AGENT_JOBSTATUS" == "Failed" ]]; then
+          # Use a distinct prefix on failure (to match GHA).
+          export BLD_ARTIFACT_PREFIX=conda_pkgs_failed
           # Archive the CONDA_BLD_PATH environments only when the job fails
           export ENV_ARTIFACT_PREFIX=conda_envs
         fi
--- a/conda_smithy/templates/azure-pipelines-osx.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-osx.yml.tmpl
@@ -69,6 +69,7 @@ jobs:
         export BLD_ARTIFACT_PREFIX=conda_pkgs_failed
         # Archive the CONDA_BLD_PATH environments only when the job fails
         export ENV_ARTIFACT_PREFIX=conda_envs
+        export WRK_ARTIFACT_PREFIX=conda_work
       fi
       ./.scripts/create_conda_build_artifacts.sh
     displayName: Prepare conda build artifacts
@@ -80,6 +81,13 @@ jobs:
     inputs:
       targetPath: $(BLD_ARTIFACT_PATH)
       artifactName: $(BLD_ARTIFACT_NAME)
+
+  - task: PublishPipelineArtifact@1
+    displayName: Store conda work directoryartifacts
+    condition: not(eq(variables.WRK_ARTIFACT_PATH, ''))
+    inputs:
+      targetPath: $(WRK_ARTIFACT_PATH)
+      artifactName: $(WRK_ARTIFACT_NAME)
 
   - task: PublishPipelineArtifact@1
     displayName: Store conda build environment artifacts
--- a/conda_smithy/templates/azure-pipelines-osx.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-osx.yml.tmpl
@@ -58,13 +58,15 @@ jobs:
   - script: |
       export ARTIFACT_STAGING_DIR="$(Build.ArtifactStagingDirectory)"
       # Archive everything in CONDA_BLD_PATH except environments
-      export BLD_ARTIFACT_PREFIX=conda_artifacts
+      export BLD_ARTIFACT_PREFIX=conda_pkgs
       export CI=azure
       export CI_RUN_ID=$(build.BuildNumber).$(system.JobAttempt)
       export CONDA_BLD_PATH=$(build_workspace_dir)
       export FEEDSTOCK_NAME=$(basename ${BUILD_REPOSITORY_NAME})
       export MINIFORGE_HOME=$(tools_install_dir)
       if [[ "$AGENT_JOBSTATUS" == "Failed" ]]; then
+        # Use a distinct prefix on failure (to match GHA).
+        export BLD_ARTIFACT_PREFIX=conda_pkgs_failed
         # Archive the CONDA_BLD_PATH environments only when the job fails
         export ENV_ARTIFACT_PREFIX=conda_envs
       fi
--- a/conda_smithy/templates/azure-pipelines-win.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-win.yml.tmpl
@@ -101,7 +101,7 @@ jobs:
 
     - script: |
         set ARTIFACT_STAGING_DIR=$(Build.ArtifactStagingDirectory)
-        set BLD_ARTIFACT_PREFIX=conda_artifacts
+        set BLD_ARTIFACT_PREFIX=conda_pkgs
         set CI=azure
         set CI_RUN_ID=$(build.BuildNumber).$(system.JobAttempt)
         set CONDA_BLD_PATH=$(build_workspace_dir)
@@ -110,6 +110,7 @@ jobs:
         set FEEDSTOCK_NAME=%FEEDSTOCK_NAME:*/=%
         set MINIFORGE_HOME=$(tools_install_dir)
         if "%AGENT_JOBSTATUS%" == "Failed" (
+            set BLD_ARTIFACT_PREFIX=conda_pkgs_failed
             set ENV_ARTIFACT_PREFIX=conda_envs
         )
         call ".scripts\create_conda_build_artifacts.bat"
--- a/conda_smithy/templates/azure-pipelines-win.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-win.yml.tmpl
@@ -106,6 +106,8 @@ jobs:
         set CI_RUN_ID=$(build.BuildNumber).$(system.JobAttempt)
         set CONDA_BLD_PATH=$(build_workspace_dir)
         set FEEDSTOCK_NAME=$(build.Repository.Name)
+        rem strip '{org}/' from `{org}/{repo}` to get feedstock name
+        set FEEDSTOCK_NAME=%FEEDSTOCK_NAME:*/=%
         set MINIFORGE_HOME=$(tools_install_dir)
         if "%AGENT_JOBSTATUS%" == "Failed" (
             set ENV_ARTIFACT_PREFIX=conda_envs
--- a/conda_smithy/templates/azure-pipelines-win.yml.tmpl
+++ b/conda_smithy/templates/azure-pipelines-win.yml.tmpl
@@ -112,6 +112,7 @@ jobs:
         if "%AGENT_JOBSTATUS%" == "Failed" (
             set BLD_ARTIFACT_PREFIX=conda_pkgs_failed
             set ENV_ARTIFACT_PREFIX=conda_envs
+            set WRK_ARTIFACT_PREFIX=conda_work
         )
         call ".scripts\create_conda_build_artifacts.bat"
       displayName: Prepare conda build artifacts
@@ -123,6 +124,13 @@ jobs:
       inputs:
         targetPath: $(BLD_ARTIFACT_PATH)
         artifactName: $(BLD_ARTIFACT_NAME)
+
+    - task: PublishPipelineArtifact@1
+      displayName: Store conda work directoryartifacts
+      condition: not(eq(variables.WRK_ARTIFACT_PATH, ''))
+      inputs:
+        targetPath: $(WRK_ARTIFACT_PATH)
+        artifactName: $(WRK_ARTIFACT_NAME)
 
     - task: PublishPipelineArtifact@1
       displayName: Store conda build environment artifacts
--- a/conda_smithy/templates/create_conda_build_artifacts.bat.tmpl
+++ b/conda_smithy/templates/create_conda_build_artifacts.bat.tmpl
@@ -12,6 +12,7 @@ rem Optional:
 rem ARTIFACT_STAGING_DIR (use working directory if unset)
 rem BLD_ARTIFACT_PREFIX (prefix for the conda build artifact name, skip if unset)
 rem ENV_ARTIFACT_PREFIX (prefix for the conda build environments artifact name, skip if unset)
+rem WRK_ARTIFACT_PREFIX (prefix for the conda work directory artifact name, skip if unset)
 
 rem OUTPUTS
 rem
@@ -19,6 +20,8 @@ rem BLD_ARTIFACT_NAME
 rem BLD_ARTIFACT_PATH
 rem ENV_ARTIFACT_NAME
 rem ENV_ARTIFACT_PATH
+rem WRK_ARTIFACT_NAME
+rem WRK_ARTIFACT_PATH
 
 {%- if conda_install_tool == "pixi" %}
 set "PATH=%USERPROFILE%\.pixi\bin;%PATH%"
@@ -59,8 +62,10 @@ cd "%CONDA_BLD_PATH%"
 if errorlevel 1 exit 1
 
 set ENVIRONMENT_PATHS=
+set WORK_PATHS=
 set "EXCLUDE_COMMON=--exclude=.git --exclude=./pkg_cache --exclude=./src_cache"
 set EXCLUDE_FROM_BUILD_ARTIFACTS=
+set EXCLUDE_FROM_WORK=
 
 for /d %%a in (. bld test) do (
     if exist %%a (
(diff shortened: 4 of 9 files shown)


h-vetinari:
What's the status here? 🙃

mgorny:
I want to finish the Linux part today and start with Windows. Ofc early feedback appreciated too, in case I'm going in the wrong direction.

mgorny:
Okay, this is a good point for feedback before I start mirroring the changes to Windows. https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/1790 demonstrates the changes: we're now getting 3 `.tar.zst` archives:

1. built artifacts (implemented as "everything except build directories [and environments]")
2. work directories (implemented as "everything in build directories except for environments")
3. environments

I suppose we may want to adjust the compression level for individual artifacts still. Oh, and probably enable multithreaded compression if it's not enabled by default.

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -294,6 +295,7 @@ jobs:
         # so random failures don't prevent rebuilds from creating artifacts.
         if [ $JOB_STATUS == "failure" ]; then
           export BLD_ARTIFACT_PREFIX="conda_artifacts"
+          export WRK_ARTIFACT_PREFIX="conda_work"
           export ENV_ARTIFACT_PREFIX="conda_envs"
```
h-vetinari, line 303: please alphabetize env vars

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -107,6 +107,7 @@ jobs:
         set FEEDSTOCK_NAME=$(build.Repository.Name)
         set MINIFORGE_HOME=$(tools_install_dir)
         if "%AGENT_JOBSTATUS%" == "Failed" (
+            set WRK_ARTIFACT_PREFIX=conda_work
             set ENV_ARTIFACT_PREFIX=conda_envs
```
h-vetinari: same

conda_smithy/templates/azure-pipelines-osx.yml.tmpl
```diff
@@ -66,6 +66,7 @@ jobs:
       export MINIFORGE_HOME=$(tools_install_dir)
       if [[ "$AGENT_JOBSTATUS" == "Failed" ]]; then
         # Archive the CONDA_BLD_PATH environments only when the job fails
+        export WRK_ARTIFACT_PREFIX=conda_work
         export ENV_ARTIFACT_PREFIX=conda_envs
```
h-vetinari: idem

conda_smithy/templates/azure-pipelines-linux.yml.tmpl
```diff
@@ -81,6 +81,7 @@ jobs:
         export MINIFORGE_HOME=$(tools_install_dir)
         if [[ "$AGENT_JOBSTATUS" == "Failed" ]]; then
           # Archive the CONDA_BLD_PATH environments only when the job fails
+          export WRK_ARTIFACT_PREFIX=conda_work
           export ENV_ARTIFACT_PREFIX=conda_envs
```
h-vetinari: idem

h-vetinari (commented):
Basically LGTM aside from some very minor nit.

W.r.t. improvements, I wish we could suppress the logs about the artefact creation some more; for big packages it's just an endless file listing that blows through the logs and makes it unnecessarily hard to scroll up to the build failure.

mgorny:
> W.r.t. improvements, I wish we could suppress the logs about the artefact creation some more; for big packages it's just an endless file listing that blows through the logs and makes it unnecessarily hard to scroll up to the build failure.

Sure, I was wondering about that too. Added `-v` literally because the old calls used to output file lists (and it made checking the results easier).

conda_smithy/templates/create_conda_build_artifacts.sh.tmpl
```diff
@@ -26,6 +27,9 @@ source .scripts/logging_utils.sh
 # and that might end up inserting extraneous quotation marks in output variables
 set -e
 
+# mangle_homebrew hides zstd on GHA macos 15 runners, so use conda-forge tools
+export PATH=${MINIFORGE_HOME}/bin:${PATH}
```
jaimergp:
```suggestion
export PATH="${MINIFORGE_HOME}/bin:${PATH}"
```

jaimergp: Or... shall we instead activate the base environment?

```suggestion
. "${MINIFORGE_HOME}/etc/profile.d/conda.sh
conda activate
```

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -294,6 +295,7 @@ jobs:
...
         else
           export BLD_ARTIFACT_PREFIX="conda_pkgs"
```
jaimergp, line 306: I know this was pre-existing, but why have `BLD_ARTIFACT_PREFIX` different in success and failure cases? (`conda_pkgs`, `conda_artifacts`, respectively).

I'm not against distinguishing them, but they should have better names in that case:

- Successful build: `conda_pkgs_ok`
- Failed build: `conda_pkgs_failed`

conda_smithy/templates/create_conda_build_artifacts.sh.tmpl
```diff
@@ -57,21 +61,59 @@ echo "ARTIFACT_UNIQUE_ID: $ARTIFACT_UNIQUE_ID"
...
+    ./*_*/pip_cache
+)
+EXCLUDE_FROM_BUILD_ARTIFACTS=(
+    "${EXCLUDE_COMMON[@]}"
+    "${BUILD_PATHS[@]}"
+)
+EXCLUDE_FROM_WORK=(
+    "${EXCLUDE_COMMON[@]}"
+    "${ENVIRONMENT_PATHS[@]}"
+)
```
jaimergp, line 119: This is going to be fun in Windows 😂 What's the plan there? We are already relying on an activated conda environment, so at that point we can go full Python, I guess?

conda_smithy/templates/github-actions.yml.tmpl
mgorny: I suppose the difference was that in the success case they were just packages, while in the failure case it also involved the work directory. But that's no longer true, so it makes sense to make them consistent.

conda_smithy/templates/create_conda_build_artifacts.sh.tmpl
mgorny: I'm going to try with CMD first. It might not be that hard with a bit of `FOR` (and then I'm going to discover delayed evaluation stuff applies and regret saying that). Well, we're going to see shortly.

conda_smithy/templates/create_conda_build_artifacts.sh.tmpl
mgorny: I'm going to try if activating it works as expected.

h-vetinari:
I merged main here and used the result to rerender a v1 & a v0 recipe; unfortunately, the v1 recipe just fails outright (that's not a fault of this PR though), and the v0 recipe does not seem to generate artefacts at all. Could you check what's going on with the v0 stuff please? We'll have a fix for the v1 stuff soon I guess, because it affects all renders (effectively making the main branch here unreleasable).

h-vetinari:
OK, nevermind, I had committed the capital error of typing a `workflow_setting` keys from memory, and got that wrong (typed `store_artifact` instead of `store_build_artifacts`); unfortunately the linter doesn't warn on this, and just happily populates the CI files with this value. I think we should switch the permitted `workflow_settings` keys to an enum that's validated in the linter...

mgorny:
Does that mean that v0 worked for you now as well?

h-vetinari:
Currently running in https://github.com/conda-forge/pyarrow-feedstock/pull/182, but looking good

h-vetinari:
Interestingly, for 1 out of 6 (intentionally) failing jobs on windows, artefact creation on [azure](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1561980&view=results) fails with:
```
bsdtar: no files or directories specified
```
Those jobs are identical up to the python version, so something is either flaky, or there are different image versions flying around that may respond differently

h-vetinari:
From a passing linux [job](https://github.com/conda-forge/pyarrow-feedstock/actions/runs/30905416110/job/91979307386?pr=182) on GHA: 
```
tar: Cowardly refusing to create an empty archive
Try 'tar --help' or 'tar --usage' for more information.
Error: Process completed with exit code 2.
```

mgorny:
Good catch. I need to think how to handle the potential case that we have no files to archive. My first thought would be that we don't want to publish empty archives *but* that could be confusing ("why is there no work artifact here?").

mgorny:
Hmm, maybe create an empty archive and add an `-empty` suffix to make it clear it's not worth downloading. Or just upload a dummy file with `-empty` marker.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: Oh, sorry, I've just noticed the comment above. Lemme restore a distinction, per your suggestion above.

mgorny:
I'm finding so many bugs in my code that I'm ashamed of myself. I'll have an update in ~30 minutes or so.

mgorny:
Okay. Rebased it, fixed all the issues reported and a few unreported. Also improved the logic for detecting whether artifacts were created, and added support for uploading `-broken` artifacts when `tar` fails (e.g. some files can't be read).

mgorny:
One more thing I could do is replace the three repeated tarring blocks with an abstract function that's called thrice. However, I'm not sure if it's worth it, as it will be less readable.

h-vetinari:
I retried this in https://github.com/conda-forge/pyarrow-feedstock/pull/182 and it looks much better, thanks. I went through the commits one by one, and I have to say there's too much churn here to (non-squash-)merge this as-is. Stuff like things being removed and then being reinstated, being introduced and then reordered, being introduced and then rewritten a couple of times.

Given how much this PR does, I'm not advocating to squash this (though I could live with that), but IMO this could be condensed to 5-8 commits or so.

mgorny:
I can rewrite the history and make it into clean, atomic changes. However, first I'd like to know what you (plural) think of deduplicating the logic for three archives into a single helper function (bash) and/or Jinja for loop (cmd). If that sounds like a good idea, I'd probably do that as the first commit, so there's much less churn afterwards.

h-vetinari:
> first I'd like to know what you (plural) think of deduplicating the logic for three archives into a single helper function (bash) and/or Jinja for loop (cmd).

completely open to this; especially if it leads to DRYer code and something that's more easily maintainable (the latter is somewhat subjective of course; I'll almost certainly be happy to follow your lead on this, if you think the mentioned refactor is an improvement)

h-vetinari:
I reran https://github.com/conda-forge/numpy-feedstock/pull/392 with a rerender based on this PR (together with #2644), and it looks like the artefact generation for osx on azure fails with:
```
./.scripts/create_conda_build_artifacts.sh: line 34: /Users/runner/miniforge3/etc/profile.d/conda.sh: No such file or directory
```

mgorny:
What you're saying is that I need to separately account for pixi, right?

h-vetinari:
Also, on failing linux builds ([CI run](https://github.com/conda-forge/numpy-feedstock/actions/runs/30990579414)), the "Store conda build environment artefacts" section was skipped completely; it's hard to tell from the logs why that happened. I think we want to see more than
```
ARTIFACT_STAGING_DIR: /home/runner/work/numpy-feedstock/numpy-feedstock
ARTIFACT_UNIQUE_ID: 30990579414_linux_64_is_freethreadingfalsepython3.14.____cp314
~/work/numpy-feedstock/numpy-feedstock/build_artifacts ~/work/numpy-feedstock/numpy-feedstock
BLD_ARTIFACT_NAME: conda_pkgs_failed_30990579414_linux_64_is_freethreadingfalsepython3.14.____cp314
BLD_ARTIFACT_PATH: /home/runner/work/numpy-feedstock/numpy-feedstock/numpy-feedstock_conda_pkgs_failed_30990579414_linux_64_is_freethreadingfalsepython3.14.____cp314.tar.zst
WRK_ARTIFACT_NAME: conda_work_30990579414_linux_64_is_freethreadingfalsepython3.14.____cp314
WRK_ARTIFACT_PATH: /home/runner/work/numpy-feedstock/numpy-feedstock/numpy-feedstock_conda_work_30990579414_linux_64_is_freethreadingfalsepython3.14.____cp314.tar.zst
~/work/numpy-feedstock/numpy-feedstock
```
under "Prepare conda build artefacts"; e.g. `created <foo>.tar.st successfully, has size <bar>`

mgorny:
Okay, starting the rewrite. Will overwrite this PR once it's ready.

mgorny:
Rewrite done. It's not 100% perfect: I've left format + patterns split on Unix, but combined them on Windows, because reimplementing the old patterns would be too much effort. Other than that, I've tried to keep it split into atomic changes. The diff between the two shows only cosmetic changes.

I didn't fix any of the remaining issues yet, but I'll be doing that as fixups, to preserve clean history.

mgorny:
@h-vetinari, how detailed should the success/"failure" messages be? The potential scenarios are:

- `tar` succeeded and we got a "good" archive
- `tar` failed and we got a "broken" archive that we're uploading (distinguishable by filename, also you will see some `tar` errors)
- `tar` failed and did not produce an archive (we're skipping the upload, also you will see some `tar` errors in "prepare" step)
- archiving skipped because there were no files to archive
- archiving skipped because prefix var was empty (the usual skip based on whether build succeeded or failed)

mgorny:
pixi should be fixed now.

mgorny:
Just noticed that I'm not handling the case of "tar failed and did not create a file at all" correctly, as it will still trigger (failing) upload.

jaimergp:
Heads up that some users are reporting the old artifacts scripts being marked as non-executable, so let's double check that regression is fixed in this PR: [#general > create_conda_build_artifacts.sh: Permission denie @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/create_conda_build_artifacts.2Esh.3A.20Permission.20denie/near/615214822)

conda_smithy/templates/create_conda_build_artifacts.bat.tmpl
jaimergp, line 1: Can we add `@ECHO OFF` at the beginning? Right now these are unnecessarily [verbose](https://github.com/conda-forge/numpy-feedstock/actions/runs/31189143405/job/92901208402?pr=392#step:10:34) (and we are running the Unix equivalents without `-x`) so I assume we didn't choose that explicitly.

jaimergp:
Also, can we document the different outcomes of artifacts in the [schema docstring](https://github.com/mgorny/conda-smithy/blob/ca1f79772917707a2b4b32bf5eafac04aff0f469/conda_smithy/schema.py#L552-L554)? Things like:

- Type of compression
- Which type gets uploaded on success vs failure
- Contents of each artifact type

Ideally this is paired with a How-to that explains how to download them and install locally for debugging.

jaimergp:
From what I can gather, there are three types of artifacts:

- Built packages: available on success and failed tests
- Working directory: available on any failure
- Build environments: available on failed build scripts

Observe these three runs:

- [Failed test](https://github.com/conda-forge/numpy-feedstock/actions/runs/31190204362/job/92904897040?pr=392): has built packages and working directory ✅ 
- [Failed script](https://github.com/conda-forge/numpy-feedstock/actions/runs/31190204362/job/92904897020?pr=392): has all three artifacts, but AFAIK a .conda artifact can't possibly exist, so what did we add in the built packages? 🤔 
- [Successful](https://github.com/conda-forge/numpy-feedstock/actions/runs/31190204362/job/92904896900?pr=392): has built packages only ✅

conda_smithy/templates/create_conda_build_artifacts.bat.tmpl
mgorny: Sure. I was actually wondering about it, given we added some verbose echoes anyway.

mgorny:
> * has all three artifacts, but AFAIK a .conda artifact can't possibly exist, so what did we add in the built packages?

It's defined "as everything except for work trees and environments", which usually means some "odd files" like rattler-build log and repository indexes. I was wondering whether we should try to detect that and skip it, but figured out it's too much effort for a corner case, and too much risk we'd mistakenly miss something.

mgorny:
Perhaps a cleaner approach would be to have "build artifacts" archive contain only `.conda` package and related index files, and to treat "work directory artifacts" as everything not in "build artifacts" or "environment artifacts". But if so, I'd rather address that as a separate PR since this one's quite big already, and we'd need some fine-tuning to get it right.

mgorny:
Addressed the review comments, squashed the fixups and rebased against `main`.

h-vetinari:
> Heads up that some users are reporting the old artifacts scripts being marked as non-executable, so let's double check that regression is fixed in this PR: [#general > create_conda_build_artifacts.sh: Permission denie @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/create_conda_build_artifacts.2Esh.3A.20Permission.20denie/near/615214822)

Sorry about that. 😑 

I don't see this addressed in the diff yet? I can also try to file a separate fix

conda_smithy/templates/azure-pipelines-win.yml.tmpl
```diff
@@ -106,6 +106,7 @@ jobs:
         set CI_RUN_ID=$(build.BuildNumber).$(system.JobAttempt)
         set CONDA_BLD_PATH=$(build_workspace_dir)
         set FEEDSTOCK_NAME=$(build.Repository.Name)
+        set FEEDSTOCK_NAME=%FEEDSTOCK_NAME:*/=%
```
h-vetinari, line 110: how do asterisks end up in `%FEEDSTOCK_NAME%` in the first place? Can we add a comment here about removing them?

h-vetinari (approved):
Basically LGTM, thanks a lot. If we plan on further distinguishing artefacts in terms of success & failure (something I'd consider worthwhile), then we should do that in the same release as this PR.

conda_smithy/templates/azure-pipelines-win.yml.tmpl
mgorny: It's a wildcard. It changes `conda-forge/foo-feedstock` into `foo-feedstock`.

mgorny:
> > Heads up that some users are reporting the old artifacts scripts being marked as non-executable, so let's double check that regression is fixed in this PR: [#general > create_conda_build_artifacts.sh: Permission denie @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/create_conda_build_artifacts.2Esh.3A.20Permission.20denie/near/615214822)
> 
> Sorry about that. 😑
> 
> I don't see this addressed in the diff yet? I can also try to file a separate fix

Sorry, I thought you meant that this PR may be causing a regression. I couldn't reproduce it nor see why it would happen here.

h-vetinari:
I've fixed it in #2647. I'd prefer making a "patch" release with that before we integrate this PR (and finalize the redesign)

conda_smithy/templates/azure-pipelines-win.yml.tmpl
h-vetinari: I think that deserves a comment. Something like
```
strip "conda-forge/" organization from `<org>/<repo>` to get feedstock name
```

mgorny: Done, in more abstract terms.

h-vetinari:
> If we plan on further distinguishing artefacts in terms of success & failure (something I'd consider worthwhile), then we should do that in the same release as this PR.

Could you prioritise this? I wouldn't like to leave smithy in a quote-unquote unreleasable state[^1]. I agree that this PR is big enough though, so good to go from my side.

[^1]: obviously that's hyperbole; we could release without further work, I'd just prefer not to spread out closely related changes if possible

h-vetinari (approved):

mgorny:
> If we plan on further distinguishing artefacts in terms of success & failure (something I'd consider worthwhile), then we should do that in the same release as this PR.

Is this about the idea of moving files around per https://github.com/conda-forge/conda-smithy/pull/2631#issuecomment-5219347929 or something else?

h-vetinari:
Exactly that one

mgorny:
Starting on that one immediately.

mgorny:
https://github.com/conda-forge/conda-smithy/pull/2649

PR #2632, opened 2026-07-25
Use `GITHUB_REF_NAME` simpler for upload checks

jakirkham:
As `GITHUB_REF_NAME` is the branch name when a `push` event occurs, there is no need to strip unrelated content from it to get the branch name. Instead it can be used as is to populate `GIT_BRANCH`.

Also when a pull request is involved, it is `PR#/merge`, which is not the upstream branch name. So it will also be correctly skipped when a PR is involved.

This also ensures that branch names with other characters like `/` are handled correctly, which `basename` would have missed.

<hr>

Checklist

### DIFF
--- a/conda_smithy/templates/github-actions.yml.tmpl
+++ b/conda_smithy/templates/github-actions.yml.tmpl
@@ -142,7 +142,7 @@ jobs:
       run: |
         export CONDA_BLD_PATH="${CONDA_BLD_PATH/#~/${HOME}}"
         export FEEDSTOCK_NAME="$(basename $GITHUB_REPOSITORY)"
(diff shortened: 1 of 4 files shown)


jakirkham:
Tested this in an external org PR: https://github.com/rapidsai/xgboost-feedstock/pull/145

Though it would work the same in conda-forge

h-vetinari (approved):
Simpler is better here. 👍

jaimergp (approved):

jaimergp:
Thanks!

jakirkham:
Thank you both! 🙏

Issue #2627, opened 2026-07-23
Remove all code/config for obsolete ci providers


h-vetinari:
These are all dead as a doornail; we should lighten the load of all the cruft (and tests for that) we're carrying around. Not sure how we want to play this w.r.t. deprecation. @jaimergp @beckermr 

Services to remove: appveyor, circleci, drone, travis, woodpecker 

* [X] Deprecated in conda-smithy as of 26.8
* [ ] Removed from conda-smithy as of 26.10, maybe 26.11 to leave some space for the depending services:
  * [ ] conda-forge-bot
  * [ ] admin-requests and feedstock creation
  * [ ] conda-forge-webservices
  * [ ] admin-migrations
  * [ ] conda-forge-feedstock-ops

<details>

C.f.
https://github.com/conda-forge/conda-smithy/blob/60bcbadff0143da134fd5564b87bbbc142bbd546/conda_smithy/schema.py#L55-L62

also (mainly) `_<foo>_specific_setup` functions in https://github.com/conda-forge/conda-smithy/blob/main/conda_smithy/configure_feedstock.py

</details>

beckermr:
We have a deprecation policy plus calver, so I think we can follow that. 

Is there something more specific you are asking about?

h-vetinari:
> Is there something more specific you are asking about?

I'm aware of the policy. My point is that it's not just a simple function to deprecate. In other words, it's not clear how we can even conceivably warn correctly in all cases that would theoretically be impacted. There are no recipes (except perhaps some long-abandoned or archived ones) that are using these CI providers. I strongly doubt that the CI setup here is still functional, much less that third-party users of conda-smithy still use them.

Given the effort to deprecate this "correctly", vs. the fact that these have all been dead for years, I consider deprecation to be a waste of time in this case. I'll go out on a limb and assume you'll disagree, but then I need inputs on _how_ you think this should be deprecated.

h-vetinari:
I guess if you count a runtime warning during rerendering a recipe with `provider: <plat>: <foo>` as a deprecation warning, then this could work (though that's not really aimed at users of smithy's API, only indirectly by them running their code and inspecting the warnings).

beckermr:
Folks use smithy outside of conda-forge, and while some of those providers are no longer around, others still work (but we do not use them).

So I assumed we should follow the policy.

OTOH, if you want to simply disregard it, then I guess go ahead, but maybe consider also removing the policy? 

[...]

PR #2633, opened 2026-07-27
deprecate obsolete CI providers

h-vetinari:
Towards #2627


h-vetinari:
pre-commit.ci autofix

h-vetinari:
A few comments:
* I've tried to be exhaustive, but haven't gone to extreme length for it (basically, I was `git grep`ing to see were given provider names showed up)
* The existing deprecation mechanism doesn't work for runtime warnings AFAICT (since we can't warn on use of config values, the discussion in #2627 was to at least provide some notice to potential users), so I'm now raising those warnings by hand.
* The current `@deprecated()` mechanism uses a `PendingDeprecationWarning` by default, which is not very noisy and more importantly, does not fail the test suite. AFAICT, after the release of 2026.7, we'll be getting `DeprecationWarning`s instead, which will probably instantly turn our CI red (until we add something like `@pytest.mark.filterwarnings("ignore::DeprecationWarning")` to affected tests). I don't really see the point of ever issuing `PendingDeprecationWarning` (even though I get that this code is copied or at least inspired by CPython). I'd just skip that and go for `DeprecationWarning` directly.

conda_smithy/cli.py
```diff
@@ -377,7 +379,9 @@ def __call__(self, args):
                 "`upload_packages: False` per provider field in"
                 "conda-forge.yml to disable package uploads."
             )
+        depr_legacy = "Provider '{provider}' is deprecated and will be removed in 2026.10, see #2627"
         if args.travis:
+            warnings.warn(depr_legacy.format(provider="Travis"), DeprecatedError)
```
jaimergp: You can use `deprecated.topic()` for these cases, no need to do manual `warnings.warn` invocations.

news/deprecate-obsolete-ci-providers.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* All functions and functionality regarding obsolete CI providers (appveyor, circleci, cirrus, drone, travis, woodpecker) have been deprecated.
```
jaimergp:
```suggestion
* CI integrations for appveyor, circleci, cirrus, drone, travis, woodpecker are now considered deprecated. All functions and functionality regarding these  obsolete CI providers will be removed in 26.10. (#2633)
```

conda_smithy/anaconda_token_rotation.py
```diff
@@ -195,6 +196,11 @@ def rotate_anaconda_token(
             )
 
 
+@deprecated(
+    "2026.7",
```
jaimergp: Not sure if we'll make it in July so maybe we need August here. TBD.

jaimergp:
The CIServices enum can be modified to have properties for the deprecated fields, where we emit `deprecated.topic()` warnings.

h-vetinari:
pre-commit.ci autofix

jaimergp (approved):
This is an excellent start I'd say. We'll see what happens with the CI and the warnings but for now I'm inclined to merge and refine later.

h-vetinari:
Thanks a lot for the help and the review! 🤩

PR #2526, opened 2026-04-13
[PyCon DE Sprints] Migrate remaining tests from unittest to pytest

NMertsch:
This PR migrates the remaining unittest-based tests to pytest, without modifying anything else.

Steps:
* Remove unused `setUp` and `tearDown` methods in `test_feedstock_io.py`
    * `setUp` creates a temporary directory and `cd`'s into it
    * `tearDown` `cd`'s back into the original directory and deletes the temporary directory
    * Each test does the same using the `parametrize` method
* Replace `self.assertEquals(expr1, expr2)` with `assert expr1 == expr2`, and similar for `assertIn`, `assertTrue`, etc.
* Remove `class X(unittest):`, unindent all its methods by 4 spaces
    * If present, apply class-level decorator to each method (`@pytest.mark.cli` in `test_lint_recipe.py`)

Suggestions for future PRs:
* Use pytest's built-in fixtures instead of hand-rolling their functionality, e.g. `monkeypatch.chdir`, `tmp_dir`.
* Activate ruff's pytest ruleset to enforce a more unified pytest usage.

Checklist
* [x] Added a ``news`` entry
* [x] Regenerated schema JSON if schema altered (`python -m conda_smithy.schema`)

PR #2639, opened 2026-07-30
Allow noarch to be null or jinja conditional

### LINKED ISSUE #2425: False positive: Invalid noarch value ${{ "python" if use_noarch }} in v1 recipe

### Issue

I used `noarch: ${{ "python" if use_noarch }}` in https://github.com/conda-forge/coverage-feedstock/pull/154

Allowing empty value for `noarch` is a feature that was added in rattler-build 0.50.0: https://github.com/prefix-dev/rattler-build/pull/1973

The linter currently complains with: `Invalid noarch value ${{ "python" if use_noarch }}. Should be one of python, generic.`. It should allow conditional with empty values.

### Installed packages

```shell
NA
```

### Environment info

```shell
NA
```

beenje:
Checklist
* [X] Added a ``news`` entry with any new deprecations added to the ``Deprecated`` section.
* [X] Regenerated linter documentation if any rules were added, removed or modified (`python -m conda_smithy.linter.messages`)

Fix #2425

Since rattler-build [0.69.0](https://github.com/prefix-dev/rattler-build/blob/main/CHANGELOG.md#0690---2026-07-15), conditional can be used for `noarch`.  Empty/null-like noarch values are treated as absent.

See https://github.com/prefix-dev/rattler-build/pull/2626

### DIFF
--- a/conda_smithy/linter/messages/recipe.py
+++ b/conda_smithy/linter/messages/recipe.py
@@ -369,13 +369,18 @@ class SourceHash(LinterMessage, _AnyRecipeMessage):
 class NoarchValue(LinterMessage, _AnyRecipeMessage):
     """
     The `build.noarch` field can only take `python` or `generic` as a value.
-    Recipe v1 also accepts `null`, `none`, `None` or `~` to leave `noarch` unset.
+    Recipe v1 also accepts a jinja expression that renders to an empty value, none, "null" or "~" to leave `noarch` unset."
     """
 
     kind = "lint"
     identifier = "R-020"
     valid: ClassVar[list[str]] = ["python", "generic"]
-    valid_v1: ClassVar[list[str]] = [*valid, "null", "none", "None", "~"]
+    # Note that null and none literals are accepted by rattler-build, but not by the linter
+    # A jinja expression is required outside python and generic
+    valid_v1: ClassVar[list[str]] = [
+        *valid,
+        "a jinja expression that renders to an empty value, none, 'null' or '~'",
+    ]
     message = "Invalid `noarch` value `${given}`. Should be one of `${valid}`."
     given: str
     recipe_version: RECIPE_VERSIONS
@@ -395,7 +400,7 @@ class NoarchValue(LinterMessage, _AnyRecipeMessage):
     ) -> object | None:
         """
         Render the value when a jinja expression is given in v1 recipe
-        Return None if the expression can't be rendered - check ignored
+        Return the given input otherwise
         """
         from conda_smithy.linter import conda_recipe_v1_linter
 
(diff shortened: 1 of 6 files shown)


wolfv (approved):

tests/test_lint_recipe.py
```diff
@@ -651,6 +651,38 @@ def test_noarch_value(self):
         lints, hints = linter.lintify_meta_yaml(meta)
         self.assertTrue(any(lint.startswith(expected) for lint in lints))
 
+    def test_recipe_v1_noarch_value_allows_conditional_no_context(self):
+        meta = {"build": {"noarch": '${{ "python" if use_noarch }}'}}
+        expected = "Invalid `noarch` value"
+        lints, hints = linter.lintify_meta_yaml(meta, recipe_version=1)
+        self.assertFalse(any(lint.startswith(expected) for lint in lints))
```
h-vetinari: The naming of the variable `expected` is at least confusing, or arguably simply wrong. No warning is expected, so doing the equivalent of `assert not any(<"expected" happens>)` is needlessly cruel to readers of the code.

conda_smithy/linter/messages/recipe.py
```diff
@@ -368,16 +369,81 @@ class SourceHash(LinterMessage, _AnyRecipeMessage):
...
+                # No context, can't render the jinja expression - skip
+                return None
+            given = render_recipe_with_context(meta).get("build", {}).get("noarch")
+            if given is None or (
+                isinstance(given, str)
+                and conda_recipe_v1_linter.JINJA_VAR_PAT.search(given)
+            ):
+                # expression rendered to unset/null
+                # or expression could not be resolved from known context
+                return None
```
h-vetinari: I was looking at the tests for guidance on what this actually means; I'm not hugely enamoured with the idea of mixing legitimate expressions of `noarch: <none>` with various errors to parse or render -- or at least, that what this looks like (though the tests correctly reject `noarch: banana`).

I don't claim to know all the context here, but absent further explanation/tests, I'd much prefer if `return None` only happened on the `set(cls.valid_v1) - set(cls.valid)` values

h-vetinari (dismissed):
A question and a naming nit

tests/test_lint_recipe.py
beenje: That's fair :-) I renamed the variable to `unexpected_lint`

conda_smithy/linter/messages/recipe.py
beenje: The `rendered_value` method will only do something for v1 recipe (`recipe_version == 1`) and if the `given` noarch value is a Jinja expression (`conda_recipe_v1_linter.JINJA_VAR_PAT.search(given)`).

For all other cases, it will return the same `given` value. Meaning for v0 or (v1 and standard strings), it will only compare to `valid_values`.

I added the test `test_noarch_value_recipe_v0_rejects_conditional`.

tests/test_lint_recipe.py
```diff
@@ -651,6 +651,38 @@ def test_noarch_value(self):
...
+
+    def test_recipe_v1_noarch_value_allows_rendered_conditional(self):
+        expected = "Invalid `noarch` value"
+        for use_noarch in (True, False):
```
beenje: I'd much prefer use `@pytest.mark.parametrize` than a for loop, but followed the existing tests here.

Is there a policy to convert tests to pytest?

h-vetinari: I think parametrization works even witin a class? You should definitely use it if you can. In the past, I've taken such cases as opportunities to modernize the tests in question (though in a separate commit)

conda_smithy/linter/messages/recipe.py
h-vetinari: OK. After a second look, what I think I stumbled over is that there's a distinction of what rattler-build allows as expressions for `noarch: <none>`, and the internal state of getting (non-string) `None` while processing that field. What I meant is that I don't want to mix those cases semantically, which is a challenge because they use the same nomenclature for (apparently?) different scenarios.

tests/test_lint_recipe.py
jaimergp: There's an open PR taking care of this conversion: https://github.com/conda-forge/conda-smithy/pull/2526

beenje: Parametrization doesn't work inside an `unittest.TestCase` class. I converted my tests to pytest.

conda_smithy/linter/messages/recipe.py
beenje: I updated the code to only allow jinja expressions that render to `"none"` or `None` (and of course `python` or `generic`). So null or none literals like `noarch: none` which are accepted by rattler-build aren't by the linter. I don't think it makes sense to have a recipe with such literals.

I first thought that if a variable was undefined, the expression wouldn't be rendered but that's not true. `render_recipe_with_context` will treat undefined variables as False due to `_MissingUndefined` used in jinja_env, which is fine.
- An empty string (no else), `"null"` or `"~"` render as `None`
- `none` and `None` render to the string `"None"`
- the quoted string `"none"` render to the string `"none"`

All those cases are treaded as missing field and this is what the linter is testing.

I think it makes more sense like that and hope it addresses your concern.

h-vetinari (approved):
LGTM, thanks

PR #2630, opened 2026-07-24
Add hint: remove obsolete cross-python ignore_run_exports abi3 workaround

killua156:
### Description

Adds linter hint **`R1-006`**, continuing the abi3 series after `R1-004`/`R1-005`.

abi3 (Python version-independent) recipes historically worked around a rattler-build limitation by manually ignoring the `python` run-export that `cross-python` injects when cross-compiling:

```yaml
requirements:
  ignore_run_exports:
    from_package:
      - cross-python_${{ target_platform }}
```

Without it, an abi3 package would pick up `cross-python`'s `python x.y.*` run-export and get pinned to a single Python version, defeating the purpose of abi3.

rattler-build now applies this run-export filtering automatically for abi3 recipes (prefix-dev/rattler-build#2252, extending the abi3 ignore to build-environment run exports), so the manual `ignore_run_exports` entry is redundant. `R1-006` detects it — at the top level or in any output, including `if/else`-wrapped entries — and hints that it should be removed.

An example of a recipe still carrying the workaround: [`fsspec-rs-feedstock` `recipe.yaml` L46-L48](https://github.com/conda-forge/fsspec-rs-feedstock/blob/main/recipe/recipe.yaml#L46-L48).

The hint is v1-only and skippable via `conda-forge.yml`'s `linter.skip: [hint_abi3_cross_python_run_exports]`.

### Checklist

- [x] Used a personal fork of the feedstock to propose changes
- [x] Ensured the license file is being packaged (n/a)
- [x] Added new hint `R1-006` with message, docs, regenerated `linter-messages.json`, news entry, and tests

Issue #2606, opened 2026-06-30
Lints for abi3 packages


isuruf:
We have an example for abi3 at https://github.com/conda-forge/python-abi3-feedstock/blob/main/recipe/example-recipe.yaml, but some packages are carrying around workarounds for bugs in conda-build and rattler-build around. We should provide some lints to remove those workarounds and also some lints for best practices.

- [x] remove python_min redefinitions that match the global pinning default - https://github.com/conda-forge/conda-smithy/pull/2581
- [x] python version independent recipes (abi3 for eg) should test for python latest too - https://github.com/conda-forge/conda-smithy/pull/2580
- [x] Remove rattler-build workaround for SP_DIR not being defined. - See for example https://github.com/conda-forge/fsspec-rs-feedstock/blob/main/recipe/recipe.yaml#L59-L63 - https://github.com/conda-forge/conda-smithy/pull/2619
- [x] Remove rattler-build workaround for cross-python run_exports for abi3 packages - See for example https://github.com/conda-forge/fsspec-rs-feedstock/blob/main/recipe/recipe.yaml#L46-L48 - github.com/conda-forge/conda-smithy/pull/2630
- [x] Check that abi3 packages use abi3audit. - github.com/conda-forge/conda-smithy/pull/2640

killua156:
Question on identifiers for the remaining abi3 lints (SP_DIR, cross-python run_exports, abi3audit): each needs a new message id, but #2580/#2581 already hold the next R1/R ids and the registry test forbids gaps.

Should I stack these PRs so the ids stay contiguous, or is there a preferred way to allocate ids across concurrent lint PRs?

isuruf:
> Should I stack these PRs so the ids stay contiguous, or is there a preferred way to allocate ids across concurrent lint PRs?

@jaimergp any thoughts on this?

jaimergp:
I'm afraid we will have to rebase as we merge them, the CI will test for duplicates and contiguity so as long as we run the tests before merging it should be caught.

isuruf:
Thanks @killua156 for sending the PRs

killua156:
thanks @isuruf for the opportunity to work on them.

PR #2640, opened 2026-07-31
Add hint: abi3 recipes should verify extension modules with abi3audit

killua156:
### Description

Last item of #2606: check that abi3 packages use `abi3audit`.

abi3 packages are built once against `python_min` but install on every later Python, so non-abi3 CPython API only fails at runtime. [`abi3audit`](https://github.com/pypa/abi3audit) catches it at build time, and the [example recipe](https://github.com/conda-forge/python-abi3-feedstock/blob/main/recipe/example-recipe.yaml) already runs it.

Adds `R1-008`, hinting v1 recipes that are Python version-independent but use `abi3audit` nowhere in their tests. It counts as used when listed in a test's `requirements.run` or invoked in a test `script`. `noarch: python` is skipped — no extension module to audit.

Follows #2630, so `R1-008` continues the series without a gap.

### Checklist

- [x] Used a personal fork of the feedstock to propose changes
- [x] Ensured the license file is being packaged (n/a)
- [x] Added the hint with message, docs, regenerated `linter-messages.json`, news entry, and tests


conda_smithy/lint_recipe.py
```diff
@@ -828,6 +829,16 @@ def run_conda_forge_specific(
             hints,
         )
 
+    # 10e: abi3 recipes should verify their extension modules with abi3audit
+    if "hint_abi3_missing_abi3audit" not in lints_to_skip and recipe_version == 1:
```
isuruf: Do the lint for recipe v0 too?
```suggestion
    if "hint_abi3_missing_abi3audit" not in lints_to_skip:
```

killua156:
Should I check the flags as well? The example recipe uses --assume-minimum-abi3 ${{ python_min }}. Right now I only check that abi3audit is used at all, so an invocation that leaves that flag out won't get flagged — I can add it if that's worth catching.

conda_smithy/lint_recipe.py
killua156: yes I was wondering if we should cover v0 recipes too. Will add it

conda_smithy/linter/hints.py
```diff
@@ -537,6 +537,77 @@ def hint_abi3_cross_python_run_exports(
...
+    for tests, build in scopes:
+        if not isinstance(build, Mapping):
+            continue
+        # `noarch: python` packages ship no compiled extension, so there is
+        # nothing for abi3audit to check
+        if build.get("noarch") == "python":
+            continue
+        if not get_version_independent(build, "python", recipe_version):
+            continue
+        if not _mentions_abi3audit(tests, recipe_version):
```
isuruf, line 627: Can you check that `python-abi3` is a host requirement? Otherwise it's not a abi3 recipe

killua156: Done now also requires `python-abi3` in host, flattening the `if: is_abi3` form so the example recipe matches. Added a case for version-independent recipes without it.

conda_smithy/lint_recipe.py
killua156: v0 is covered.

Issue #2445, opened 2026-01-04
Does not lint v1 recipes correctly; misspelled keys, undefined variables not diagnosed


h-vetinari:
In a recent PR I did
```yaml
      run_constraint:
        - if: win
          then:
            # avoid pulling in old MSVC activation
            - vs2019_win-64 <0.0a0
```
I had run
```
>conda smithy lint --conda-forge
E:\conda-forge\tinygrad-feedstock is in fine form
```
before pushing, but that missed the fact that this key is misspelled. Rerendering (and CI) showed the issue
```
Error:   x Failed to parse recipe

Error:
  x invalid field `run_constraint`.
    ,-[recipe\recipe.yaml:87:7]
 86 |                 - triton
 87 |       run_constraint:
    :       ^^^^^^^|^^^^^^
    :              `-- here
 88 |         - if: win
    `----
  help: valid options for requirements are build, host, run, run_constraints, run_exports, ignore_run_exports
```

But this should really be caught by the linter.

CC @wolfv @baszalmstra

h-vetinari:
Awesome! I had just seen https://github.com/prefix-dev/rattler-build/pull/2057 mentioned in the meeting notes and was wondering if that might help 🙃

PR #2650, opened 2026-08-11
Fix crash on unexpected v1 recipe sections and lint properly for them

mgorny:
Checklist

- fix linter crashing when unexpected top-level sections are found (when linting section order)
- update the unexpected section lint to distinguish multi-output / single-output v1 recipes (by the presence of `outputs` section) and lint appropriate
- add missing `tests` key to allowed multi-output v1 recipe keys

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -226,7 +226,7 @@
       "kind": "lint",
       "added_in": "<3.56",
       "deprecated_in": "",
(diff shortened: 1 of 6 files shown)


mgorny:
Just got a reply from Wolf, I've been reading CEP 41 and `requirements` key isn't valid then. Updated the commits not to add it, just `tests`.

h-vetinari:
LGTM; could you check if https://github.com/conda-forge/conda-smithy/issues/2445 is fixed, i.e. add a test if a mis-spelled key (like `run_constraint` instead of `run_constraints`) ends up causing a linter error?

mgorny:
> LGTM; could you check if #2445 is fixed, i.e. add a test if a mis-spelled key (like `run_constraint` instead of `run_constraints`) ends up causing a linter error?

It isn't. This check covers only top-level keys. I'd love to go beyond that but that's a bigger effort, and I guess it's better handled by something like YAML schema. I don't know if there are existing schemas for that (a quick search doesn't reveal anything), and if there aren't, I don't know if we really want to go down the rabbit hole of creating one, possibly via the pydantic abstraction.

h-vetinari (approved):
Thanks!

PR #2653, opened 2026-08-12
Update R1-0004/5 to allow for more flexible definition of "latest" python

### LINKED ISSUE #2645: Define "latest supported" Python for R1-004?
### Comment:

I've started seeing `R1-004` (added in #2580) in a bunch of feedstocks, and generally agree with the sentiment!

However, using  `python_version: *` ends up meaning "the latest _solveable_ python", not "the actual latest _supported by conda-forge_ python". Practically, the solve generally (blessedly) flies by so fast one is unlikely to note that `*` resolved to `actual_latest - 0.1`. The worst case is a non-`noarch` eventually needs an untested `noarch: python`, which can make it hard to find the root cause (e.g. a broken `win-64` upload somewhere).

To this end, I've been adding an explicit context value `python_check_max` in [~100 feedstocks](https://github.com/search?q=org%3Aconda-forge+python_check_max+path%3Arecipe.yaml&type=code), and now am down to [8 stragglers](https://github.com/search?q=org%3Aconda-forge+%2Fpython_check_max%3A+%223.13%22%2F+path%3Arecipe.yaml&type=code). This has frequently uncovered upstream feedstocks (or even upstream PyPI packages) which need help migrating, or better still, can be made to do a fraction of the PRs/builds with `noarch: python`, `version_independent`, etc.

"What is the latest supported" is kind of a thing to put a name on, and would probably only tick over when some significant fraction of "big rocks" (e.g. `numpy`, `grpcio`) have made their way through the `python3XX` migration season, but likely not before the migration is "closed". 

Some paths to building a process:
- hoist such a `conda-forge-pinning#python_max` (or `python_conda_forge_latest`)
- embed the "latest supported python" in `conda-smithy`, checking `R1-004` against the evaluated meta for _either_ the "theoretically" accurate `*` or the known latest 
- somehow publish a package which will bring a curated version along
  - seems kinda disasterous, but could handle `python-freethreading`, which will become more relevant when there is an `python-abi3t`
- softening to "checks more than one" 

Selfishly happy to PR any of the above to keep ❌ out of my inbox 😊

bollwyvl:
Checklist

References:
- fixes #2645


conda_smithy/linter/hints.py
```diff
@@ -392,26 +392,21 @@ def hint_redundant_python_min(meta, recipe_text, recipe_version, hints):
...
+    """
     for req in flatten_v1_if_else(run_reqs or []):
         if isinstance(req, str) and req.strip().split()[0] == "python" and "<" in req:
             return True
 
     for test in tests_section or []:
         if not isinstance(test, Mapping) or "python" not in test:
             continue
-        python_version = test.get("python", {}).get("python_version", {})
+        python_version = test["python"].get("python_version", [])
```
h-vetinari, line 406: is it assured that the test section _always_ has a `"python"` key? That seems like an unnecessary footgun. I'm happy to be convinced otherwise, but barring a more detailed argument[^1], I'd prefer to keep this as

[^1]: it's definitely not obvious from the callsites of `_python_tests_cover_latest`.
```suggestion
        python_version = test.get("python", {}).get("python_version", [])
```

conda_smithy/linter/hints.py
```diff
@@ -392,26 +392,21 @@ def hint_redundant_python_min(meta, recipe_text, recipe_version, hints):
...
         if isinstance(python_version, str):
             python_version = [python_version]
-        if not isinstance(python_version, list):
-            python_version = []
-        # Check that the latest-Python marker is the exact entry `"*"`. Since
-        # flatten_v1_if_else always returns a list, `"*" in ...` will never
-        # (substring-)match for version pins like `${{ python_min }}.*`;
-        # v1 also forbids bare `- *`, so we know it must be a string.
-        if "*" not in flatten_v1_if_else(python_version):
+        if len(set(flatten_v1_if_else(python_version))) < 2:
```
h-vetinari, line 409: I think we should be stricter here. `["3.10.1", "3.10.*"]` would pass this check, for example.

I'm thinking we'd want to ensure that it's basically exactly two values out of
```python
[f"3.{x}.*" for x in range(python_min_minor, python_min_minor + 6)] + ["*"]
```

h-vetinari (commented):
Thanks. One question and a request for tightening the check here.

conda_smithy/linter/hints.py
bollwyvl: The `if` two lines above is checking whether `python` is present.

h-vetinari: ah, apologies. couldn't see the forest for the trees (and had assumed the previous formulation must have had a good reason)

conda_smithy/linter/hints.py
bollwyvl: I tried a test with:
```yaml
context:
  python_min: "3.10"
  python_check_max: "3.10"
#...
  python_version: [${{ python_min }}.*, ${{ python_check_max }}.*]
```

... and it indeed passed without a hint, so I assume it is not looking at a rendered version, so if `python_min` is in the mix, there will be no way without the other suggestion of adding a new "well-known" variable for "the latest fully migrated-enough `python`."

h-vetinari: > so I assume it is not looking at a rendered version

We should be able to look at the rendered values in the linter...? 🤔

bollwyvl: In a PR case like the bot, sure. For per-keystroke cases like https://github.com/conda-forge/conda-forge.github.io/issues/2523, full rendering might be more expensive than is desirable, and would likely need to be gated.

h-vetinari: I'm pretty sure the linter is not anywhere close to running at per-keystroke speeds yet, but as an aspirational goal that's fair enough. Can we find a middle ground that one of the two values should be `${{ python_min }}.*`, and the other should _not_ contain the substring `python_min`. That would address my concern.

bollwyvl: The presence of `python_min` is handled elsewhere, so 756309e just checks whether it is in _all_ of them, and adds a test.

isuruf (approved):

h-vetinari (approved):
I installed smithy from this PR and tested that
```yaml
tests:
  - python:
      python_version:
        - ${{ python_min }}
        - ${{ python_min }}
```
and

```yaml
tests:
  - python:
      python_version:
        - ${{ python_min }}
        - ${{ python_min }}.*
```
both successfully cause a linter warning, which is good enough for me. Thanks

bollwyvl:
Thanks, all!

PR #2657, opened 2026-08-24
Don't ask for abi3audit when the test is conditional

killua156:
The `abi3audit` hint fires on recipes that already run `abi3audit`, when the test sits behind a conditional.

`_mentions_abi3audit` iterated the v1 `tests` list directly, so a test element that is an `if`/`then` block never matched: the mapping has `if`/`then` keys, not `requirements`/`script`. A guarded `abi3audit` test was invisible and the hint fired anyway:

```yaml
    tests:
      - python:
          imports:
            - mylib
      - if: is_abi3
        then:
          requirements:
            run:
              - abi3audit
          script:
            - if: unix
              then: abi3audit $SP_DIR/mylib/mylib.abi3.so -s -v
              else: abi3audit %SP_DIR%/mylib/mylib.pyd -s -v
```

`flatten_v1_if_else` can't be reused here. It treats every mapping as a conditional, but a v1 test element is itself a mapping, so a plain `- python: {imports: [...]}` entry would hit `req["then"]` and raise `KeyError`. This adds a small flattener for the tests list instead.

### Checked against a real feedstock

`conda-forge/finance_enums-feedstock` has this shape. Its `recipe.yaml` currently duplicates the whole `abi3audit` test into the `noarch: generic` alias output, which is what made this look like an alias problem. Running the hint over three variants of that recipe:

| variant | `main` | this PR |
| --- | --- | --- |
| upstream as-is (audit duplicated into the alias) | no hint | no hint |
| duplicate dropped from the alias output | no hint | no hint |
| audit guarded by `if: is_abi3` | **hint** | no hint |

So the alias was never the trigger, and that duplicated test in the alias is already unnecessary today. Guarding the audit with `if: is_abi3` is what trips the hint.

Separately, an alias output should not be asked to audit an extension module it does not ship, so `noarch: generic` is now skipped alongside `noarch: python`.

Both behaviours are covered by new parametrized tests. They fail on `main` and pass here; the eleven existing `abi3audit` cases are unaffected.

Checklist

* [x] Added a `news` entry with any new deprecations added to the `Deprecated` section.
* [ ] Regenerated schema JSON if schema altered (`python -m conda_smithy.schema`) (no schema change)
* [ ] Regenerated linter documentation if any rules were added, removed or modified (`python -m conda_smithy.linter.messages`) (no message change)

### DIFF
--- a/conda_smithy/linter/hints.py
+++ b/conda_smithy/linter/hints.py
@@ -537,11 +537,33 @@ def hint_abi3_cross_python_run_exports(
             return
 
 
+def _flatten_v1_test_elements(test_section) -> list:
+    """Expand top-level `if`/`then`/`else` blocks in a v1 `tests` list.
+
+    `flatten_v1_if_else` cannot be reused here because it treats every mapping
+    as a conditional, while a v1 test element is itself a mapping.
+    """
+    flattened = []
+    for test in test_section:
+        if isinstance(test, Mapping) and "if" in test:
+            for branch in ("then", "else"):
+                value = test.get(branch)
+                if isinstance(value, list):
+                    flattened.extend(_flatten_v1_test_elements(value))
+                elif value is not None:
+                    flattened.append(value)
+        else:
+            flattened.append(test)
+    return flattened
+
+
 def _mentions_abi3audit(test_section, recipe_version) -> bool:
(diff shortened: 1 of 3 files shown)


killua156:
The failing `tests` jobs are unrelated to this PR.

isuruf:
Thanks

PR #2544, opened 2026-05-04
fix: context rendering with variant

wolfv:
@baszalmstra found an issue converting the `vc-feedstock` to v1 recipe. The context section requires some values coming from the variant definition.

Rendering the context _without_ the variant failed the rerendering.

This still needs some testing. Alternatively, we could be more lenient in `rattler-build-conda-compat` and ignore lines in context that fail to render. Sister PR: https://github.com/conda-forge/rattler-build-conda-compat/pull/121

Checklist
* [ ] Added a ``news`` entry
* [ ] Regenerated schema JSON if schema altered (`python -m conda_smithy.schema`)
* [ ] Regenerated linter documentation if any rules were added, removed or modified (`python -m conda_smithy.linter.messages`)

Issue #2561, opened 2026-05-26
Generate linter skips automatically?


h-vetinari:
After #2408, we could consider generating the code corresponding to stuff like
```yaml
        linter:
            skip:
                - lint_noarch_selectors
```
from the JSON config. Sometimes feedstocks have a valid reason for ignoring a hint/lint, and having custom implementation for each skip seems like a waste of time if we could just as well generate it.

Probably we'd need to align
https://github.com/conda-forge/conda-smithy/blob/a393ff525c42701f401bd516f06cffd7343399ae/conda_smithy/schema.py#L69-L74
with the lint names in https://github.com/conda-forge/conda-smithy/blob/main/conda_smithy/data/linter-messages.json?

We may want to make this configurable in the JSON, e.g. `skippable: bool` (or `skippable: phrase` to enable transitioning away from the existing naming that dates back to discussion in #2090) whether a lint supports being skipped, or whether it's mandatory always-on (a boolean is something that we can easily flip, which I think would be a good trade-off compared to allowing everything to be skippable by default)

CC @mgorny @jaimergp

jaimergp:
Yep I want to work on that next, once we have calver and deprecation protocols, because we'll need to break API.

I assumed in my design that hints are skippable and lints are not, but the bool flag is ok too. The identifiers for the configs are supposed to be the XY-123 codes or the class name.

(Sorry I'm terse, replying from my phone on the go)

PR #2659, opened 2026-08-25
fix: recover pinned names that are still templates

baszalmstra:
v1 recipes can build an output name from a variant variable. At lint time there is no variant config, so that name never resolves and stays literal text with a `${{ ... }}` in it. The pin lints can't cope with that and report valid recipes.

Concretely, an output named `vcomp${{ vc_major }}` with `vc_major` coming from `conda_build_config.yaml`:

```yaml
outputs:
  - package:
      name: vcomp${{ vc_major }}
  - package:
      name: vc${{ vc_major }}_runtime
    requirements:
      run_exports:
        - ${{ pin_subpackage("vcomp" ~ vc_major) }}
```

```
pin_compatible should be used instead of pin_subpackage for `vcomp${{` because it is
not a known outputs of this recipe: ['vcomp${{ vc_major }}', 'vc${{ vc_major }}_runtime']
```

The pin is fine. `vcomp${{` is the bug.

Closes https://github.com/conda-forge/conda-smithy/pull/2544

## Cause

`pin_subpackage` and `pin_compatible` are stubbed during linting and render to `"<prefix><name>"` — `conda_smithy/utils.py` for v0, `rattler_build_conda_compat.jinja.objects` for v1. Both are `f"subpackage_pin {args[0]}"`, so the name is the entire remainder.

`check_pins` instead read `pin.split()[1]`, which assumes the name has no spaces. Unresolved templates do have spaces, so it got a fragment, and the fragment matched nothing.

## Fix

Take the remainder rather than the second token.

Nothing has to be skipped, which is the point. `subpackage_names` is rendered by the same code and holds the same unresolved template, so pin and output name compare equal exactly when they should, and both lints keep working on names that can't be resolved:

- `pin_subpackage` naming a real output stops being flagged
- `pin_subpackage` naming nothing is still flagged, now under its full name instead of `notanoutput${{`
- `pin_compatible` naming a real output is still flagged, which the old code also missed
- `pin_compatible` naming nothing stays quiet

A test each. Three of the four fail on `main`; the fourth is a regression guard that passes there for the wrong reason.

## Why not #2544

#2544 goes at the same problem by switching the linter to `resolve_recipe_metadata`. I tried it: `test_pin_compatible_in_run_exports[1]` and `test_pin_compatible_in_run_exports_output[1]` both fail, because that function deliberately does not render recipe bodies — `run_exports` included — so the `subpackage_pin ...` strings `check_pins` looks for are never produced and the v1 pin lints stop firing at all. That may be why it stalled.

This PR keeps the rendering and fixes the parsing instead.

## Relationship to rattler-build-conda-compat#139

Independent as code: this patch doesn't need it, and the tests here use a bare undefined variant variable so they reproduce on `main` with nothing else applied.

Not independent in practice for every recipe, though. `lint_pin_subpackages` still calls `render_recipe_with_context`, and for a recipe whose `context` section *computes* something from a variant variable (`${{ (vcver | split(".")) [0] }}`) that call raises `UndefinedError` before the pin check is ever reached. Such recipes need #139 as well. That's the shape that sent me here — conda-forge/vc-feedstock converting to v1.

## Testing

`tests/test_lint_recipe.py`: 4246 passed, 2 skipped.

`test_license_files_up_to_date` also fails, and it isn't this change. It fetches the SPDX list from `spdx/license-list-data` at run time and compares it to the bundled copy, which has gone stale — currently failing on #2617 and #2657 too, with `main` last green on 22 Aug. #2660 syncs it.

Checked while writing this, in case it matters to a reviewer: keyword arguments to the pin functions (`exact`, `upper_bound`, `max_pin`) don't add tokens, so the remainder really is just the name; `.strip()` doesn't change any previously passing case; and the third consumer of these stub strings, `_is_valid_requirement_spec` in `linter/hints.py`, only skips strings carrying these prefixes and never parses a name out of them, so it's unaffected.


h-vetinari:
> The trade-off: a genuinely wrong `pin_subpackage` goes unreported when its argument is variant-dependent. That seems clearly better than flagging correct recipes, and matches what the recipe-name lint already does.

Not that I feel super-strongly about it, but we could also provide an opt-out of that check, while keeping the false positives by default. We have a pattern for this in `conda-forge.yml` already, e.g.
```yaml
linter:
  skip:
  - lint_stdlib
```

This will eventually be unified with #2561, but the principle is there already

baszalmstra:
> > The trade-off: a genuinely wrong `pin_subpackage` goes unreported when its argument is variant-dependent. That seems clearly better than flagging correct recipes, and matches what the recipe-name lint already does.
> 
> Not that I feel super-strongly about it, but we could also provide an opt-out of that check, while keeping the false positives by default.

I asked another clanker and they found a better fix. Please check it again!

beckermr (approved):

PR #2661, opened 2026-08-25
Add hint for unforwarded v1 script variant variables

wolfv:
This adds a recipe v1 lint hint when a Bash or batch build script references a variant variable that is not referenced in `recipe.yaml` and therefore might not be forwarded by rattler-build.

The check compares `$VAR` / `${VAR}`, `%VAR%`, and `!VAR!` references against keys from recipe variant files, rendered `.ci_support` configs, and migrations. It recommends explicit forwarding through `build.script.env` rather than adding implicit script scanning to rattler-build.

Related to https://github.com/prefix-dev/rattler-build/issues/2587 and https://github.com/conda-forge/conda-forge.github.io/issues/1744#issuecomment-5324923625.

Checklist
* [x] Added a `news` entry.
* [x] Regenerated linter documentation (`python -m conda_smithy.linter.messages`).

Tests:
* `pytest tests/test_lint_recipe.py -k unforwarded_v1_script_variant_variables -q`
* `pytest tests/test_linter_messages_registry.py -q`
* `pre-commit run --files ...`

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -1031,6 +1031,20 @@
       "path": "recipe/recipe.yaml"
     },
     {
+      "name": "UnforwardedVariantVariables",
+      "identifier": "R1-008",
+      "category": "R1",
+      "kind": "hint",
+      "added_in": "2026.8",
+      "deprecated_in": "",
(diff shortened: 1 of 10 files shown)


beckermr (approved):
LGTM!

I had some concern that hinting for all scripts found in the recipe dir might cause false positives, but I don't think folks should be storing unrelated scripts there anyways, so hopefully this is ok for now.

isuruf (changes requested):
I think this is a rattler-build bug. There are hundreds of v0 feedstocks if converted to v1 would hit this hint. We should fix this in rattler-build instead of going around like this

beckermr:
I think @isuruf is referring to this PR as well:

xref: https://github.com/prefix-dev/rattler-build/pull/2748

PR #2665, opened 2026-08-28
Fix default `workflow_settings` paths values for unix/win cross

### LINKED ISSUE #2658: BUG: Artefact creation for linux->win cross-platform builds
Not sure if this ever worked (since it's a rare combination), but I noticed in https://github.com/conda-forge/ctng-compilers-feedstock/pull/223 that the artefact creation for the windows builds simply fails with
```
conda-build directory does not exist
Error: Process completed with exit code 1.
```
Presumably the code here is not ready for dealing with windows-y things on a unix host. CC @mgorny, since you worked extensively on that part of the code recently

mgorny:
Checklist

Fixes #2658

Fix the default ``build_workspace_dir`` and ``tools_install_dir`` values for cross Unix-Windows builds, by basing them on the build platform rather than the target platform.

### DIFF
--- a/tests/test_configure_feedstock.py
+++ b/tests/test_configure_feedstock.py
@@ -2916,16 +2916,16 @@ def test_tools_build_paths_gha(py_recipe, jinja_env, win64_label: str, cross: bo
     with conda_build_yml.open() as f:
         workflow = yaml.safe_load(f)
 
-    built_on_linux_64 = "win_64" if cross else "linux_64"
-    built_on_osx_arm64 = "win_arm64" if cross else "osx_arm64"
-    built_on_win_64 = "linux_64" if cross else "win_64"
-    built_on_win_arm64 = "osx_arm64" if cross else "win_arm64"
+    target_built_on_linux_64 = "win_64" if cross else "linux_64"
+    target_built_on_osx_arm64 = "win_arm64" if cross else "osx_arm64"
+    target_built_on_win_64 = "linux_64" if cross else "win_64"
+    target_built_on_win_arm64 = "osx_arm64" if cross else "win_arm64"
 
     expected = {
-        built_on_linux_64: ("~/miniforge3", "build_artifacts"),
-        built_on_osx_arm64: ("~/miniforge3", "~/miniforge3/conda-bld"),
-        built_on_win_arm64: (r"C:\Miniforge", r"C:\\bld\\"),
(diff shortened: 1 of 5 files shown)


mgorny:
I'm still working on adding tests for this.

tests/test_configure_feedstock.py
```diff
@@ -2946,16 +2974,21 @@ def test_tools_build_paths_azure(py_recipe, jinja_env):
...
 
+    built_on_linux_64 = "win_64" if cross else "linux_64"
+    built_on_osx_arm64 = "win_arm64" if cross else "osx_arm64"
+    built_on_win_64 = "linux_64" if cross else "win_64"
+    built_on_win_arm64 = "osx_arm64" if cross else "win_arm64"
+
     expected = {
         "linux": {
-            "linux_64": ("~/miniforge3", "build_artifacts"),
+            built_on_linux_64: ("~/miniforge3", "build_artifacts"),
```
h-vetinari: This reads very strange to me, especially if the value of `built_on_linux_64` is `"win_64"`, in which case it's built _on_ `win-64`. If I understand the intent correctly, this should be named something like `build_platform_for_linux_64` or maybe `bp_for_linux_64`? And once that's renamed, the question arises how the unix paths would work on a windows agent?

h-vetinari (commented):
Thanks a lot. Looks good at first glance, but the tests confuse the heck out of me.

tests/test_configure_feedstock.py
mgorny: No, it's host platform. I.e. `built_on_linux_64` means "this host platform is built on `linux-64` build platform". That's why you have unix paths on linux/osx keys, and Windows paths on win keys.

mgorny: I'm not particularly fond of that part, but I couldn't come up with a cleaner way of expressing it, short of splitting the tests in two and therefore duplicating all the logic.

h-vetinari: I think this warrants being slightly more verbose, in the interest of being nice to the reader. Maybe `target_built_on_linux_64` etc.

h-vetinari:
LGTM (thanks a lot), ~except that we're missing a new entry. Can you please add one?~

h-vetinari (approved):
Thanks! :)

PR #2666, opened 2026-08-29
Zip MACOSX_DEPLOYMENT_TARGET/MACOSX_SDK_VERSION with c_stdlib_version

hmaarrfk:
> [!NOTE]
> **Disclaimer:** this PR was written by Claude, Anthropic's coding agent, acting as my bot under my direction. I've reviewed it and I take responsibility for it, but I'm flagging the provenance so you can weight the analysis accordingly. Opening as a draft for that reason — happy to iterate on the approach before it's considered for real.

### Problem

`_merge_deployment_target` rewrites `MACOSX_DEPLOYMENT_TARGET` and `MACOSX_SDK_VERSION` per variant so they track `c_stdlib_version`. Neither key is a loop variable of its own, and nothing records that relationship in `zip_keys` — they survive only via the `always_keep_keys` set. So `_get_used_key_values_by_input_order` falls through to its "rest of the keys that are not zipped" branch and hands *every* config the full list of values.

That is invisible while `c_stdlib_version` has a single value, which is nearly every feedstock. But a feedstock that builds several macOS deployment target tiers side by side gets every tier's value written into every one of its `.ci_support/osx_*` files:

```yaml
# .ci_support/osx_arm64_c_stdlib_version14.5python3.13.____cp313.yaml
MACOSX_DEPLOYMENT_TARGET:
- '14.5'
- '26.0'      # <- belongs to the other tier
MACOSX_SDK_VERSION:
- '14.5'
c_stdlib_version:
- '14.5'
```

`MACOSX_SDK_VERSION` escapes only because that feedstock happens to zip it explicitly.

### Why it matters

The build tool then reads the two-valued key as a variant axis of its own and builds each tier twice. From `rattler-build build --render-only` on the 14.5 config above, two outputs:

| build string | `c_stdlib_version` | `MACOSX_DEPLOYMENT_TARGET` |
| --- | --- | --- |
| `py313h5605d13_1400` | 14.5 | 14.5 |
| `py313hd7257d0_1400` | 14.5 | 26.0 |

Same name, version and build number; they differ only in the hash, and one of them carries a deployment target that has nothing to do with its own tier. On conda-forge/mlx-feedstock#104 the osx jobs took 15–25 min with the duplicate present against 7–14 min once each config built once.

The current workaround is to hand-edit the offending value out of each rendered file after every re-render, which silently regresses the moment anyone (or the autotick bot) re-renders.

### Fix

Add the two keys to the `zip_keys` group that holds `c_stdlib_version`, creating one when the feedstock does not already zip it, so the existing zip machinery in `_get_used_key_values_by_input_order` and `break_up_top_level_values` slices them per config.

Guarded on `c_stdlib_version` actually having more than one distinct value, so **single-tier feedstocks render byte-for-byte as before** and this introduces no mass re-render churn. I verified that directly: re-rendering a single-tier feedstock with and without the patch produces identical `.ci_support` output.

### Testing

- New `test_stdlib_multiple_deployment_targets` plus a `stdlib_multiple_deployment_targets_recipe` fixture. It fails on `main` (`assert content["MACOSX_DEPLOYMENT_TARGET"] == [version]`) and passes with the fix.
- Full `tests/test_configure_feedstock.py`: 269 passed, 13 skipped.
- `tests/test_configure_feedstock.py` + `tests/test_lint_recipe.py`: 4524 passed. Two `test_lint_recipe.py` failures (`test_lint_recipe_parses_v1_duplicate_keys`, `test_license_files_up_to_date`) reproduce on unpatched `main` and are unrelated.
- Both branches of the new helper are exercised: the pre-existing-zip-group case (mlx's real config) and the no-zip-group case (a synthetic CBC with a bare two-value `c_stdlib_version`).

### Alternatives considered

Declaring `MACOSX_DEPLOYMENT_TARGET` in the recipe-local `conda_build_config.yaml` and zipping it there does render correctly, but it trips `lint_osx_pins` (*"needs to be removed or replaced by `c_stdlib_version`"*), which has no `linter.skip` guard, and it promotes the key to a top-level loop variable so every osx job is renamed to `osx_arm64_MACOSX_DEPLOYMENT_TARGET14.5c_stdlib_version14.5python3.13...`. Adding it to `zip_keys` without declaring values fails earlier, in conda-build: `zip_key entry MACOSX_DEPLOYMENT_TARGET in group frozenset({...}) does not have any settings`.

mlx-feedstock has since worked around this on its own side with rattler-build's `build.variant.ignore_keys`, which is available because the deployment target in that file is inert for the build anyway — `MACOSX_DEPLOYMENT_TARGET` in the build environment comes from the `activate.d` script of `macosx_deployment_target_osx-arm64`, which sets it (and `-mmacosx-version-min`, and `-DCMAKE_OSX_DEPLOYMENT_TARGET`) unconditionally from its own version. That workaround is recipe-format-specific though, and the rendered files are still wrong.

### Motivating case

conda-forge/mlx-feedstock#104 — an osx-arm64 matrix building 14.5 and 26.0 tiers so macOS 26 users get MLX's NAX/Metal 4 kernels while older macOS keeps a working package.

### DIFF
--- a/conda_smithy/configure_feedstock.py
+++ b/conda_smithy/configure_feedstock.py
@@ -691,8 +691,60 @@ def _merge_deployment_target(container_of_dicts, has_macdt):
             }
         )
         result.append(new_dict)
+    result = _zip_osx_versions_with_stdlib(result)
     # ensure we keep type of wrapper container (set stays set, etc.)
     return type(container_of_dicts)(result)
+
+
+def _zip_osx_versions_with_stdlib(list_of_dicts):
+    """Record that MACOSX_DEPLOYMENT_TARGET/MACOSX_SDK_VERSION follow c_stdlib_version.
+
+    ``_merge_deployment_target`` above derives both keys from ``c_stdlib_version``
+    for every variant, so the three co-vary by construction.  Neither of the two
+    is a loop variable of its own though, so unless that relationship is spelled
+    out in ``zip_keys``, a feedstock that builds several deployment-target tiers
+    side by side (i.e. one that gives ``c_stdlib_version`` more than one value)
+    gets *all* the tiers' values written into *every* rendered .ci_support file,
+    and the build tool then builds each tier once per tier.
+
+    This is a no-op unless ``c_stdlib_version`` actually varies, so the single
+    tier case -- which is nearly every feedstock -- renders exactly as before.
+    """
+    stdlib_versions = {
+        var_dict["c_stdlib_version"]
+        for var_dict in list_of_dicts
+        if "c_stdlib_version" in var_dict
+    }
+    if len(stdlib_versions) < 2:
+        return list_of_dicts
+
+    result = []
+    for var_dict in list_of_dicts:
+        keys_to_zip = [
+            key
+            for key in ("MACOSX_DEPLOYMENT_TARGET", "MACOSX_SDK_VERSION")
+            if key in var_dict
+        ]
+        if "c_stdlib_version" not in var_dict or not keys_to_zip:
+            result.append(var_dict)
+            continue
+
+        groups = var_dict.get("zip_keys", [])
+        # zip_keys may be a single group given as a flat list of keys
+        if groups and not isinstance(groups[0], (list, tuple)):
+            groups = [groups]
+        groups = [list(group) for group in groups]
+
+        for group in groups:
+            if "c_stdlib_version" in group:
+                break
+        else:
+            group = ["c_stdlib_version"]
+            groups.append(group)
+        group.extend(key for key in keys_to_zip if key not in group)
+
+        result.append(HashableDict({**var_dict, "zip_keys": groups}))
+    return result
 
 
 def _collapse_subpackage_variants(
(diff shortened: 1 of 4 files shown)


h-vetinari:
This approach does not make a lot of sense to me. We should just remove `MACOSX_DEPLOYMENT_TARGET` IMO, it's just legacy cruft at this point. We kept it when the stdlib-machinery was introduced to not also have to do surgery on ci-setup, especially while the transition was ongoing.

But now that 2-3 years have gone by, it's fair to expect that all feedstocks (and their osx variant configs) have a `c_stdlib_version`, so we can change https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/main/recipe/download_osx_sdk.sh to reflect that.

Then we can remove this whole merge logic.

hmaarrfk:
You are correct that this doesn’t make sense. 

Thanks reading my mind and finding the underlying bug. 

I’m trying to build a package for different deployment targets and it seems problematic on OSX

PR #2672, opened 2026-09-03
Lint on redundant `build_platform` and `provider` entries

### LINKED ISSUE #2654: Lint on doubly-defined platforms
To specify a non-default platform `foo`  in `conda-forge.yaml`, one needs to either add
```yaml
build_platform:
  foo: bar
```
to cross-compile from `bar`, or 
```yaml
provider:
  foo: default  # or name of CI provider
```
to compile natively (resp. in emulation) on the respective provider.

If both are present, `build_platform:` wins, but silently - which leaves lots of stale config entries lying around. I think we should add a lint if (non-default) platforms are defined both in `build_platform:` as well as in `provider:`. Only one can be active at a time anyway.

We may also want to lint on redundant `build_platform:` specifications like
```yaml
build_platform:
  foo: foo
```
i.e. native compilation, which is the default anyway.

mgorny:
Checklist

Fixes #2654

### DIFF
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -208,6 +208,18 @@
       "path": ""
     },
     {
+      "name": "RedundantBuildPlatform",
+      "identifier": "CF-015",
+      "category": "CF",
+      "kind": "lint",
+      "added_in": "2026.9",
+      "deprecated_in": "",
+      "documentation": "Lint when `build_platform` is the same as the platform.",
+      "message": "Platform ${platform} specifies redundant build_platform=${platform}. Please remove the unnecessary entry.\n",
+      "examples": [],
+      "path": ""
+    },
+    {
       "name": "NoDuplicateKeys",
       "identifier": "FC-001",
       "category": "FC",
--- a/conda_smithy/data/linter-messages.json
+++ b/conda_smithy/data/linter-messages.json
@@ -196,6 +196,18 @@
       "path": ""
     },
     {
+      "name": "RedundantBuildPlatformAndProvider",
+      "identifier": "CF-014",
+      "category": "CF",
+      "kind": "lint",
+      "added_in": "2026.9",
+      "deprecated_in": "",
+      "documentation": "Lint when the same platform is listed in `build_platform` and `provider`.",
+      "message": "Platform ${platform} specifies both build_platform=${build_platform} and provider=${provider}. Use the former for cross-compilation or the latter for native build.\n",
+      "examples": [],
+      "path": ""
+    },
+    {
       "name": "NoDuplicateKeys",
       "identifier": "FC-001",
       "category": "FC",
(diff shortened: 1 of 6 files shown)


isuruf (changes requested):
I'd rather this was added to CondaForgeYAMLCleanUp in the bot instead of linting and asking the maintainers to fix it themselves.

mgorny:
Well, I'd love to see that too. Should I stack on top of https://github.com/conda-forge/conda-forge-bot/pull/6240 that's waiting for 3 months now, or do it separately?

h-vetinari:
> I'd rather this was added to CondaForgeYAMLCleanUp in the bot instead of linting and asking the maintainers to fix it themselves.

Why not both? the bot may be able to apply the cleanup for each PR _it_ makes, but for changes by maintainers themselves, we should also lint if they introduce this redundancy

jaimergp:
What about starting with the migrator for some time so active feedstocks benefit from it, and then we ship the linter rules? A couple months or so?

isuruf:
> but for changes by maintainers themselves, we should also lint if they introduce this redundancy

It's rare that maintainers add this redundancy. These are usually bot's work where we switched from cross compile to native back and forth. How do we lint that this is introduced by the maintainers?

isuruf:
> What about starting with the migrator for some time so active feedstocks benefit from it, and then we ship the linter rules? A couple months or so?

yes, that's okay, but I'd leave the lint for a half a year so.

h-vetinari:
> It's rare that maintainers add this redundancy.

That's just not true. Contributors are getting this wrong all the time (probably from cargo culting existing recipes, not least since the bots also often [created](https://github.com/conda-forge/conda-forge-bot/issues/6281) duplication, and it's not well-documented what's required to get a platform going, so people add both "to be sure"). This happens even to very experienced contributors; here's two such examples from the last few days where this lint would have helped:
* https://github.com/conda-forge/mamba-feedstock/pull/393 (`linux_ppc64le` still duplicated)
* https://github.com/conda-forge/conda-feedstock/pull/320 (all non-x64 linux platforms duplicated)

It makes zero sense to me to wait with the lint. By all means, let's let the bots clean up the bulk of it, but let's also avoid people reintroducing this regularly during that time.

isuruf:
I'm against any lint that needs the maintainers' immediate attention when it's a non-important lint. Especially for something that's being there for years.

h-vetinari:
> I'm against any lint that needs the maintainers' immediate attention when it's a non-important lint.

The easy fix for this concern is to make it a hint rather than a lint.

Where we appear to disagree is that for me, this is a teaching opportunity for people to understand the setup between `build_platform` and `provider` better, rather than cargo-culting whatever they see working (or "working") somewhere else. In other words, the visibility is not a negative, and if we let the bot clean up all of those instances before, then fewer people will understand this aspect in the end.

Using a hint seems to achieve both goals:
* we let people know about redundant config and why that's better avoided
* it doesn't have the urgency of a ❌ in CI

PR #2682, opened 2026-09-14
feat: add `upload_branches` configuration option; deprecate `upload_on_branch`; deprecate uploading from all branches

### LINKED ISSUE #1904: add default branches that are ignore by smithy for uploads
Right now, one can set an appropriate key in the conda-forge.yml to have smithy only do uploads for certain branches. This is useful, but @jaimergp and I discussed a separate scheme that we think could be a nice default and enable some new things. The idea is to always disallow uploads from anything that matches the following globs/regex:

 - `_*` (i.e., any branch that starts with an underscore)
 - `^[^\S].-patch-[0-9].$` (i.e., any branch that matches the pattern of default GH branch names for edits in the UI, `<username>-patch-<number>`.
 - any branch that matches the default branch name for PR reversions (need to look this up)

If we moved to this system, then we'd avoid common user errors of uploads happening when they do not  fork the feedstock and we could use branches that start with `_` for automated tasks like rerendering.

Thoughts @conda-forge/core?

beckermr:
Checklist

This PR 

- [x] adds a new config field `upload_branches`
- [x] deprecates `upload_on_branch`
- [x] deprecates uploading from any branch
- [ ] adds tests for the above

It needs an associated PR on conda-forge-ci-setup to support the new UPLOAD_BRANCHES environment variable.

closes #1904


conda_smithy/schema.py
```diff
@@ -1073,6 +1074,35 @@ class ConfigModel(BaseModel):
...
+        ```yaml
+        upload_branches:
+          - main
+        ```
+        """),
+    )
+
+    upload_branches: list[str] = Field(
+        default=["**"],
+        deprecated=True,
```
h-vetinari: also deprecating the new spelling sounds like a C&P error?

conda_smithy/schema.py
```diff
@@ -1073,6 +1074,35 @@ class ConfigModel(BaseModel):
...
+        description=cleandoc("""
+        This parameter lists all branches from which built packages are uploaded to the
+        `conda-forge` channel. This field supports recurisve glob patterns (e.g., `v*`,
```
h-vetinari:
```suggestion
        `conda-forge` channel. This field supports recursive glob patterns (e.g., `v*`,
```

conda_smithy/schema.py
```diff
@@ -1073,6 +1074,35 @@ class ConfigModel(BaseModel):
...
+        `**/v*`, etc.). You can use this configuration option to enable uploads
+        on the `main` branch and version branches prefixed with `v` by setting it to
+
+        ```yaml
+        upload_branches:
+          - main
+          - v*
+        ```
+
+        The current default of `**` is deprecated as of `conda-smithy` version 2026.9.15.
```
h-vetinari, line 1102: sounds like we'll want to add a lint after the initial transition

beckermr:
cc @conda-forge/core

conda_smithy/schema.py
```diff
@@ -1073,6 +1074,35 @@ class ConfigModel(BaseModel):
...
+        description=cleandoc("""
+        This parameter lists all branches from which built packages are uploaded to the
+        `conda-forge` channel. This field supports recurisve glob patterns (e.g., `v*`,
+        `**/v*`, etc.). You can use this configuration option to enable uploads
+        on the `main` branch and version branches prefixed with `v` by setting it to
+
+        ```yaml
+        upload_branches:
+          - main
+          - v*
```
h-vetinari, line 1099: I've always preferred `vN.M.x` as a way to spell branches names for maintenance versions `N.x` or `N.M.x`; however, I've also always kept existing branch patterns. Lots of feedstocks with relevant migration branches (python, LLVM, etc.) only use the number as the branch name, currently.

What I'm getting at is that it'd be nice to do the equivalent of a regex à la `'^\d.*'` (or `'^\d\.x'`, or `'^\d(\.\d)*\.x'`), i.e. only those that start with a digit, or whatever pattern the feedstock has chosen.

If that isn't easily possible with glob syntax (or some special opt-in for "allow version-only branches" where we control the implementation), then I guess we can also rename the respective branches on those feedstocks, where relevant.

h-vetinari (commented):
I fully support the direction; thanks for working on this!

conda_smithy/schema.py
carterbox: I also have this concern. What is the story for versioned branches don't use the v-prefix? Can we use the python regex module instead of glob?

beckermr: The github actions template for smithy uses the branches in the `push` event, and Github only supports glob-style patterns. See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushbranchestagsbranches-ignoretags-ignore. 

So unless we want to run CI for push on all branches, we need to stick to github's supported syntax. 

Github supports more glob syntax than Python supports. We will need to use the `wcmatch` package (or similar) to allow for patterns like `+([0-9]).+([0-9]).+([0-9])` which match the example branch names without `v`. For example:

```python
In [10]: wcmatch.glob.globmatch("5.6.7", "+([0-9]).+([0-9]).+([0-9])", flags=wcmatch.glob.EXTGLOB)
Out[10]: True

In [11]: wcmatch.glob.globmatch("50.6.7", "+([0-9]).+([0-9]).+([0-9])", flags=wcmatch.glob.EXTGLOB)
Out[11]: True
```

This logic will go in the ci support package.


If we use regex here instead, it will break the github template as it is written now.

conda_smithy/schema.py
beckermr: What do you want to lint on?

conda_smithy/schema.py
beckermr: This was meant to be a deprecation of the "**".

conda_smithy/schema.py
h-vetinari: I meant linting on
```yaml
upload_branches:
  - '**'
```

conda_smithy/schema.py
h-vetinari: IMO we don't need a full regex-based solution; I'm fine with extended globs (assuming they're not too horrible to write for feedstock maintainers). I'd also be fine to _not_ support branches beginning with a digit; though then we should probably provide a small utility in admin-requests that renames branches from (e.g.) `20.x` to `v20.x`, and updates `abi_migration_branches` as well...

beckermr: It turns out github's filter syntax is not support by wcmatch. So we'll have to solve that.

isuruf: We could support any glob syntax python supports by quering the branch names and listing any that matches when rerendering

conda_smithy/schema.py
beckermr: Ahhhh, yes that makes sense! I will make that a hint now that says it is deprecated, and then we can make it a lint later for the actual removal.

conda_smithy/schema.py
beckermr: I'd rather not do that @isuruf because the webservices rerenders on a copy of the files in the repo without the full git history.

isuruf: `configure_feedstock.py` already asks for some details from azure about the build id. We could do the same from pygithub API (with no token or a token that has no permissions)

beckermr: I'd rather write the code to support any pattern allowed here:

Screenshot 2026-09-14 at 12 07 00 PM

I think it is a pretty simple modification to the python stdlib code for converting globs to regex in `fnmatch`.

jaimergp:
Two things:

- You can leave Travis, Drone and Appveyor templates unmodified, since they are deprecated and scheduled for deprecation in October.
- What's the story for different values of `upload_branches` in different branches. e.g. if we have `main`'s value set to `[main, lts/*]` but `lts/0.12.x` has a different value, which one is used for rerenders? `upload_on_branch` needed to be set per branch so I'm not sure what the intended approach is now.

The way I thought about doing this was to:

- Create an `admin-migrations` migration that traversed all feedstocks and annotated the branch name in each non-`main`'s `conda-forge.yml`'s `upload_on_branch` field. In pseudocode:
  ```python
  for feedstock in feedstock:
    for branch in feedstock.branches:
      if branch.name in ("main", "master"):
        continue
      cfyml = edit_yaml("conda-forge.yml")
      cfyml.setdefault("upload_on_branch", branch.name)
      cfyml.save()
   ```
- Once migrated, swap the default for `upload_on_branch` to `main` only.
- Any new branches after this need to explicitly set `upload_on_branch` and rerender.
- We could also add a conda-forge-admin command to create LTS branches more easily.

beckermr:
Ack you are totally right!

beckermr:
Let me start over 🤦  with just the new hint.

conda_smithy/schema.py
carterbox: I'm convinced that trying to use regex instead is not worth the effort, but I think `v*` is still a very broad glob that we should discourage and provide alternatives to. `v1*` is better, but supported the listed patterns would even better.

PR #2686, opened 2026-09-15
Support native linux_riscv64 builds on GitHub Actions

velonica0:
## Problem
linux_riscv64 is in the schema (via known_subdirs), but the hardcoded platform tables in configure_feedstock.py were never updated for it, so a native riscv64 build on GitHub Actions cannot be rendered:

```
provider:
  linux_riscv64: github_actions   # or `default`, which resolves to github_actions
  File "conda_smithy/configure_feedstock.py", line 1939, in _github_actions_specific_setup
    data["gha_os"] = GITHUB_ACTIONS_RUNS_ON[data["build_platform"]]["os"]
KeyError: 'linux-riscv64'
```

This matters now: conda-forge/docker-images#337 added linux-anvil-riscv64:rocky10, conda-forge/conda-forge-pinning-feedstock#8942 adds it to docker_image, and conda-forge/conda-forge-bot#6772 switches the riscv64 migrator to native builds via provider: {linux_riscv64: default} — every PR that migrator opens would fail at re-render.

## Changes
- add linux_riscv64 to ALL_PLATFORMS, DEFAULT_PROVIDERS, NATIVE_CI_PROVIDER, FANCY_PLATFORM_NAMES, GITHUB_ACTIONS_RUNS_ON, mirroring linux_ppc64le (emulated on ubuntu-latest; generic linux self-hosted label only, as GitHub defines none for riscv64)
- list linux_riscv64 in the provider docstring; regenerate conda-forge.json
- regression test for github_actions / default / emulated

## Verified
- Native: re-rendering simdjson-feedstock with provider: {linux_riscv64: github_actions} now emits the linux_riscv64_ job (os: ubuntu, runs_on: ['ubuntu-latest'], binfmt step). DOCKER_IMAGE is still the linux-anvil-comp7 fallback until the pinning PR lands — out of scope here.
- Cross: re-rendering snappy-feedstock (build_platform: linux_riscv64: linux_64) with released vs patched conda-smithy gives identical output.


conda_smithy/configure_feedstock.py
```diff
@@ -155,6 +157,7 @@
     "linux_64": "github_actions",
     "linux_aarch64": "github_actions",
     "linux_ppc64le": "travis",
```
h-vetinari, line 159: travis as a CI provider is deprecated; let's switch the default to GHA as well.
```suggestion
    "linux_ppc64le": "github_actions",  # emulated
```

tests/test_configure_feedstock.py
```diff
@@ -2514,6 +2514,47 @@ def test_github_actions_labels(py_recipe, jinja_env, label):
         raise AssertionError("Bad label? Check test parameters.")
 
 
+@pytest.mark.parametrize("provider", ["github_actions", "default", "emulated"])
+def test_github_actions_native_linux_riscv64(py_recipe, jinja_env, provider):
```
h-vetinari: It's good practice to write a test, but I think the benefit of this one is marginal-to-none. There's nothing riscv-specific in the code that's being added in this PR, it's just extra configuration for things that are already existing and very stable.

So IMO: either make the test less riscv-specific (cover ppc, s390, aarch, etc.), or delete it.

news/linux-riscv64-github-actions.rst
```diff
@@ -0,0 +1,25 @@
+**Added:**
+
+* Native ``linux_riscv64`` builds can now be rendered on GitHub Actions
+  (``provider: {linux_riscv64: github_actions}`` or ``default``); previously
+  this failed with ``KeyError: 'linux-riscv64'``. (#PR)
```
h-vetinari:
```suggestion
  using the usual interface (``provider: {linux_riscv64: X}`` or where ``X``
  is ``github_actions`` or ``default``) in ``conda-forge.yml``. (#2686)
```

h-vetinari (commented):
Thanks for this

conda_smithy/configure_feedstock.py
```diff
@@ -154,8 +156,9 @@
 NATIVE_CI_PROVIDER = {
     "linux_64": "github_actions",
     "linux_aarch64": "github_actions",
-    "linux_ppc64le": "travis",
-    "linux_s390x": "travis",
+    "linux_ppc64le": "github_actions",  # emulated
```
isuruf: `native` should be a native provider. Please revert this change. Or make it a dummy. Don't add an emulated provider here

h-vetinari: hm yeah, that's fair. I guess we should error out if someone requests `provider: {<platform>: native}` where we don't actually have a native provider set up

luhenry: What does “native” mean here? As-in native hardware? Because the riscv64 machines are not emulation, they are proper hardware

conda_smithy/configure_feedstock.py
```diff
@@ -200,6 +204,16 @@
         "hosted_labels": ("ubuntu-latest",),
         "self_hosted_labels": ("linux",),
     },
+    "linux-riscv64": {
+        "os": "ubuntu",
+        "hosted_labels": ("ubuntu-latest",),
```
luhenry, line 209: It must be `ubuntu-24.04-riscv` to run on native hardware cc @isuruf

luhenry (changes requested):

conda_smithy/configure_feedstock.py
h-vetinari: GitHub itself does not offer native risc-v runners. We can revisit this default once we've hooked up runners from RISE and/or OpenKylin

velonica0: linux_ppc64le/linux_s390x are back to travis, and linux_riscv64 is a None dummy (the key has to exist because test_platforms_populated requires all four tables to share the same keys).

luhenry: We discussed at last week’s Conda community meeting we would start with runners from RISE (as they are available already through GitHub app) and we can evolve after indeed. We can discuss at this weeks one again if you want

conda_smithy/configure_feedstock.py
```diff
@@ -2898,6 +2907,12 @@ def _load_forge_config(forge_dir, exclusive_config_file, forge_yml=None):
 
     for plat, ci in NATIVE_CI_PROVIDER.items():
         if config["provider"][plat] == "native":
+            if ci is None:
+                raise RuntimeError(
+                    f"No native CI provider is available for '{plat}'; "
+                    "use 'default'/'emulated' instead, or cross-compile via "
+                    "'build_platform'."
+                )
```
h-vetinari, line 2915: Thanks, that looks good. Could you add a simple test for this, please?

news/linux-riscv64-github-actions.rst
```diff
@@ -0,0 +1,26 @@
+**Added:**
+
+* Native ``linux_riscv64`` builds can now be rendered on GitHub Actions using
+  the usual interface (``provider: {linux_riscv64: X}`` where ``X`` is
+  ``github_actions`` or ``default``) in ``conda-forge.yml``. (#2686)
```
h-vetinari, line 5: As of the current state of the PR, we're still only using emulated builds here when someone does `linux_riscv64: github_actions`. Of course we could merge this as-is so that at least the rerenders pass (and this would match what we're doing for `provider: {linux_ppc64le: default}` too). However, I'm not a big fan of silently falling back to emulation; I'm wondering if this PR by itself is really useful, or whether we should take the time to figure out the interaction with the runners mentioned by @luhenry ...

luhenry: I don't see the point of using emulation which will make build 20-30x slower, especially when we have hardware available. The only point I would see is if it enables testing (which it does in some cases) but again with the available hardware it's the same, just faster.

conda_smithy/configure_feedstock.py
isuruf: Yes, let's use `ubuntu-24.04-riscv`

conda_smithy/configure_feedstock.py
```diff
@@ -155,6 +157,7 @@
     "linux_64": "github_actions",
     "linux_aarch64": "github_actions",
     "linux_ppc64le": "travis",
+    "linux_riscv64": None,  # no CI service offers native riscv64 runners
```
isuruf, line 160: Let's use github_actions here

conda_smithy/configure_feedstock.py
```diff
@@ -144,6 +145,7 @@
     "linux_64": "github_actions",
     "linux_aarch64": "github_actions",
     "linux_ppc64le": "github_actions",  # emulated
+    "linux_riscv64": "github_actions",  # emulated
```
isuruf, line 148: Remove `emulated` comment here

news/linux-riscv64-github-actions.rst
luhenry: @velonica0 would you be open to hook the OpenKylin machines through the RISE GitHub app so that we would have best of both worlds: 1. An easy to use GitHub app for Conda-forge, and 2. The faster machines of OpenKylin for exclusive use of the conda-forge organization? Feel free to reach out at luhenry@qti.qualcomm.com if you want to discuss it further.

conda_smithy/configure_feedstock.py
velonica0: Done in e0d6be7.

conda_smithy/configure_feedstock.py
velonica0: Added test_native_provider_unavailable_raises in e0d6be7

news/linux-riscv64-github-actions.rst
velonica0: @luhenry thanks, we're interested in connecting the openKylin machines through the RISE app. I'll follow up by email to work out the details

PR #2690, opened 2026-09-17
Lint for missing stdlib("c") with compiler("ocaml")

MementoRC:
Checklist
* [x] Added a ``news`` entry

OCaml links native code through the C toolchain.
A recipe with only compiler("ocaml") gets no pinned sysroot.
It also gets no __glibc run-export.
The binary can then need a newer glibc than it declares.

Example: conda-forge/ocaml-dune-feedstock#17.
dune failed with "GLIBC_2.38 not found" in an alma9 container.

This adds ocaml to the compilers that require stdlib("c").
It uses the same check as rust and go-cgo.


beckermr:
It looks like the license data is out of date again. Maybe we need to add this update to pre-commit so it is easy to run and we can use pre-commit to autofix the issue.

beckermr:
This can happen in another PR. For now we can simply update the list here, though I forget how. advice @conda-forge/core?

MementoRC:
CI note: the only failure is `test_license_files_up_to_date`, which is unrelated to this change.

The SPDX list in the repo has drifted from upstream (e.g. `BSD-Source-alt-GPL`, `ZPL-2.1`). `main` last passed on 2026-09-10 at 7f648e58, which is this branch's base, and other open PRs fail the same way. The fix is `python -m conda_smithy.linter.update_licenses_list`, which I left out so this stays a lint-only change. Happy to add it here if you prefer.

Everything else passed: 5416 tests, including the stdlib cases for ocaml.

PR #2694, opened 2026-09-18
Split the upload step from building

### LINKED ISSUE #2614: Split upload from builds
> Similar to https://github.com/conda-forge/conda-smithy/issues/1472 in the sense of instrumenting "passing artifacts"

There are several upsides to this:

- From a security standpoint, we'd be better positioned if we were to separate uploads from builds. Even in the same workflow, different runners can run under different environments, which allows to scope tokens differently. It would also mitigate against cache poisoning attacks.
- Allows us to retry failed uploads without having to rebuild the whole thing again, which is particularly painful in large packages like pytorch.
- We can use different types of runners for the uploads (e.g. everything on Linux).
- We _could_  (if we want) aggregate packages in a single upload step.

I think this is completely feasible in GHA, and AZP shouldn't pose much friction either.

mgorny:
Checklist

Fixes #2614


conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -274,6 +262,34 @@ jobs:
         fi
         echo "status=$STATUS" >> $GITHUB_OUTPUT
 
+    - name: Upload packages
```
jaimergp, line 265: This needs to be a separate _job_, not _step_ in the same job. Otherwise we are in the same machine with the same secrets and disk state.

conda_smithy/templates/github-actions.yml.tmpl
```diff
@@ -132,12 +132,9 @@ jobs:
         CONFIG: {% raw %}${{ matrix.CONFIG }}{% endraw %}
         MINIFORGE_HOME: {% raw %}${{ matrix.tools_install_dir }}{% endraw %}
         DOCKER_IMAGE: {% raw %}${{ matrix.DOCKER_IMAGE }}{% endraw %}
-{%- if upload_on_branch %}
-        UPLOAD_ON_BRANCH: {{ upload_on_branch }}
-{%- endif %}
-        UPLOAD_PACKAGES: {% raw %}${{ matrix.UPLOAD_PACKAGES }}{% endraw %}
```
jaimergp, line 138: What's the behaviour for undefined variables here?

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -129,94 +129,105 @@ fi
...
         upload_package {% if conda_forge_output_validation %}--validate --feedstock-name="${FEEDSTOCK_NAME}" {% endif %}{% if private_upload %} --private{% endif %} "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
+
+        ( endgroup "Uploading packages" ) 2> /dev/null
     fi
+    ;;
 
-    ( endgroup "Uploading packages" ) 2> /dev/null
-fi
+*)
+    echo "Invalid STEP_ACTION: ${STEP_ACTION}" >&2
```
jaimergp, line 227: Should probably report valid values too.

conda_smithy/templates/github-actions.yml.tmpl
mgorny: What do you mean? The goal is for the build script not to use the variable anymore.

conda_smithy/templates/build_steps.sh.tmpl
mgorny: I'm thinking of taking the approach of splitting the scripts instead. That is:

1. Have `run_docker_build.sh` as a generic "start Docker and run specified script" wrapper.
2. Adding a new `upload_packages.sh` script for uploading from both Linux and macOS jobs (I'd love to use the same script for Windows too, but not sure if that's going to fly).
3. Splitting the init logic from `build-steps.sh` and `run_osx_build.sh` into reusable scripts for Linux and macOS respectively, and using them from `build-steps.sh`, `run_osx_build.sh` and `upload_packages.sh`.

See also #2695 for initial cleanup of the current scripts.

PR #2695, opened 2026-09-18
Reduce the divergence between Linux and macOS build scripts

mgorny:
Checklist

Reduce the divergence between Linux (`build-steps.sh`) and macOS (`run-osx-build.sh`) build scripts that seem to have accumulated over time.  Notably:

- Add a new early "Configuring build environment" group, and sync the "Provisioning base env ..." and "Configuring conda" groups.
- Declare a bunch of extra variables (especially on macOS) to make more commands match between the two scripts.
- Sync comments, argument order and output commands whenever possible.


conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -129,6 +133,9 @@ fi
 
 ( endgroup "Configuring conda" ) 2> /dev/null
 
+# make the build number clobber
+make_build_number "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
```
mgorny, line 137: Open question: on macOS, we're doing this only with `{%- if conda_build_tool != "rattler-build" %}`. Is there a reason we're doing this unconditionally on Linux?

mgorny:
I'm mostly doing this to make #2694 easier, and ideally share more code across platforms long-term.

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -37,30 +37,37 @@ solver: {{ conda_solver }}
 CONDARC
 
 {%- if conda_install_tool == "micromamba" %}
+
+( startgroup "Provisioning base env with micromamba" ) 2> /dev/null
```
jaimergp, line 43: I don't think we can nest groups :(

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -37,30 +37,37 @@ solver: {{ conda_solver }}
...
     pip {{ conda_install_tool_deps }} {{ conda_build_tool_deps }} {{ " ".join(remote_ci_setup) }}
+( endgroup "Provisioning base env with micromamba" ) 2> /dev/null
 {%- elif conda_install_tool == "pixi" %}
+( startgroup "Provisioning base env with pixi" ) 2> /dev/null
 pushd "${FEEDSTOCK_ROOT}"
 arch=$(uname -m)
 if [[ "$arch" == "x86_64" ]]; then
   arch="64"
 fi
+PIXI_ENV=docker-build-linux-${arch}
```
jaimergp, line 61: Let's make sure this doesn't get in the way of Pixi's internals. I think they set this too?

conda_smithy/templates/build_steps.sh.tmpl
```diff
@@ -14,7 +14,7 @@ source ${FEEDSTOCK_ROOT}/.scripts/logging_utils.sh
 {# This closes the matching `startgroup` on `run_docker_build.sh` #}
 ( endgroup "Start Docker" ) 2> /dev/null
 
-( startgroup "Configuring conda" ) 2> /dev/null
+( startgroup "Configuring build environment" ) 2> /dev/null
```
jaimergp, line 19: You are not closing this group, IIUC.

conda_smithy/templates/build_steps.sh.tmpl
jaimergp: This command is backed up by https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/f7bd98b4cfe77fbfbbb4af1741ba1f7ab9eec60b/recipe/conda_forge_ci_setup/build_utils.py#L287. I don't think this is relevant these days, tbh. rattler-build doesn't support clobbering either, and the code doesn't target rattler-build at all anyway.

So:

* This should only apply to conda-build
* It can probably be removed, since this code is 8 years old and designed to target dual compiler setups

xref https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/19

cc @isuruf, @hmaarrfk, @mariusvniekerk, who have touched that function in the past

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -1,14 +1,25 @@
...
+# PLEASE NOTE: This script has been automatically generated by conda-smithy. Any changes here
+# will be lost next time ``conda smithy rerender`` is run. If you would like to make permanent
+# changes to this script, consider a proposal to conda-smithy so that other feedstocks can also
+# benefit from the improvement.
+
 # -*- mode: jinja-shell -*-
 
-source .scripts/logging_utils.sh
+set -xeuo pipefail
+export FEEDSTOCK_ROOT=.
```
jaimergp: What's this for? Simple parity with Linux? Might be dangerous to run this way that depends so much on the working directory. Shall we have some fallbacks to `BASH_SOURCE` or something like that?

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -1,14 +1,25 @@
...
 
+# PLEASE NOTE: This script has been automatically generated by conda-smithy. Any changes here
+# will be lost next time ``conda smithy rerender`` is run. If you would like to make permanent
+# changes to this script, consider a proposal to conda-smithy so that other feedstocks can also
+# benefit from the improvement.
+
 # -*- mode: jinja-shell -*-
 
-source .scripts/logging_utils.sh
+set -xeuo pipefail
```
jaimergp, line 10: Yay!

conda_smithy/templates/run_osx_build.sh.tmpl
jaimergp: Ok now I see that we were already doing this implicitly later in the code. Still I think it wouldn't be a bad idea.

conda_smithy/templates/build_steps.sh.tmpl
mgorny: Ah, good catch, implied by the unclosed group above :-).

conda_smithy/templates/build_steps.sh.tmpl
mgorny: FWICS they use `PIXI_ENVIRONMENT_NAME`, and we don't export it anyway.

conda_smithy/templates/run_osx_build.sh.tmpl
mgorny: `${PWD}` then?

mgorny:
Here's an updated test run: https://github.com/mgorny/dbfread-feedstock/actions/runs/35607560110

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -92,30 +106,28 @@ export CONDA_SOLVER="{{ conda_solver }}"
...
-{{ conda_install_tool }} install --update-specs --quiet --yes --channel conda-forge --strict-channel-priority \
+{%- if conda_install_tool in ("conda", "mamba") %}
+{{ conda_install_tool }} install --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
     pip {{ conda_install_tool_deps }} {{ conda_build_tool_deps }} {{ " ".join(remote_ci_setup) }}
 {%- endif %}
 
 {% if local_ci_setup %}
 conda uninstall --quiet --yes --force {{ " ".join(remote_ci_setup) }}
-pip install --no-deps {{ recipe_dir }}/.
+pip install --no-deps "${RECIPE_ROOT}/."
```
jaimergp, line 116:
```suggestion
pip install --no-deps "${RECIPE_ROOT}"
```

This is redundant, isn't it?

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -92,30 +106,28 @@ export CONDA_SOLVER="{{ conda_solver }}"
...
 {%- endif %}
-
-echo -e "\n\nSetting up the condarc and mangling the compiler."
-setup_conda_rc ./ ./{{ recipe_dir }} ./.ci_support/${CONFIG}.yaml
+# set up the condarc
+setup_conda_rc "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
```
jaimergp, line 150: Now seeing this, I'm noticing it's different than in Linux, probably because of the Docker container. Philosophically, I'd like to move away from system-level or user-level changes (everything should be ephemeral whenever possible), but that's out of scope here.

conda_smithy/templates/run_osx_build.sh.tmpl
```diff
@@ -151,66 +162,80 @@ echo -e "\n\nRunning the build setup script."
 
 {%- if conda_build_tool != "rattler-build" %}
 
-echo -e "\n\nMaking the build clobber file"
-make_build_number ./ ./{{ recipe_dir }} ./.ci_support/${CONFIG}.yaml
+# make the build number clobber
+make_build_number "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
```
jaimergp, line 166: Flagging this for removal based on the outcome at https://github.com/conda-forge/conda-smithy/pull/2695/files#r4061720048

jaimergp:
Thank you making the log groups more consistent, btw! Did you check on Windows, or is there more work down the road where you will take a look later anyway?

conda_smithy/templates/run_osx_build.sh.tmpl
jaimergp: Either that or `${FEEDSTOCK_ROOT:-${PWD}}`? But I don't know if we want to allow overrides there. Do we do that in other platforms?

conda_smithy/templates/run_osx_build.sh.tmpl
mgorny: In osx, yes (it's `./.`). But it's here for consistency with Linux where it will be `some_path/.`. That said, I have no clue if the extra `/.` will really ever make any difference; I'm guessing it's meaningful when the path is a single directory.

mgorny:
I didn't think of looking at Windows, but it definitely makes sense to include that. I'll do in in a few minutes, unless this PR ends up considered ready by then, then I can do it as a followup.

conda_smithy/templates/run_osx_build.sh.tmpl
mgorny: I don't think so. FWICS respecting `FEEDSTOCK_ROOT` in `build-steps.sh` stems from the fact that it's called from a few other scripts that set `FEEDSTOCK_ROOT` (to a fixed value).

mgorny:
Okay, I've started the work on Windows. More to come tomorrow.

beckermr (commented):
A small comment. When you all are done with this PR, please rerender 

- the [conda-forge-ci-setup-feedstock](https://github.com/conda-forge/conda-forge-ci-setup-feedstock)
- at least one other feedstock with both linux and osx builds

with the changes in smithy to ensure things still build correctly.

conda_smithy/templates/run_osx_build.sh.tmpl
mgorny: Thinking about it further, I'll make all these paths absolute; that should be safer long-term.

mgorny:
This was more involving than anticipated but I'm done with initial Windows sync. Now I need to resync osx with Linux, and then Windows with Linux (for stuff that isn't in osx).

mgorny:
Question:

Both macOS, Linux and Windows have (the equivalent of):

```bash
{%- if conda_install_tool in ("conda", "mamba") %}
# Provision the necessary dependencies to build the recipe later
echo Installing dependencies
{{ conda_install_tool }} install --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
    pip {{ conda_install_tool_deps }} {{ conda_build_tool_deps }} {{ " ".join(remote_ci_setup) }}
{%- endif %}
```

However, Linux additionally has:

```bash
{%- if conda_install_tool == "mamba" and (conda_build_tool_deps != "" or conda_install_tool_deps != "") %}
{{ conda_install_tool }} update --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
    pip {{ conda_install_tool_deps }} {{ conda_build_tool_deps }} {{ " ".join(remote_ci_setup_update) }}
{%- endif %}
```

Is this because mamba is preinstalled in the Docker image, or is this something we want to mirror on other platforms?

mgorny:
Okay, I'm done here. Some of the changes look unnecessarily verbose, but they generally make the script closer to one another. The results are best seen in a tool like `vimdiff`. I've also aimed for more general consistency in whitespace use and messaging.

The latest rerenders:
- conda-forge-ci-setup: https://github.com/mgorny/conda-forge-ci-setup-feedstock/commit/f4f408dc7ddbcdb680c9ae2a1151a051ca38e4bf
- dbfread; conda + rattler-build: https://github.com/mgorny/dbfread-feedstock/commit/4cf61d7349d42faef96c9c0203db0e9a2186b948
- dbfread: pixi + rattler-build: https://github.com/mgorny/dbfread-feedstock/commit/d28e2b967191f39ada30e6a2ca3ef1eedb606b6e
- dbfread: conda + conda-build: https://github.com/mgorny/dbfread-feedstock/commit/41dd6e751422c7e04ac057ee9d18f15497779478

PR #2697, opened 2026-09-21
bump python version for root environment of build tools

h-vetinari:
Was just in the process of testing this in https://github.com/conda-forge/systemd-feedstock/pull/83 when the release happened, I wasn't fast enough to flag this in time. In any case, due to the fact that linux-riscv64 and win-arm64 only have python builds starting from 3.14, we cannot use a hard pin on 3.12 here.
```
  error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ conda-forge-ci-setup =4 * does not exist (perhaps a typo or a missing channel);
      └─ python =3.12 * does not exist (perhaps a typo or a missing channel).
```
We could distinguish per platform, but IMO the time is right to just use 3.14 here.


beckermr:
Ack sorry! Will cut another release after this pr.

h-vetinari:
nothing to apologise, we all overlooked (slash didn't test) this one so far

Issue #2698, opened 2026-09-21
fix remaining issues for enabling native riscv builds


h-vetinari:
Not everything's working yet based on https://github.com/conda-forge/numpy-feedstock/pull/401:
```
Run .scripts/create_pagefile.sh "${PAGEFILE_SIZE}"
+ PAGEFILE_SIZE=16
+ SWAPFILE=/swapfile
+ [[ ubuntu-24.04-riscv == *namespace-profile-* ]]
+ swapon --show
+ grep -q '^/swapfile'
+ [[ -f /swapfile ]]
+ sudo fallocate -l 16GiB /swapfile
+ sudo chmod 600 /swapfile
+ sudo mkswap /swapfile
Setting up swapspace version 1, size = 16 GiB (17179865088 bytes)
no label, UUID=13d1363a-d2ed-47ac-8658-f017290504f0
+ sudo swapon /swapfile
swapon: /swapfile: swapon failed: Invalid argument
```

This specific issue may have to be fixed somewhere between the docker-images, ci-setup & smithy, but this issue is intended to collect all required fixes before general v0/v1 recipes can use native riscv builds

* [x] #2686
* [x] #2697
* [x] `workflow_settings:`
  * [x] fix `pagefile_size`
  * [x] test `create_build_artifacts`
  * [x] test `clean_disk_space`

CC @lucascolley @mgorny @luhenry @velonica0 in case you're interested

h-vetinari:
FWIW, ongoing PRs to test this:
* https://github.com/conda-forge/numpy-feedstock/pull/401
* https://github.com/conda-forge/scipy-feedstock/pull/331
* https://github.com/conda-forge/systemd-feedstock/pull/83

h-vetinari:
When removing the swapfile, the numpy builds (v1 recipe) start OK.

velonica0:
> When removing the swapfile, the numpy builds (v1 recipe) start OK.

The `swapon` failure is because the RISE runner is a container with an overlayfs root, which cannot host a swap file (`swapon(2)` returns `EINVAL` regardless of what the host has). #2699 makes `create_pagefile.sh` detect that and skip with a warning instead of aborting the job. Actual swap for these nodes has to come from RISE; proposed in riseproject-dev/riscv-runner#114.

velonica0:
For spicy, the native riscv64 build works; the failures are:

[...]

PR #2699, opened 2026-09-22
Skip the swap file on runners whose filesystem cannot hold one

velonica0:
Fixes the `pagefile_size` item of #2698.

On the RISE `ubuntu-24.04-riscv` runners (#2686) the job runs in a container with an overlayfs root. `fallocate` and `mkswap` succeed, but `swapon(2)` needs a filesystem that can map the file to disk blocks, which overlayfs cannot, so it fails with `Invalid argument` and `set -e` aborts the job before the build starts (conda-forge/numpy-feedstock#401).

`create_pagefile.sh` now checks the filesystem type first (`stat -f -c %T`) and skips with a `::warning::` on `overlay*`/`tmpfs`; if `swapon` fails for another reason it warns, removes the file and continues. Runners that can hold a swap file (GitHub/Azure VMs, Namespace) are unchanged.

This does not give the RISE jobs swap; that has to be provisioned on the nodes (riseproject-dev/riscv-runner#114). It is the prerequisite: without it the step aborts the job even once the node has swap. Until then, feedstocks that need more than ~15 GiB on a 4-core node should keep `build_platform: linux_riscv64: linux_64`.

Verified all script paths with stubbed `stat`/`sudo`/`swapon`; `pytest -k "pagefile or workflow_settings"` passes.


conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -17,4 +44,9 @@ fi
 sudo fallocate -l "${PAGEFILE_SIZE}GiB" "${SWAPFILE}"
 sudo chmod 600 "${SWAPFILE}"
 sudo mkswap "${SWAPFILE}"
-sudo swapon "${SWAPFILE}"
+if ! sudo swapon "${SWAPFILE}"; then
```
h-vetinari: IMO this should continue to raise if it doesn't succeed. It's very annoying to expect a swap file (e.g. because of very long running jobs) and then find out that CI silently dropped it.

In such cases (aside from providers which cannot swap at all yet; those are excluded further up already), people should either restart CI or restrict `pagefile_size` to appropriate platforms

conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -4,10 +4,37 @@ set -ex
...
+        echo "::warning::${1}"
+    else
+        echo "WARNING: ${1}"
+    fi
+}
+
+# swapon(2) needs a filesystem that can map the swap file to disk blocks.
+# overlayfs (the root filesystem of containerized runners, e.g. the RISE
+# riscv64 runners) and tmpfs cannot, and fail with "Invalid argument".
+swapfile_supported() {
```
h-vetinari: The name of the function doesn't fit with the selective blocklist for two known cases where this doesn't work. But given the number of filesystems in linux, I doubt this list is exhaustive...

I would rather name this something like `fs_cant_swap` and then invert the if condition for the warning below

h-vetinari (changes requested):
Thanks for the PR and the upstream issue for swap support

conda_smithy/templates/create_pagefile.sh.tmpl
velonica0: Renamed to `fs_cant_swap` and inverted the condition.

conda_smithy/templates/create_pagefile.sh.tmpl
velonica0: Restored the hard failure.

conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -4,10 +4,37 @@ set -ex
...
 if [[ ${GHA_RUNS_ON} == *namespace-profile-* ]]; then
     SWAPFILE=/namespace/scratch/swapfile
 fi
+SWAPDIR=$(dirname "${SWAPFILE}")
+
+# Swap is a host-wide resource; don't try to find another writable volume
+# in a container, as the file may be deleted with the container while the
+# kernel still uses it as swap.
+if fs_cant_swap "${SWAPDIR}"; then
+    warn "${SWAPDIR} is on $(stat -f -c %T "${SWAPDIR}"), which cannot hold a swap file; skipping pagefile_size=${PAGEFILE_SIZE}"
```
h-vetinari: I'd use 
```suggestion
    warn "The path ${SWAPFILE} is on $(stat -f -c %T "${SWAPDIR}"), which cannot hold a swap file; skipping pagefile_size=${PAGEFILE_SIZE}"
```
here. The warning looks a bit strange otherwise
```
Warning: / is on overlayfs, which cannot hold a swap file; skipping pagefile_size=16
```

conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -4,10 +4,37 @@ set -ex
...
+    local fstype
+    fstype=$(stat -f -c %T "${1}")
+    [[ ${fstype} == overlay* || ${fstype} == tmpfs ]]
+}
+
 SWAPFILE=/swapfile
 if [[ ${GHA_RUNS_ON} == *namespace-profile-* ]]; then
     SWAPFILE=/namespace/scratch/swapfile
 fi
+SWAPDIR=$(dirname "${SWAPFILE}")
```
h-vetinari: can you move the dirname into `fs_cant_swap` please? then we don't need the extra variable in the overall script

h-vetinari (commented):
Tested this in https://github.com/conda-forge/numpy-feedstock/pull/401 and it looks good; just some minor nits about log messages and code simplification

conda_smithy/templates/create_pagefile.sh.tmpl
velonica0: Moved into fs_cant_swap, SWAPDIR is gone.

conda_smithy/templates/create_pagefile.sh.tmpl
```diff
@@ -4,10 +4,35 @@ set -ex
...
+
 SWAPFILE=/swapfile
 if [[ ${GHA_RUNS_ON} == *namespace-profile-* ]]; then
     SWAPFILE=/namespace/scratch/swapfile
 fi
+# Swap is a host-wide resource; don't try to find another writable volume
+# in a container, as the file may be deleted with the container while the
+# kernel still uses it as swap.
+if fs_cant_swap "${SWAPFILE}"; then
+    warn "The path ${SWAPFILE} is on $(stat -f -c %T "$(dirname "${SWAPFILE}")"), which cannot hold a swap file; skipping pagefile_size=${PAGEFILE_SIZE}"
```
h-vetinari: Looking at the current warning text,
```
Warning: The path /swapfile is on overlayfs, which cannot hold a swap file; skipping pagefile_size=16
```
I think we can improve this a little bit more still
```suggestion
    warn "The path ${SWAPFILE} is on a filesystem of type $(stat -f -c %T "$(dirname "${SWAPFILE}")"), which cannot hold a swap file; skipping pagefile_size=${PAGEFILE_SIZE}"
```

h-vetinari (approved):
Thank you!

